PR #2, opened 2022-10-24
Kill tensorspec

tcbegley:
This PR removes `TensorSpec`s from `tensordict` which are RL specific and hence out of scope for this library.

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -372,8 +372,10 @@ class TensorDictModule(nn.Module):
         # check if there is a non-initialized lazy module
         for m in self_copy.module.modules():
             if hasattr(m, "has_uninitialized_params") and m.has_uninitialized_params():
-                # TODO: what would we do here in the absence of a spec?
-                raise RuntimeError("Module has uninitialised parameters...")
+                raise ValueError(
+                    "Modules with uninitialized lazy components cannot be converted to "
+                    "functional modules."
+                )
 
         module = self_copy.module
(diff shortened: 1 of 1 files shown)


theOGognf:
I was thinking about opening an issue to request moving tensorspecs into tensordict when I came across this issue. It's kind of a bummer that tensorspec was moved out of tensordict. Most of my projects that use tensordicts often end up using tensorspecs while also having little-to-no additional benefit from using torchrl. Because of this, I often end up [copying the tensorspec implementation from torchrl directly into my project](https://github.com/theOGognf/rlstack/blob/main/src/rlstack/specs.py) to avoid adding torchrl as another dependency. This isn't to say torchrl isn't a great project - it's just that I've noticed that I tend to only use tensorspecs from it when working on niche RL projects or projects that have nothing to do with RL.

Is it worth opening an issue to track if it's desirable to move tensorspecs back into tensordict?

vmoens:
Yep I can see that point.
The context is that we want(ed) to move TensorDict to torch core and tensor spec seems to be the extra bit that our pytorch core colleagues might be more insecure about. TensorSpec is a historical RL related concept (specs are everywhere in RL). If we move it here, we'll probably need more specs too to cover non RL domains. Not to say it's a bad idea but that will require more maintenance.
If you want, could you open a separate issue (referencing our chat here) and I'll add a few folks in the discussion?

theOGognf:
No problem - thanks for the prompt reply

PR #5, opened 2022-10-27
[Feature] TensorDict support for first class dimensions

tcbegley:
This PR adds support for first class dimensions to `TensorDict`.

I've added logic that is able to calculate how dimensions update when indexed or ordered which allows us to index and order both `MetaTensor` and empty `TensorDict` instances. I've tried to cover as many edge cases as possible in the tests.


vmoens (commented):
LGTM. 2 points of attention
- Can we test this with the other tensordict subclasses like we do in test_tensordict.py (stacked, viewed, unsqueezed, squeezed, etc)?
- Can we provide an example in the readme under the functorch section?

vmoens (approved):
LGTM Brilliant work, as usual

vmoens:
We’re using `_something` not for file-specific primitives but also for
feauters that are not exposed to the users.
For example _getitem_batch_size is used across files but we may not want
people to rely on it for code outside of TensorDict (+ we don’t necessarily
want to write a proper deep doc about it). Happy to discuss/revise those
conventions though!

shagunsodhani:
> We’re using `_something` not for file-specific primitives but also for feauters that are not exposed to the users. For example _getitem_batch_size is used across files but we may not want people to rely on it for code outside of TensorDict (+ we don’t necessarily want to write a proper deep doc about it). Happy to discuss/revise those conventions though!

Makes sense! Thanks for the clarification :)

PR #11, opened 2022-11-02
[Feature] In-place functionalization

vmoens:
## Description

Implements an in-place functional modules using tensordict.
TL;DR allows to do the following hacks:

```python
l = nn.Sequential(
    Linear(3, 4),
    LinearWithKwargs(4, 4),
)
params = make_functional(l)

x = torch.randn(3)
z = l(x, params=params)

y = l[0](x, params=params["0"])
z_bis = l[1](y, params=params["1"])

assert torch.allclose(z, z_bis)
```

It also allows you to retrieve the original module back in its original, stateful state (see below).

AFAICT it's fully vmap compatible.

A few words on behaviours:
In functorch, params was put at the beginning of the args. Here, it's put last. To be more precise, it is the first keyword argument with a default value (`None`). The reason for that is that we want the sub-modules to be able to ignore the params (as they're placed in each sub-module when the parent module is being called). But we'd still like to modify their `forward` method. Putting the `params` first would have been messy, with complicated if / else (if first arg is `None`, then consider that it is a `params` that is absent -- or something of the likes of it).

Another important deviation to note is that we're just using Params, not buffers.
This is highly opinionated but I don't really see the point of having 2 different pipelines, one for params and one for params and buffers. If I had to implement the `parameters`-only version, I would go for an arg in the `make_functional` that would make it ignore the buffers. Happy to discuss this with whomever is interested!

Finally we do the modifications in place. This should not be an issue though as one can simply call `_swap_states(module, params)` and gets back the original module.

**Runtime**: 
The script below gets the following runtimes, which means that our implementation is marginally slower on extremely shallow networks (as networks grow bigger, the compute time dedicated to tensor operations will take over the compute time of allocating and de-allocating the tensors to their modules):
```
instantiation, functorch: 1.798754305
instantiation, tensordict: 1.4238669880000003
exec, functorch: 4.167005495000001
exec, tensordict: 4.753181284
```
Code:

```python
from functorch import make_functional_with_buffers as functorch_make_functional
from tensordict.nn.functional_modules import make_functional
import torch
# we use deepcopy as our implementation modifies the modules in-place
from copy import deepcopy
from torch import nn

import timeit

## Creation
net = nn.Sequential(nn.Linear(2, 2), nn.Linear(2, 2), nn.Linear(2, 2),
                    nn.Sequential(nn.Linear(2, 2), nn.Linear(2, 2),
                                  nn.Linear(2, 2)))
print("instantiation, functorch:", timeit.timeit("functorch_make_functional(deepcopy(net))", globals={
    "functorch_make_functional": functorch_make_functional, "net": net,
    "deepcopy": deepcopy}, number=1000))

print("instantiation, tensordict:", timeit.timeit("make_functional(deepcopy(net))", globals={
    "make_functional": make_functional, "net": net,
    "deepcopy": deepcopy}, number=1000))

## Execution
# x = torch.randn(2, 2, 128)
x = torch.randn(2, 2)
fmodule, params, buffers = functorch_make_functional(deepcopy(net))
print("exec, functorch:", timeit.timeit("fmodule(params, buffers, x)", globals={
    "fmodule": fmodule, "params": params,
    "buffers": buffers, "x": x}, number=10000))

fmodule = deepcopy(net)
params = make_functional(fmodule)
print("exec, tensordict:", timeit.timeit("fmodule(x, params=params)", globals={
    "fmodule": fmodule, "params": params,
    "x": x}, number=10000))
```

cc @zou3519 @tcbegley @samdow 

TODO:
- [ ] Remove refs to Functional modules in common and sequential
- [ ] update docstrings
- [ ] Edit tests


tensordict/nn/common.py
```diff
@@ -25,14 +24,9 @@
     _has_functorch = False
 
 import torch
-from functorch import FunctionalModule, FunctionalModuleWithBuffers, vmap
-from functorch._src.make_functional import _swap_state
+from functorch import FunctionalModule, FunctionalModuleWithBuffers
```
zou3519, line 27: Looking at the API, nn.utils.stateless.functional_call seems to provide something similar (in-place functional modules, but not exactly):

```
l = nn.Sequential(
    Linear(3, 4),
    LinearWithKwargs(4, 4),
)
params = l.named_parameters()
x = torch.randn(3)
nn.utils.stateless.functional_call(l, params, x)
```
What are the advantages of this tensordict-based approach?

> Another important deviation to note is that we're just using Params, not buffers.
This is highly opinionated but I don't really see the point of having 2 different pipelines, one for params and one for params and buffers.

I never liked the old make_functional design where we separated modules that had params and modules had params+buffers

vmoens: > What are the advantages of this tensordict-based approach?

I don't know if everyone would agree about the fact that these are advantages, but here's my take:
- You can access the params/buffers in a very similar way as you access your modules. For instance, once you've called `make_functional` you can do `module.sub_module[some_index](input, params["sub_module", "some_index"]`. Think of a transformer where you need to query the encoder / decoder separately. You can collect your params in a TensorDict just once and play with it as you like. With the pytorch approach, you'd need either (1) to query `named_parameters` for each sub-module (which can be expensive) or (2) write some complicated indexing methods that, actually, are taken care of by tensordict.
-  You benefit from all the tensordict features (compat with vmap, indexing, reshaping, stacking for ensembling, pre-allocation etc)
Here are some examples: 
```python
vmap(module.sub_module, (None, 0))(x, params["sub_module"].expand(10))
module.sub_module(x.cuda(), params["sub_module"].cuda())
module(x, params.detach())
```
- You query your model directly, no need for a helper function. IMO it's much more good-looking, and you can revert things and get back your original module. With `functional_call` you still have a model that is full of params, which may come with some problems (ie. memory allocation of a whole network, or loosing sight of why those params are there etc).
- You have a functional module where you can replace all the params if you want using `_swap_state` (though I guess the torch implementation does something similar under the hood).
Hope that clarifies things a bit!

PR #12, opened 2022-11-03
[Doc] Minor edits of README

vmoens:
## Description

Minor edits of the README


README.md
```diff
@@ -134,16 +134,69 @@ y = vmap(fmodule, (0, None))(weights, x)
...
+def foo():
+    tensordict = TensorDict({}, [])
+    tensordict["a"] = torch.randn(3)
+    tensordict["b"] = TensorDict({"c": torch.zeros(2)}, [])
+    return tensordict
+```
+and you would like to call this function repeatedly. You could do this in two ways.
+The first would simply be to stack the calls to the function:
+```python
+tensordict = torch.stack([foor() for _ in range(N)])
```
tcbegley, line 152:
```suggestion
tensordict = torch.stack([foo() for _ in range(N)])
```

README.md
```diff
@@ -134,16 +134,69 @@ y = vmap(fmodule, (0, None))(weights, x)
...
+        a: Tensor(torch.Size([10, 3]), dtype=torch.float32),
+        b: TensorDict(
+            fields={
+                c: Tensor(torch.Size([10, 2]), dtype=torch.float32)},
+            batch_size=torch.Size([10]),
+            device=None,
+            is_shared=False)},
+    batch_size=torch.Size([10]),
+    device=None,
+    is_shared=False)
```
tcbegley, line 173: Even though it's not realistic output, would it make sense to replace `10` with `N` here?

Alternatively maybe we could add `N = 10` above?

tcbegley (approved):
LGTM

README.md
vmoens: fixed

README.md
```diff
@@ -183,16 +183,69 @@ torch.testing.assert_close(
...
+and you would like to call this function repeatedly. You could do this in two ways.
+The first would simply be to stack the calls to the function:
+```python
+tensordict = torch.stack([foo() for _ in range(N)])
+```
+However, you could also choose to preallocate the tensordict:
+```python
+tensordict = TensorDict({}, [N])
+for i in range(N):
+    tensordict[i] = foo()
```
shagunsodhani, line 207: Nitpick: The syntax is a bit confusing. `tensordict = TensorDict({}, [N])` looks like a list of tensordicts :) Not sure what we can change but sharing the perspective as a new user.

PR #15, opened 2022-11-03
[BugFix] Merge partial sub-tensordicts with flatten ones if unflattened

### LINKED ISSUE #8: [BUG] Partially instantiated sub-tensordict silently collide with flatten ones when unflattened
## Describe the bug

When a tensordict is created with mixed nested and flattened keys, a call to `unflatten_keys` silently squashes them against each other.

## To Reproduce

```python
tensordict = TensorDict({"a": [1, 2], "c.a": [1, 2], "c": TensorDict({"b": [1, 2]}, [])}, [])
print(tensordict.unflatten_keys("."))
```
This produces the following output
```
TensorDict(
    fields={
        a: Tensor(torch.Size([2]), dtype=torch.int64),
        c: TensorDict(
            fields={
                a: Tensor(torch.Size([2]), dtype=torch.int64)},
            batch_size=torch.Size([]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```
where the `"c.b"` key has disappeared.

## Expected behavior

Both `"c.b"` and `"c.a"` should be there.

hi-fei:
## Description

Merge the instantiated sub-tensordict to the unflatted sub-tensordict to avoid collision.  

## Motivation and Context

close #8 

## Types of changes

## Checklist

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1510,13 +1510,13 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
 
         for key, list_of_keys in to_unflatten.items():
             tensordict = TensorDict({}, batch_size=self.batch_size, device=self.device)
+            if key in self:
+                tensordict.update(self[key])
             for (old_key, new_key) in list_of_keys:
                 value = self[old_key]
                 tensordict[new_key] = value
                 if inplace:
                     del self[old_key]
-                if key in self:
-                    tensordict.update(self[key])
             out.set(key, tensordict.unflatten_keys(separator=separator))
         return out
 
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1515,6 +1515,8 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
                 tensordict[new_key] = value
                 if inplace:
                     del self[old_key]
+                if key in self:
+                    tensordict.update(self[key])
             out.set(key, tensordict.unflatten_keys(separator=separator))
         return out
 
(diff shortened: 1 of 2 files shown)


test/test_tensordict.py
```diff
@@ -2479,6 +2479,12 @@ def test_tensordict_device(self, device):
         assert tensordict["c"].device == device
 
 
+@pytest.mark.parametrize("separator", [".", "-"])
+def test_unflatten_keys_collision(separator):
+    td = TensorDict({"a": [1, 2], f"c{separator}a": [1, 2], "c": TensorDict({"b": [1, 2]}, [])}, [])
+    ref = TensorDict({"a": [1, 2], "c": TensorDict({"a": [1, 2], "b": [1, 2]}, [])}, [])
+    assert repr(td.unflatten_keys(separator)) == repr(ref)
```
tcbegley, line 2486: Comparing the results of `repr` feels like it might be a bit brittle. Might be better to do something like

```python
from tensordict.tensordict import assert_allclose_td

assert_allclose_td(td.unflatten_keys(separator), ref)
```

tensordict/tensordict.py
```diff
@@ -1515,6 +1515,8 @@ def unflatten_keys(
                 tensordict[new_key] = value
                 if inplace:
                     del self[old_key]
+                if key in self:
+                    tensordict.update(self[key])
```
tcbegley, line 1519: I don't think this needs to be in the inner for-loop right? Could be done immediately after instantiating `tensordict`.

hi-fei: Yeah. Just moved it to the outer loop.

test/test_tensordict.py
hi-fei: Got it.

tcbegley (approved):
LGTM!

Issue #14, opened 2022-11-03
[Feature Request] Improve first-class dimensions support


tcbegley:
## Motivation

#5 added support for first-class dimensions. There is however at least one missing feature and some ways we could consider improving this:

## Missing Feature

It should be possible to index / order with tuples of first-class dimensions for convenient flattening / reshaping operations. Consider the following example from the `torchdim` repo.
```python
i, j, k = dims(3)
j.size = 2
A = torch.rand(6, 4)
a = A[(i, j), k] # split dim 0 into i,j
print(i.size, j.size, k.size)
> 3 2 4

r = a.order(i, (j, k)) # flatten j and k
print(r.shape)
> torch.Size([3, 8])
```

Currently this will lead to an error if attempted on a `TensorDict`.

## Possible improvement

One possible enhancement would be to allow the user to specify first-class dimensions in the `batch_size` argument when instantiating a `TensorDict`. E.g.

```python
d1, d2 = dims(2)
td = TensorDict({}, batch_size=[d1, d2, 3, 4])

# roughly equivalent to
td = TensorDict({}, batch_size=[1, 2, 3, 4])[d1, d2]
```
 
The difficulty here is likely to be that if the tensordict is empty, we do not know the size of the first class dimensions and they will be unbound until something is added to the tensordict. Assuming that can be worked around it might be a nice convenience?

tcbegley:
#19 resolves the missing feature documented above.

Having thought more about the feature proposal of allowing first-class dimensions in the `batch_size` argument of `TensorDict`, I don't think it's really all that much more convnient, e.g.

```python
d1, d2 = dims(2)
td = TensorDict({}, batch_size=[d1, d2, 3, 4])
# vs
td = TensorDict({}, batch_size=[1, 2, 3, 4])[d1, d2]
```

from my example above, the first example is marginally more compact, but is actually missing information on the size of `d1` and `d2`. You could do something complicated to infer the size from the contents of the TensorDict, but it's a fair bit of complexity for minimal user benefit and there are still cases where behaviour would not be well-defined, e.g.

```python
d = dims(1)
# what is the batch_size of td here? it's ambiguous
td = TensorDict({}, batch_size=[d]).order(d)
```

PR #19, opened 2022-11-04
[Feature] Support splitting / flattening with first-class dims

tcbegley:
This PR adds support for splitting / flattening TensorDict batch dimensions using first-class dimensions from functorch as raised in #14 

An example:

```python
# splitting dimensions
td = TensorDict({"a": torch.rand(6, 3, 3)}, [6, 3])
d1, d2, d3 = dims(3)
d1.size = 3
td_dim = td[(d1, d2), d3]  # split first dimension in two
td2 = td_dim.order(d1, d2, d3)
assert td2.batch_size == torch.Size([3, 2, 3])
assert td2["a"].shape == torch.Size([3, 2, 3, 3])

d1, d2 = dims(2)
td_dim = td[d1, d2]
td2 = td_dim.order((d1, d2))
assert td2.batch_size == torch.Size([18])
assert td2["a"].shape == torch.Size([18, 3])
```

### DIFF
--- a/tensordict/metatensor.py
+++ b/tensordict/metatensor.py
@@ -26,6 +26,7 @@ from tensordict.utils import (
     _get_ordered_shape,
     _get_ordered_dims,
     _get_shape,
+    _is_first_class_dim,
 )
 
 try:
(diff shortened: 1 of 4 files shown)


vmoens (approved):
LGTM thanks!

PR #20, opened 2022-11-04
[Feature] Add TensorDictKeysView

tcbegley:
This PR adds a new object `_TensorDictKeysView` which is returned when accessing `tensordict.keys()` and holds a reference to the original tensordict. This allows us to modify the behaviour of `key in tensordict.keys()` and use tuples for nested key lookups. For example

```python
import torch

td = TensorDict(
    {"a": TensorDict({"b": torch.rand(1, 2)}, [1, 2]), "c": torch.rand(1)}, 
    [1],
)

assert "a" in td.keys()
assert ("a",) in td.keys()
assert ("a", "b") in td.keys()
assert ("a", "c") not in td.keys()
```

This PR also disallows checks like `item in tensordict` except in the case of `tensordict in lazy_stacked_tensordict` where it will check if the tensordict is one of the tensordicts that will be lazily stacked when items are accessed.

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -3489,6 +3489,23 @@ def merge_tensordicts(*tensordicts: TensorDictBase) -> TensorDictBase:
     return TensorDict({}, [], device=td.device).update(d)
 
 
+class _LazyStackedTensorDictKeysView(_TensorDictKeysView):
+    def __len__(self):
+        return len(self.tensordict.valid_keys)
+
+    def __iter__(self):
+        return iter(self.tensordict.valid_keys)
+
+    def __contains__(self, key):
+        if isinstance(key, str):
+            return key in self.tensordict.valid_keys
+        raise TypeError(
+            "TensorDict keys are always strings. Membership checks on a "
+            "LazyStackedTensorDict do not currently support tuples of strings for "
+            "nested lookups, so only strings are allowed."
+        )
+
+
 class LazyStackedTensorDict(TensorDictBase):
     """A Lazy stack of TensorDicts.
 
@@ -3819,9 +3836,8 @@ class LazyStackedTensorDict(TensorDictBase):
             del self._orig_batch_size
         self._batch_size = new_size
 
-    def keys(self) -> Iterator[str]:
-        for key in self.valid_keys:
-            yield key
+    def keys(self) -> _LazyStackedTensorDictKeysView:
+        return _LazyStackedTensorDictKeysView(self)
 
     def _update_valid_keys(self) -> None:
         valid_keys = set(self.tensordicts[0].keys())
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1520,7 +1520,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
     def unflatten_keys(
         self, separator: str = ",", inplace: bool = False
     ) -> TensorDictBase:
-        to_unflatten = defaultdict(lambda: list())
+        to_unflatten = defaultdict(list)
         for key in self.keys():
             if separator in key[1:-1]:
                 split_key = key.split(separator)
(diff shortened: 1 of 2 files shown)


tcbegley:
@vmoens the behaviour of `.keys` on `LazyStackedTensorDict` and `SavedTensorDict` is quite different to `TensorDict` at the moment, they each produce generators of the top level keys. Would you prefer that they be brought in line with the behaviour of `TensorDict`? It seems a bit challenging:

- In the case of `SavedTensorDict`, I guess we wouldn't want to load the data for every membership check, so does that mean traversing the contents at instantiation and capturing all keys (nested or otherwise) in some kind of data structure?

- In the case of `LazyStackedTensorDict` if I understand directly the keys are those that are in the intersection of all keys in the list of tensordicts to be stacked. So again I'm wondering if one would need to traverse and extract keys at instantiation to properly support this...

Let me know whether you think this is ok as is or whether it's worth spending some time on these two cases.

vmoens:
SavedTensordict should be deprecated. You can use TensorDict with memmap and save it with torchsnapshot. I would not worry too much about it!

vmoens:
> In the case of LazyStackedTensorDict if I understand directly the keys are those that are in the intersection of all keys in the list of tensordicts to be stacked. So again I'm wondering if one would need to traverse and extract keys at instantiation to properly support this...

Hmm that's gonna be a big overhead
Can we just put an error message if someone uses `__contains__` with a tuples of str with LazyStacked?

tcbegley:
I've addressed the outstanding bits:

- not done anything with `SavedTensorDict` per your comment
- added a `_LazyStackedTensorDictKeysView` class to inform the user that only strings are supported for now.

So I think this is ready for a final review.

vmoens (approved):
LGTM

tensordict/tensordict.py
```diff
@@ -89,6 +88,38 @@
 _STR_MIXED_INDEX_ERROR = "Received a mixed string-non string index. Only string-only or string-free indices are supported."
 
 
+class _TensorDictKeysView:
+    """
+    Wrapper class that enables richer behaviour of `key in tensordict.keys()`
```
shagunsodhani, line 93: Maybe add something along the lines of "_TensorDictKeysView is returned when accessing tensordict.keys() and holds a reference to the original tensordict. This class enable us to modify the behaviour of key in tensordict.keys() and use tuples for nested key lookups. For example ... " (taken from the description of the PR)

tensordict/tensordict.py
```diff
@@ -1524,6 +1555,19 @@ def __len__(self) -> int:
         """Returns the length of first dimension, if there is, otherwise 0."""
         return self.shape[0] if self.batch_dims else 0
 
+    def __contains__(self, key):
+        # by default a Mapping will implement __contains__ by calling __getitem__ and
```
shagunsodhani, line 1559: probably this is out of scope of this PR but why not fall back to `return key in self.keys()` in this case? Membership check is a very common op and users would expect this to work.

vmoens: This is an open discussion.
Here's the summary of our thoughts, correct me if I'm missing something @tcbegley 

The default behaviour of `__contains__` is to try `__getitem__` and return False if an error is thrown. For the lazy tensordicts (views and stacks) getitem is expensive, so we don't want to do that.
- we try as much as possible to avoid behaviours that are not well defined and/or duplicated (there should not be more than one simple way of doing something). It's a general policy for pytorch. With a stack of tensordicts you might say "I want to know if this tensordict is in the stack" and use `in` for that. If we enable that it would be strange to use `in` timely look up for keys. I also think a good error message goes a long way.
Happy to reconsider though! I'm not 100% convinced by my own arguments :)

tensordict/tensordict.py
tcbegley: I'll be doing some more work on this today, so I'm happy to expand the docstring, I think that's a good idea.

tensordict/tensordict.py
shagunsodhani: Regarding `The default behaviour of __contains__ ...` I understand the rationale behind not supporting lazy tds so it is fine to raise an error for that case.

Regarding `behaviours that are not well defined and/or duplicated`, I think user experience triumps this. Even PT added numpy compatibility (eg supporting both dims and axis). This is how the user experience would look right now (feel free to correct me if I am wrong):

1. User tries `key in td`.
2. Runs into an error.
3. Calls key in td.keys()`

1 is surprising behavior, more so given that all python containers support "in". 3 is confusing because python programmers are conditioned to prefer `key in _dict` over `key in _dict.keys()` as the former is O(1) while the latter is O(n)

Issue #22, opened 2022-11-07
[Feature Request] Nested keys compatibility improvement


vmoens:
## Motivation

We need to improve the compatibility with nested keys. This should include:
- [ ] `TensorDict.get(nested_key)`
- [ ] `TensorDict.set(nested_key, value)` (+ creation of the sub-td if the original key is missing)
- [ ] `TensorDict.set_(nested_key, value)`
- [ ] `TensorDict.select(*keys)`
- [ ] `TensorDict.exclude(*keys)`
- [ ] `TensorDict.keys()`
- [ ] `TensorDict.set_default(key, value)`

We will cover once it's implemented
- [ ] `TensorDict.pop(key)`

cc @tcbegley

PR #24, opened 2022-11-07
[Feature] Support nested keys in get, set and .keys()

tcbegley:
This PR addresses some of the issues raised in #22, specifically:

- [x] `TensorDict.get(nested_key)`
- [x] `TensorDict.set(nested_key, value)` (+ creation of the sub-td if the original key is missing)
- [x] `TensorDict.set_(nested_key, value)`
- [x] Return nested keys when iterating over `tensordict.keys()` (and by extension `tensordict.values()` and `tensordict.items()`) if the keyword argument `include_nested` is set to `True`. 

Functionality overview:

```python
>>> t = TensorDict({"a": TensorDict({"b": TensorDict({"c": [1, 2]}, [])}, [])}, [])
>>> t.keys()
<tensordict.tensordict._TensorDictKeysView at 0x12be4cc70>
>>> list(t.keys())
['a']
>>> list(t.keys(include_nested=True))
[('a', 'b', 'c'), ('a', 'b'), 'a']
>>> list(t.items())
[('a',
  TensorDict(
      fields={
          b: TensorDict(
              fields={
                  c: Tensor(torch.Size([2]), dtype=torch.int64)},
              batch_size=torch.Size([]),
              device=None,
              is_shared=False)},
      batch_size=torch.Size([]),
      device=None,
      is_shared=False))]
>>> list(t.items(include_nested=True))
[(('a', 'b', 'c'), tensor([1, 2])),
 (('a', 'b'), TensorDict(
      fields={
          c: Tensor(torch.Size([2]), dtype=torch.int64)},
      batch_size=torch.Size([]),
      device=None,
      is_shared=False)),
 ('a',
  TensorDict(
      fields={
          b: TensorDict(
              fields={
                  c: Tensor(torch.Size([2]), dtype=torch.int64)},
              batch_size=torch.Size([]),
              device=None,
              is_shared=False)},
      batch_size=torch.Size([]),
      device=None,
      is_shared=False))]
>>> t.set(("a", "b", "d"), [1])
TensorDict(
    fields={
        a: TensorDict(
            fields={
                b: TensorDict(
                    fields={
                        c: Tensor(torch.Size([2]), dtype=torch.int64),
                        d: Tensor(torch.Size([1]), dtype=torch.int64)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False)},
            batch_size=torch.Size([]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
>>> ("a", "b", "c") in t.keys()
  File "/Users/vmoens/venv/rl/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3441, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-10-ae4387df488a>", line 1, in <module>
    ("a", "b", "c") in t.keys()
  File "/Users/vmoens/Repos/RL/tensordict/tensordict/tensordict.py", line 162, in __contains__
    raise TypeError(
TypeError: Nested membership checks with tuples of strings is only supported when setting `include_nested=True`.
>>> ("a", "b", "c") in t.keys(include_nested=True)
True
>>> ("a", "b", "c") in t
  File "/Users/vmoens/venv/rl/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3441, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-12-6472026aa155>", line 1, in <module>
    ("a", "b", "c") in t
  File "/Users/vmoens/Repos/RL/tensordict/tensordict/tensordict.py", line 1645, in __contains__
    raise NotImplementedError(
NotImplementedError: TensorDict does not support membership checks with the `in` keyword. If you want to check if a particular key is in your TensorDict, please use `key in tensordict.keys()` instead.
```


tensordict/tensordict.py
```diff
@@ -3741,9 +3770,13 @@ def _stack_onto_(
...
     def get(
         self,
-        key: str,
+        key: NESTED_KEY,
         default: Union[str, COMPATIBLE_TYPES] = "_no_default_",
     ) -> COMPATIBLE_TYPES:
+        # TODO: the stacking logic below works for nested keys, but the key in
+        # self.valid_keys check will fail and we'll return the default instead.
+        # For now we'll advise user that nested keys aren't supported, but it should be
+        # fairly easy to add support if we could add nested keys to valid_keys.
```
tcbegley, line 3779: Should we raise an error here and tell the user that nested keys aren't supported by `LazyStackedTensorDict` yet, or wait for nested keys to be added to `self.valid_keys`?

vmoens: We can go in two phases
(1) raise an issue or add a bullet point in #22 for bookkeeping, and add nested keys to `valid_keys` in a second PR
(2) do it all together

I feel (1) makes more sense, we're in no rush and it's a specific usage

tcbegley:
I've made some progress here, but the changes (as will be clear from the diff) are getting quite complicated. A few things to call out

- Some of the complexity is because it's not trivial to add support for nested keys to `LazyStackedTensorDict` or `SavedTensorDict`.
- `.keys()`, `.values()` and `.items()` all default to returning nested keys (except for the above two types), but that is not true of `.items_meta()` or `.values_meta()`. I'm not sure whether it's more confusing to have the inconsistency or to return nested meta information? Maybe the default of returning nested keys isn't right itself?
- Probably could do with adding a test or two for the iteration over nested keys and items etc, but I wanted to push what I have atm because there were a lot of failing tests to fix and I think I now have the existing ones working.

vmoens:
> .keys(), .values() and .items() all default to returning nested keys (except for the above two types), but that is not true of .items_meta() or .values_meta(). I'm not sure whether it's more confusing to have the inconsistency or to return nested meta information? Maybe the default of returning nested keys isn't right itself?

We can default to non-nested.
What do you think @shagunsodhani? All default to nested or all default to non-nested?

tcbegley:
I've changed the behaviour so that nested keys are never returned by default, you must always opt-in with `include_nested=True`.

To ensure we don't have weirdness like

```python
assert ("a", "b") in tensordict.keys()
assert ("a", "b") not in list(tensordict.keys())
```

(i.e. inconsistent behaviour of `__iter__` and `__contains__`) it is now also necessary to set `include_nested=True` when doing a membership check, i.e.

```python
("a", "b") in tensordict.keys(include_nested=True)
```

I also removed the `include_leaf_tensordicts` keyword argument, as it seemed like it was basically always `True` and added a fair bit of confusion.

PR #39, opened 2022-11-14
[Feature] Support nested keys in select method

tcbegley:
## Description

This PR adds support for nested keys in `TensorDict.select`.

Example:

```python
>>> import torch

>>> td = TensorDict({"nested": TensorDict({"a": torch.rand(1), "b": torch.rand(1)}, [])}, [])
>>> td.select(("nested", "b"))

TensorDict(
    fields={
        nested: TensorDict(
            fields={
                b: Tensor(torch.Size([1]), dtype=torch.float32)},
            batch_size=torch.Size([]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```

## Motivation and Context

cf. #22

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -2530,8 +2530,9 @@ class TensorDict(TensorDictBase):
         return self
 
     def select(self, *keys: NESTED_KEY, inplace: bool = False) -> TensorDictBase:
+        existing_keys = self.keys(include_nested=True)
         for key in keys:
-            if key not in self.keys(include_nested=True):
+            if key not in existing_keys:
                 raise KeyError(f"Key {key} was not found.")
 
         nested_keys = defaultdict(list)
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -2529,18 +2529,43 @@ class TensorDict(TensorDictBase):
             return self.clone()
         return self
 
-    def select(self, *keys: str, inplace: bool = False) -> TensorDictBase:
-        d = {key: value for (key, value) in self.items() if key in keys}
-        d_meta = {
-            key: value
-            for (key, value) in self.items_meta(make_unset=False)
-            if key in keys
-        }
+    def select(self, *keys: NESTED_KEY, inplace: bool = False) -> TensorDictBase:
+        for key in keys:
+            if key not in self.keys(include_nested=True):
+                raise KeyError(f"Key {key} was not found.")
+
+        nested_keys = defaultdict(list)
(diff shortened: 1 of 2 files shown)


tensordict/tensordict.py
```diff
@@ -2529,18 +2529,43 @@ def contiguous(self) -> TensorDictBase:
...
-    def select(self, *keys: str, inplace: bool = False) -> TensorDictBase:
-        d = {key: value for (key, value) in self.items() if key in keys}
-        d_meta = {
-            key: value
-            for (key, value) in self.items_meta(make_unset=False)
-            if key in keys
-        }
+    def select(self, *keys: NESTED_KEY, inplace: bool = False) -> TensorDictBase:
+        for key in keys:
+            if key not in self.keys(include_nested=True):
```
vmoens, line 2534: I would suggest to minimize the creation of the key object, as it's likely to be quite expensive. 
Why not
```python
        existing_keys = set(self.keys(include_nested=True))
        for key in keys:
            if key not in existing_keys:
                raise KeyError(f"Key {key} was not found.")
```

tensordict/tensordict.py
```diff
@@ -3855,6 +3880,14 @@ def _update_valid_keys(self) -> None:
         self._valid_keys = sorted(valid_keys)
 
     def select(self, *keys: str, inplace: bool = False) -> LazyStackedTensorDict:
+        # TODO: Add support for nested keys.
+        for key in keys:
+            if not isinstance(key, str):
```
vmoens, line 3885: I guess we'll need to make this work later, do you want to keep it for a second PR?

vmoens (commented):
Let me know what you think of these 2 comments

tensordict/tensordict.py
tcbegley: Yes, I think let's leave it for a second PR, though I think that can come pretty quick once your changes to `LazyStackedTensorDict` get merged.

tensordict/tensordict.py
tcbegley: I've added this without the `set` because otherwise we run into problems in the recursion with tuples of length 1 not being found in the set of keys (because in `_TensorDictKeysView.__iter__` they are automatically swapped out for strings.

I think this is fine because the `__contains__` method is not generating all keys each time, it's doing a series of O(1) lookups of keys so still efficient, and `__contains__` handles tuples of length 1 automatically for us.

vmoens (approved):
LGTM!

PR #40, opened 2022-11-14
[BugFix] tensordict.set(nested_key, value) points to the wrong metadata dict

vmoens:
## Description

Fixes a bug where the following code snippet tries to access the wrong meta-data dictionary

```python
td = TensorDict({'a': {'a': [1]}}, [])
print(list(td.items_meta()))
td.set(('a', 'a'), [2])
```


tcbegley (approved):
Ah good catch, thanks!

PR #41, opened 2022-11-14
[Feature] Support nested keys in exclude method

tcbegley:
## Description

**Note** This branch is currently targeting `nested-select` branch as it contains those changes also. #

This PR adds support for nested keys in `TensorDictBase.exclude()`. In the case of `LazyStackedTensorDict` we raise an exception if the keys aren't all strings until support for nested keys is added there more broadly.

## Motivation and Context

cf. #22

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -950,7 +950,10 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         raise NotImplementedError(f"{self.__class__.__name__}")
 
     def exclude(self, *keys: str, inplace: bool = False) -> TensorDictBase:
-        keys = [key for key in self.keys() if key not in keys]
+        include_nested = any(isinstance(key, tuple) for key in keys)
+        keys = [
+            key for key in self.keys(include_nested=include_nested) if key not in keys
+        ]
         return self.select(*keys, inplace=inplace)
 
     @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


tensordict/tensordict.py
```diff
@@ -941,7 +941,10 @@ def select(self, *keys: str, inplace: bool = False) -> TensorDictBase:
         raise NotImplementedError(f"{self.__class__.__name__}")
 
     def exclude(self, *keys: str, inplace: bool = False) -> TensorDictBase:
-        keys = [key for key in self.keys() if key not in keys]
+        include_nested = any(isinstance(key, tuple) for key in keys)
+        keys = [
+            key for key in self.keys(include_nested=include_nested) if key not in keys
+        ]
```
vmoens, line 956: I'm curious
`select` will fail if a key does not exist in the tensordict, but exclude won't right?
Do you think it's fine or should we also raise an exception in that case?

vmoens (approved):
great stuff!

tensordict/tensordict.py
tcbegley: Good question. I think I would probably argue for no error.

1) Semantically the behaviour `tensordict.exclude("bad_key")` should mean "I don't want the result to have the key "bad_key". If bad_key doesn't exist in the tensordict anyway, then this is consistent. On the other hand `tensordict.select("bad_key")` should mean "I want a tensordict containing the key "bad_key". If "bad_key" doesn't exist then this can't be fulfilled and shouldn't fail silently.

2) It makes sense to me that these methods would be idempotent. I.e.

    ```python
    tensordict.select("key") == tensordict.select("key").select("key")
    tensordict.exclude("key") == tensordict.exclude("key").exclude("key")
    ```
    The latter won't work if you raise an error when the key has already been removed. If someone was being conservative and excluding certain keys at different places in code this could cause problems.


A compromise could be to add a flag to raise an error if keys are not present?

vmoens: I'm good with the current implementation :)

vmoens (approved):
Wonderful thanks for this! Very elegant!

PR #44, opened 2022-11-15
[Feature] non-strict select

vmoens:
## Description

Implements `tensordict.select(*keys, strict=False)` that allows for keys that are not present to be passed to select (in which case they are ignored).

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -931,7 +931,9 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         raise NotImplementedError(f"{self.__class__.__name__}")
 
     @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


tensordict/tensordict.py
```diff
@@ -3937,7 +3950,9 @@ def _update_valid_keys(self) -> None:
             valid_keys = valid_keys.intersection(td.keys())
         self._valid_keys = sorted(valid_keys)
 
-    def select(self, *keys: str, inplace: bool = False) -> LazyStackedTensorDict:
+    def select(
+        self, *keys: str, inplace: bool = False, strict: bool = None
```
tcbegley, line 3954: Should this be

```suggestion
        self, *keys: str, inplace: bool = False, strict: bool = False
```

tcbegley (approved):
LGTM

PR #45, opened 2022-11-15
[Feature] Nested keys support for set_default

tcbegley:
## Description

Add support for nested keys to `.set_default` method.

Note that currently given a `LazyStackedTensorDict`, `set_default` will work if passed an existing key, but will fail when falling back to `set` which raises an error. Would it be better to raise an error earlier and disallow use of nested keys in `set_default` entirely?

## Context

cf. #22

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1884,7 +1884,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         self._is_locked = value
 
     def set_default(
-        self, key: str, item: COMPATIBLE_TYPES, inplace: bool = False, **kwargs
+        self, key: NESTED_KEY, item: COMPATIBLE_TYPES, inplace: bool = False, **kwargs
     ) -> COMPATIBLE_TYPES:
         """Returns the value of the key if the key is in the tensordict. If not, insert key with a value of item and returns item.
 
@@ -1899,7 +1899,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
             the value of the key or item if the key is not in the tensordict
 
         """
-        if key in self.keys():
+        if key in self.keys(include_nested=isinstance(key, tuple)):
             return self.get(key)
         else:
             self.set(key, item, inplace=inplace, **kwargs)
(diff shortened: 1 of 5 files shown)


test/test_tensordict.py
```diff
@@ -2488,6 +2488,21 @@ def test_setitem_nested():
     assert (tensordict["a", "b", "c"] == 1).all()
 
 
+def test_setdefault_nested():
```
vmoens, line 2491: looks good but do you think we can test that for the other tensordict classes too?

tcbegley: Good idea, have moved the test up into the main test class and added nested items to test this out on the various different types.

tensordict/tensordict.py
```diff
@@ -4391,6 +4373,27 @@ def set(
         self._save(td)
         return self
 
+    def set_default(
+        self, key: NESTED_KEY, item: COMPATIBLE_TYPES, inplace: bool = False, **kwargs
+    ) -> COMPATIBLE_TYPES:
+        """Returns the value of the key if the key is in the tensordict. If not, insert
```
vmoens, line 4379: ideally the first line should not be split, ie

"""Returns the value of the key if the key is in the tensordict. If not, insert key with a value of item and returns item.

vmoens: I guess it should be `inserts` with an s?

vmoens (approved):
LGTM just a minor comment in the docstring

tensordict/tensordict.py
tcbegley: Ah yes, it does read a bit funny. I just copied it from `set_default`, but let me take a pass at rewriting it.

tensordict/tensordict.py
```diff
@@ -1885,26 +1885,25 @@ def is_locked(self, value: bool):
...
 
         """
-        if key in self.keys():
-            return self.get(key)
-        else:
-            self.set(key, item, inplace=inplace, **kwargs)
-            return item
+        if key not in self.keys(include_nested=isinstance(key, tuple)):
+            self.set(key, default, **kwargs)
+        return self.get(key)
```
tcbegley, line 1906: I slightly rewrote this. Because it's possible that `default` changes shape when added to the tensordict, I think it's better to return the value from the tensordict rather than just returning the value that was passed in. wdyt?

tensordict/tensordict.py
tcbegley: Have amended the docstring, taking inspiration from `dict.setdefault`.

PR #51, opened 2022-11-15
[BugFix] Nested tensordict collision

### LINKED ISSUE #29: [BUG] Conflicting keys silently collide in `tensordict.unflatten_keys`
## Describe the bug

If two keys have the same resulting index in `flatten_keys` and `unflatten_keys`, they will be overwritten without an error being raised. This may lead to unexpected behaviours down the line.

## To Reproduce

We would expect this code to fail:
```python
import torch
t = TensorDict({'a.a': torch.randn(3), 'a': {'a': torch.randn(3)}}, [])
print(t.unflatten_keys('.'))
```
as `("a", "a")` already exists.
Similarly, 
```
import torch
t = TensorDict({'a.a': torch.randn(3), 'a': {'a': torch.randn(3)}}, [])
print(t.flatten_keys('.'))
```
should also fail as `"a.a"` is also an existing key.

khundman:
## Description
 
Added checks for key collisions in `flatten_keys()` and `unflatten_keys()` functions with accompanying tests. Also removed a previous test that is no longer relevant now that errors are being thrown for collisions. 
 
## Motivation and Context
 
Addresses #29 - silent key collisions when flattening or unflattening a tensordict are undesirable as the may cause unexpected behaviors.  Collisions caused by flattening or unflattening will now throw errors indicating a collision has occurred so they can be explicitly handled.
 
(close #29)
 
## Types of changes
 
## Test Plan
 
passing all `pre-commit` tests:
image
 
passing new test cases:
image
 
No errors in full suite of tests:
image
 
## Checklist
 

test/test_tensordict.py
```diff
@@ -2829,13 +2829,57 @@ def test_update_nested_dict():
...
+        _ = td3.flatten_keys(separator)
+
+        _ = td1.unflatten_keys(separator)
+        _ = td2.unflatten_keys(separator)
+        _ = td3.unflatten_keys(separator)
+        _ = td4.unflatten_keys(separator)
+        _ = td5.unflatten_keys(separator)
+
+    _ = td4.flatten_keys(separator)
+    _ = td5.flatten_keys(separator)
```
vmoens, line 2882: This should raise an error no? `{"a": {"b" tensor}}` should collide with `{"a": {"b": {"c": tensor}}}`

tensordict/tensordict.py
```diff
@@ -1626,9 +1627,26 @@ def flatten_keys(
...
         for key, meta_value in self.items_meta():
             if meta_value.is_tensordict():
                 to_flatten.append(key)
+            else:
+                already_flat.append(key)
+
+        for key in already_flat:
+            if separator in key:
+                _existing_val = None
+                try:
```
vmoens, line 1640: try/except are expensive if the except is reached (they're very cheap otherwise).
https://stackoverflow.com/a/2522013/4858862
Since we just 'pass' in the except it might occur that this operation is repeated very frequently in the code, thereby affecting performance.

test/test_tensordict.py
```diff
@@ -2829,13 +2829,57 @@ def test_update_nested_dict():
...
+        [],
+    )
+
+    td5 = TensorDict(
+        {f"a{separator}b": torch.zeros(3), "a": {"b": {"c": torch.zeros(3)}}}, []
     )
-    ref = TensorDict({"a": [1, 2], "c": TensorDict({"a": [1, 2], "b": [1, 2]}, [])}, [])
-    assert assert_allclose_td(td.unflatten_keys(separator), ref)
+
+    with pytest.raises(Exception):
```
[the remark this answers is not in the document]
vmoens: Also let's use a more specific error type, and match the error message in the `raises` decorator.

test/test_tensordict.py
khundman: If my understanding is correct, as long as they flatten to different keys there's no collision, e.g. `td4` and `td5`  flatten to:

```
td4 = {
    "a.b.c.d": torch.zeros(3),
    "a.b.c": torch.zeros(3)
}

td5 = {
    "a.b": torch.zeros(3),
    "a.b.c": torch.zeros(3)
}
```

However, unflattening these cause a collision because the `c` key in `td4` has multiple values (`d` nested key and `torch.zeros(3)`. Same for the `b` key in `td5`.

test/test_tensordict.py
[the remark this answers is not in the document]
khundman: Changed this to `KeyError` to be consistent with https://github.com/pytorch-labs/tensordict/blob/main/tensordict/tensordict.py#L2332

test/test_tensordict.py
vmoens: Right! Flatten works, unflatten shouldn't!

test/test_tensordict.py
```diff
@@ -2926,13 +2926,75 @@ def test_update_nested_dict():
...
+            "a": {"b": {"c": torch.zeros(3)}},
+        },
+        [],
+    )
+
+    td5 = TensorDict(
+        {f"a{separator}b": torch.zeros(3), "a": {"b": {"c": torch.zeros(3)}}}, []
+    )
+
+    with pytest.raises(KeyError):
```
vmoens, line 2967: can we add a `raises(KeyError, match="error message")` here?

tensordict/tensordict.py
```diff
@@ -1617,9 +1617,28 @@ def flatten_keys(
...
     ) -> TensorDictBase:
         to_flatten = []
+        already_flat = []
         for key, meta_value in self.items_meta():
             if meta_value.is_tensordict():
                 to_flatten.append(key)
+            else:
+                already_flat.append(key)
+
+        def exists(obj, chain):
```
vmoens, line 1627: Can we move this out of the method, for it not to be re-defined at every call?

test/test_tensordict.py
khundman: Ok, then I think having the checks for flattening of `td4` and `td5` outside of a `pytest.raises` decorator is correct as is (we don't want these to raise errors).

tensordict/tensordict.py
```diff
@@ -1669,6 +1688,12 @@ def unflatten_keys(
             out = self
 
         for key, list_of_keys in to_unflatten.items():
+
+            if key in out.to_dict():
```
shagunsodhani, line 1692: isnt `out` already a TD? If yes, we can check for membership without mapping it back to `dict` no ?

vmoens: Indeed, `to_dict` is an expensive operation. I see there is one here and one above. 
In this case I would simply do
```python
        keys = set(self. Keys()) # pre-compute the keys
        for key, list_of_keys in to_unflatten.items():

            if key in keys:
                ...
```

shagunsodhani: Curious if even `keys = set(self. Keys())` is also needed, as keys in the dict are unique

tensordict/tensordict.py
```diff
@@ -1617,9 +1617,28 @@ def flatten_keys(
...
+                already_flat.append(key)
+
+        def exists(obj, chain):
+            _key = chain.pop(0)
+            if isinstance(obj, Tensor):
+                return None
+            if _key in obj:
+                return exists(obj[_key], chain) if chain else obj[_key]
+
+        for key in already_flat:
```
vmoens, line 1634: why not just

```
existing_keys = self.keys(True)  # returns a set of keys with nested keys: e.g. `("a", ("a", "b"), "c")`
key_split = key. Split(sep)
if sep in key and key_split in existing_keys and not self._get_meta(key_split, TensorDictBase):  # if a key is to be separated and its split version is already there, raise an error
    raise KeyError(...)
```

tensordict/tensordict.py
```diff
@@ -1617,9 +1617,28 @@ def flatten_keys(
...
+            if separator in key:
+                _existing_val = exists(self.to_dict(), key.split(separator))
+
+                if isinstance(_existing_val, Tensor):
```
vmoens, line 1638: Careful here: a memmap-tensor is a tensor like (it behaves like a tensor) but it does not inherit from it.
I'd rather say `if not isinstance(_existing_val, TensorDictBase)`

tensordict/tensordict.py
khundman: changed `.to_dict()` to `.keys()` in both of these instances

vmoens: > Curious if even `keys = set(self. Keys())` is also needed, as keys in the dict are unique

It's not for the uniqueness but because querying the key does not come for free and we don't want to do it several times in a loop. A set is preferable than a list as membership is faster to compute (I think?)

tensordict/tensordict.py
```diff
@@ -1620,6 +1620,12 @@ def flatten_keys(
         for key, meta_value in self.items_meta():
             if meta_value.is_tensordict():
                 to_flatten.append(key)
+            elif separator in key and isinstance(
+                _exists(self, key.split(separator)), (MemmapTensor, Tensor)
```
vmoens, line 1624: I don't think we need exist 
We could just ask if key is in TensorDict.keys(True) which will return all the keys, including nested ones.

khundman: Working on this...thanks for pointing out the utility for returning nested keys.

The test to make sure below works with `flatten_keys()` is failing with the new approach:

```python
td5 = TensorDict(
        {f"a{separator}b": torch.zeros(3), "a": {"b": {"c": torch.zeros(3)}}}, []
     )
```

vmoens: Can you push this with the failing test?

khundman: Pushed. Here's what's happening:
```python
breaking key: a, b
breaking key split: ('a', 'b')
all keys (including nested): ['a-b', ('a', 'b', 'c'), ('a', 'b'), 'a']
```
The recursive fashion in which the nested keys are included causes the flattened key to match and throws and error for this situation.

tensordict/tensordict.py
khundman: > It's not for the uniqueness but because querying the key does not come for free and we don't want to do it several times in a loop. A set is preferable than a list as membership is faster to compute (I think?)

makes sense, I'll update this

tensordict/tensordict.py
vmoens: I tried something, let me know if you think it's appropriate

khundman: Looks good, thanks

vmoens (approved):
LGTM! Thanks for this!

PR #52, opened 2022-11-16
[Test] Test keys, items and values ordering

vmoens:
## Description

The possible discrepancy between the ordering of keys, items and values keeps me awake at night.
This PR will hopefully put me to sleep.


vmoens:
@tcbegley those tests may pass even if the order is not guarantee (I guess there is only a small chance but who knows)
Do you think we could improve that?

Issue #28, opened 2022-11-08
[Feature Request] tensordict.split


vmoens:
## Motivation

It would be nice to have a `tensordict.split` method that would work as [`tensor.split`](https://pytorch.org/docs/stable/generated/torch.split.html).
We would also need to overload `torch.split` as we do for `torch.stack` if possible.

Example:
```python
tensordict = TensorDict({}. [10])
tensordict.split(5, 0)  # results in 2 tensordict of batch_size [5] and [5]
tensordict.split([5, 3, 2], 0)  # results in 3 tensordict of batch_size [5], [3] and [2]
torch.split(tensordict, [5, 3, 2], 0)  # same as above
```

PR #56, opened 2022-11-18
[Feature] `TensorDict.split` method

wonnor-pro:
## Description

- Implement `tensordict.split` method that works like [tensor.split](https://pytorch.org/docs/stable/generated/torch.split.html).
- Overload `torch.split` to support tensordict.
- Add test cases in `test_tensordict.py`

## Motivation and Context

As requested in issue #28, It would be nice to have a tensordict.split method that would work as [tensor.split](https://pytorch.org/docs/stable/generated/torch.split.html).
We would also need to overload torch.split as we do for torch.stack if possible.

## Types of changes

## Checklist

My changes involve a new function in API, the docstring is added to the function. The auto-generated file is not included in the commit.

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1487,7 +1487,9 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
             for idx, split_tensor in enumerate(split_tensors):
                 dictionaries[idx][key] = split_tensor
         return [
-            TensorDict(dictionaries[i], batch_sizes[i], device=self.device)
+            TensorDict(
+                dictionaries[i], batch_sizes[i], device=self.device, _run_checks=False
+            )
             for i in range(len(dictionaries))
         ]
 
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -3151,6 +3151,11 @@ def pad_sequence_td(
         return out
 
 
+@implements_for_td(torch.split)
+def _split(td: TensorDict, split_size_or_sections: Union[int, List[int]], dim: int = 0):
+    return td.split(split_size_or_sections, dim)
+
+
 class SubTensorDict(TensorDictBase):
     """A TensorDict that only sees an index of the stored tensors.
 
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1443,10 +1443,12 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         batch_sizes = []
         if self.batch_dims == 0:
             raise RuntimeError("TensorDict with empty batch size is not splittable")
-        if dim >= self.batch_dims:
+        if not (-self.batch_dims <= dim < self.batch_dims):
             raise IndexError(
                 f"Dimension out of range (expected to be in range of [-{self.batch_dims}, {self.batch_dims - 1}], but got {dim})"
             )
+        if dim < 0:
+            dim += self.batch_dims
         if isinstance(split_size, int):
             rep, remainder = divmod(self.batch_size[dim], split_size)
             rep_shape = [
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1440,7 +1440,6 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
             A list of TensorDict with specified size in given dimension.
 
         """
-        original_batch_size = self.batch_size
         batch_sizes = []
         if self.batch_dims == 0:
             raise RuntimeError("TensorDict with empty batch size is not splittable")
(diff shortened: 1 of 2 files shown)


tensordict/tensordict.py
```diff
@@ -1424,6 +1424,36 @@ def reshape(
             batch_size = shape
         return TensorDict(d, batch_size, device=self.device)
 
+    def split(
+        self, split_size: Union[int, List[int]], dim: int = 0
+    ) -> List[TensorDictBase]:
+        """Splits each tensor in the TensorDict with the specified size in the given dimension, like `torch.split`.
+        Returns a list of TensorDict with the view of split chunks of tensor.
```
vmoens, line 1431: I would add an empty line before this

tensordict/tensordict.py
```diff
@@ -1424,6 +1424,36 @@ def reshape(
...
+
+        The list of TensorDict maintains the original order of the tensor chunks.
+
+        Args:
+            split_size (int) or (list(int)): size of a single chunk or list of sizes for each chunk
```
vmoens, line 1436: `split_size (int of List[int]): ...`

tensordict/tensordict.py
```diff
@@ -1424,6 +1424,36 @@ def reshape(
...
+            A list of TensorDict with specified size in given dimension.
+
+        """
+        dictionaries = []
+        batch_size = []
+        for key, item in self.items():
+            split_tensors = torch.split(item, split_size, dim)
+            if not dictionaries:
+                dictionaries = [dict() for _ in range(len(split_tensors))]
+                batch_size = [split_tensor.shape[:-1] for split_tensor in split_tensors]
```
vmoens, line 1449: I'm a bit confused by the batch size

If you have this case

```
t = TensorDict({}, [3, 10, 5])
t.split([3, 5, 2], 1)
```
you would get 3 (empty) tensordicts of batch size `[3, 3, 5]`, `[3, 5, 5]` and`[3, 2, 5]`.
here that would not be the result i think

test/test_tensordict.py
```diff
@@ -2935,6 +2935,52 @@ def test_unflatten_keys_collision(separator):
...
+    assert tds[1].shape == torch.Size([5])
+
+    tds = td.split([5, 2, 3], 0)
+    assert len(tds) == 3
+    assert tds[0].shape == torch.Size([5])
+    assert tds[1].shape == torch.Size([2])
+    assert tds[2].shape == torch.Size([3])
+
+    td = TensorDict(
+        source={"a": torch.zeros(5, 6, 5), "b": torch.zeros(5, 6, 10, 1)},
```
vmoens, line 2953: I guess that if you swap the values of a and b, since you take the shape of the first tensor up to the last-but-one dimension, you'll end up with a batch size that is not accurate no?

test/test_tensordict.py
```diff
@@ -2935,6 +2935,52 @@ def test_unflatten_keys_collision(separator):
     assert assert_allclose_td(td.unflatten_keys(separator), ref)
 
 
+def test_split():
```
vmoens, line 2938: Another version of this test would be to test with this setting:

Put the following tensordicts in a `pytest.mark.parametrize`:
```
    td = TensorDict({}, [10])
    td = TensorDict({"a": torch.zeros(10)}, [10])
    td = TensorDict({"a": torch.zeros(10, 5, 3), "b": torch.zeros(10, 5, 3, 4)}, [10, 5, 3])
```
and have a test where we do
```
for d in range(td.batch_dims):
    rep = td.shape[d] // 2
    split_list = [2] * rep + [td.shape[d] % 2]  % should return [2, 1] for size 3 and [2, 2, 1] for size 5
    for splits in (2, split_list):
        tds = td.split(splits, d)
```

etc

What do you think?

vmoens (commented):
Let's have a look at the computation of the batch size :)

test/test_tensordict.py
wonnor-pro: I agree, I'll take the shape of the first tensor up to the batch-size length to ensure the batch size reflects the first few common dimensions.

tensordict/tensordict.py
wonnor-pro: The `batch_size` and the first dimensions of each of the tensors must be compliant. I previously did not consider this as legal inputs. I'll add the edge cases to the split method. On a side note, can you share more context in how empty tensordict with batch size will be used? Thanks!

tensordict/tensordict.py
wonnor-pro: I'll update the type format of args `split_size(int or List(int))`.

test/test_tensordict.py
wonnor-pro: Yeah I considered use `pytest.mark.parametrize` as well. 
I think pros are that it will make the test case reflect more of the options we provide in split_size (int or list(int)), but it would also make the test case less expressive and affect the readability, which is important in test files. What do you think?

tensordict/tensordict.py
vmoens: I would not compute the batch size of the tensordicts based on the size of the tensors
`TensorDict({}, [])` is not splittable
`TensorDict({"a": torch.randn(3)}, [])` is not splittable either (because you can only split along the batch size)
`TensorDict({}, [3, 4])` is splittable (an empty tensordict is like saying "everything you put in this tensordict will have a shape of `[3, 4, *other]`

test/test_tensordict.py
vmoens: Beware of the fact that the batch size is not necessarily the shape up to the last common dimension. It can be anything in between 0 and that.

test/test_tensordict.py
wonnor-pro: As sync-ed up offline, the test function will be moved to TestTensorDict class for wider coverage of test cases. Thanks for the feedback!

tensordict/tensordict.py
wonnor-pro: The support to empty TensorDict is now implemented. Thanks for the feedback!

wonnor-pro:
I am currently investigating inappropriate behaviour when `dim` is negative in some cases.

wonnor-pro:
The inappropriate behaviours for negative `dim` have been solved.

vmoens (approved):
Great work!

PR #58, opened 2022-11-19
[BugFix] Fix locking mechanism

vmoens:
## Description

Fixes the locking mechanism for nested tds

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -2548,6 +2548,9 @@ class TensorDict(TensorDictBase):
                 "memmap is not compatible with gradients, one of Tensors has requires_grad equals True"
             )
         for key, value in self.items():
+            if isinstance(value, TensorDictBase):
+                self._tensordict[key] = value.memmap_()
+                continue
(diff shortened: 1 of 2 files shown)


tensordict/tensordict.py
```diff
@@ -1938,10 +1941,16 @@ def set_default(
         return self.get(key)
 
     def lock(self):
-        self.is_locked = True
+        self._is_locked = True
```
shagunsodhani, line 1944: This looks cool! Maybe I missed this in the docs, but what does locking do ?

vmoens: It's not well documented :( 
The idea is to prevent writing new tensors in a tensordict.
If you put your TD in shared memory, we don't have a mechanism to update the tensors it contains in every process. Hence we lock it, such that the keys are fixed and can't be changed.
You can update it in-place though, simply not add or remove keys.

shagunsodhani: Got it - thanks for the clarification :)

PR #62, opened 2022-11-21
[Feature] In-place functionalisation docstrings + cleanup

tcbegley:
## Description

This PR updates examples in docstrings for the new in-place functionalisation API and also deprecates the `make_functional_with_buffers` methods on probabilistic and sequential tensordict modules (these would break anyway since they call the `make_functional_with_buffers` method of the underlying modules which has already been deprecated.


vmoens (approved):
Great stuff thanks!

PR #68, opened 2022-11-23
[BugFix] Optional functorch dependency

vmoens:
## Description

Ensures that dependency on functoch is only optional for code and test execution

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -10,16 +10,7 @@ from copy import deepcopy
 from textwrap import indent
 from typing import Any, Iterable, List, Optional, Sequence, Union
 
-try:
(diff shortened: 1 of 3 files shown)


tcbegley (approved):
LGTM!

shagunsodhani:
I am curious why this is needed given that functorch ships with torch now. Is this for the previous versions ?

vmoens:
> I am curious why this is needed given that functorch ships with torch now. Is this for the previous versions ?

Yes, people that are using Polymetis are bound to use torch 1.10, which does not support functorch.
Hence we make it possible to use the library without it.
The kind of thing that makes our life much easier and the code very neat 🫠

PR #74, opened 2022-11-24
[BugFix] Mutable default arguments

vmoens:
## Description

Solves an issue with python function definition in loops where the loop variable is overwritten at each iteration in already defined functions.
More info: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

### DIFF
--- a/tensordict/nn/functional_modules.py
+++ b/tensordict/nn/functional_modules.py
@@ -101,7 +101,7 @@ of dimensionality {arg.dim()} so expected in_dim to satisfy
             arg
             if in_dim is None
             else arg.apply(
-                lambda _arg: _add_batch_dim(_arg, in_dim, vmap_level),  # noqa: B023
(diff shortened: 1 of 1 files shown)


tensordict/nn/functional_modules.py
```diff
@@ -101,7 +101,7 @@ def _create_batched_inputs(flat_in_dims, flat_args, vmap_level: int, args_spec):
             arg
             if in_dim is None
             else arg.apply(
-                lambda _arg: _add_batch_dim(_arg, in_dim, vmap_level),  # noqa: B023
+                lambda _arg, in_dim=in_dim: _add_batch_dim(_arg, in_dim, vmap_level),
```
tcbegley, line 104: Why does this work? Isn't `in_dim` still going to be a reference to the same mutable object?

tcbegley (approved):
LGTM and I learnt something!

PR #75, opened 2022-11-24
[Formatting] Minor formatting improvements

vmoens:
## Description

Cleaning up formatting

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -177,7 +177,10 @@ class _TensorDictKeysView:
         return key
 
     def __len__(self):
-        return len(list(self))
+        i = 0
+        for _ in self:
+            i += 1
+        return i
 
     def _items(self, tensordict=None):
         if tensordict is None:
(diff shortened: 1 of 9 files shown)


tcbegley (dismissed):
Nice! One of the changes needs to be reverted to fix the recursion error, see inline comment.

tensordict/tensordict.py
```diff
@@ -177,7 +177,7 @@ def _combine_keys(self, prefix, key):
         return key
 
     def __len__(self):
-        return len([key for key in self])
+        return len(list(self))
```
tcbegley, line 180: I think your errors are coming from this line. I remember running into the same issue, it's because casting something to a `list` calls `__len__` first which leads to a recursion error. The list comprehension doesn't need `__len__`, just calls `__iter__`.

Perhaps we revert this change and add `# noqa` plus an explanation.

vmoens: there's a better option (IMO)
```python
    def __len__(self):
        i = 0
        for _ in self:
            i += 1
        return i
```

like this we're not creating a list for the sole purpose of looking at its length
What do you think

tcbegley: Nice, I agree, this is better.

PR #81, opened 2022-11-25
[BugFix] TensorDictSequential inheritance fix

tcbegley:
## Description

By creating a return value using `self.__class__`, child classes inheriting from `TensorDictSequential` will return an instance of the child type rather than `TensorDictSequential` without us having to redefine `select_subsequence`.

cf. pytorch/rl#700


vmoens (approved):
LGTM!

PR #84, opened 2022-11-28
[Feature] Add leaves_only kwarg to keys / values / items methods

tcbegley:
## Description

This PR adds a new `leaves_only` kwarg (default False) to the `.keys`, `.values` and `.items` methods of `TensorDict` and subclasses. When set to True, keys and values corresponding to nested tensordicts are omitted from the iteration (though their children are not).

Example

```python
import torch

td = TensorDict({"a": torch.rand(1), "b": TensorDict({"c": torch.rand(1)}, [])}, [])

assert set(td.keys(include_nested=False, leaves_only=False)) == {"a", "b"}
assert set(td.keys(include_nested=True, leaves_only=False)) == {"a", "b", ("b", "c")}
assert set(td.keys(include_nested=False, leaves_only=True)) == {"a"}
assert set(td.keys(include_nested=True, leaves_only=True)) == {"a", ("b", "c")}
```

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -138,9 +138,12 @@ class _TensorDictKeysView:
         >>> assert set(td.keys()) == {("a", "b"), "c"}
     """
 
-    def __init__(self, tensordict: "TensorDictBase", include_nested: bool):
+    def __init__(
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM
We should add that to torchrl now :)

PR #85, opened 2022-11-28
[Feature] Lazystack insert / append

tcbegley:
## Description

This PR adds `insert` and `append` methods to `LazyStackedTensorDict` corresponding to the analogous methods for `list`.

Under the hood `append` simply calls `insert` with `index = len(self.tensordicts)`. While `my_list.insert(len(my_list), item)` is fractionally slower than `my_list.append(item)`, this is dwarfed by the checks which I think it makes sense to share.

Example:

```python
>>> import torch
>>> td = TensorDict({"a": torch.zeros(4)}, [4])
>>> lstd = torch.stack([td] * 2)
>>> lstd.insert(1, TensorDict({"a": torch.ones(4)}, [4]))
>>> lstd["a"]
tensor([[[0.],
         [0.],
         [0.],
         [0.]],

        [[1.],
         [1.],
         [1.],
         [1.]],

        [[0.],
         [0.],
         [0.],
         [0.]]])
```

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -4512,6 +4512,8 @@ class LazyStackedTensorDict(TensorDictBase):
         N = len(self.tensordicts)
         self._batch_size = self._compute_batch_size(batch_size, self.stack_dim, N)
         self._update_valid_keys()
+        # recreate _dict_meta to ensure shapes of stacked tensors are correct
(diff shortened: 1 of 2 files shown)


vmoens (commented):
LGTM 
Would be nice to check the repr too if you don't mind

vmoens (commented):
Let's add docstrings too!

vmoens (approved):
LGTM

PR #87, opened 2022-11-28
Create a TensorDict from kwargs in TensorDictModule.forward()

alexanderlobov:
https://github.com/pytorch/rl/issues/601

## Types of changes

## Checklist

- [x] have updated the tests accordingly (*required for a bug fix or a new feature*).

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -192,6 +192,9 @@ class TensorDictModule(nn.Module):
         tensordict_out: Optional[TensorDictBase] = None,
         **kwargs,
     ) -> TensorDictBase:
+        """When the tensordict parameter is not set, kwargs are used to create
+        an instance of TensorDict.
+        """
         if tensordict is None:
             tensordict = make_tensordict(**kwargs)
             # do not pass used kwargs to _call_module
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -11,7 +11,7 @@ from typing import Any, Iterable, List, Optional, Sequence, Union
 
 import torch
 
-from tensordict.tensordict import TensorDictBase
+from tensordict.tensordict import make_tensordict, TensorDictBase
 from tensordict.utils import _nested_key_type_check, _normalize_key, NESTED_KEY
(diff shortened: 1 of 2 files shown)


tensordict/nn/common.py
```diff
@@ -54,6 +56,17 @@ def _check_all_nested(list_of_keys):
         _nested_key_type_check(key)
 
 
+def dispatch_kwargs(func):
+    @functools.wraps(func)
+    def wrapper(self, tensordict=None, *args, **kwargs):
+        if tensordict is None:
+            tensordict = make_tensordict(**kwargs)
```
vmoens, line 63: We should check that all the in_keys are in the kwargs 
Any extra kwarg should be passed to the module forward

eg in the test 
`forward(a=tensor, c=True)`
Would pass the "c" kwarg to the forward of the module and build a TD with "a".

vmoens (commented):
Thanks for this
I think we can write a couple of extra tests.
+ see my comment on kwargs checks

tensordict/nn/common.py
alexanderlobov: Should we throw an exception of not all in_keys are in the kwargs? TensorDictModule.forward() allow now such a case:

```
    def forward(
        self,
        tensordict: TensorDictBase,
        tensordict_out: Optional[TensorDictBase] = None,
        **kwargs,
    ) -> TensorDictBase:
...
        tensors = tuple(tensordict.get(in_key, None) for in_key in self.in_keys)
```

It uses None as a default value if tensordict does not have a key from `self.in_keys`.

tensordict/nn/common.py
```diff
@@ -186,12 +203,16 @@ def _call_module(
...
         return out
 
+    @dispatch_kwargs
     def forward(
         self,
         tensordict: TensorDictBase,
         tensordict_out: Optional[TensorDictBase] = None,
         **kwargs,
     ) -> TensorDictBase:
+        """When the tensordict parameter is not set, kwargs are used to create
```
vmoens, line 213: This should be on a single line

Issue #102, opened 2022-12-06
[Feature Request] data class like behaviour


vmoens:
## Motivation

Some users have pointed out that data classes offer advantages over dict-like classes:
- predictability: one knows in advance what fields to expect, the code is transparent although less generic
- autocomplete: IDEs can keep track of the attributes through type hinting 
- dedicated behaviour: a class can have a set of dedicated methods that act on its attributes.

## Proposed API

We could have a mixed approach where a data class would build a tensordict and interface with it. Attributes would point to keys. The methods would be shared across classes.

```python
@tensordict_data  # better name?
class MyData:
   float_tensor
   sparse_tensor

   def stuff(self):
       return self.float_tensor + self.sparse_tensor 

data = MyData(float_tensor=a, sparse_tensor=b, batch_size=[3, 4])
data.reshape(-1) # returns a new MyData with shape/ batch_size of [12]
data.batch_size # picks the batch size of the tensordict
data.missing_key # returns a dedicated error since this key is not expected
```

## Challenges
Any such dataclass methods (such as reshape, split etc) should return a new instance of the same class. 
Upon creation (at loading time) one should check that the data class attributes do not collide with the tensordict ones.
Nesting with data classes will be slightly harder than with tensordict (see [this thread](https://www.linkedin.com/feed/update/urn:li:activity:7003878068733779968?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A7003878068733779968%2C7004083519140323328%29))

roccajoseph:
I started working on it.

tcbegley:
Resolved by #107

PR #107, opened 2022-12-09
[Feature] Prototype of dataclass like interface for TensorDict

roccajoseph:
## Description

Prototype for a dataclass like interface for TensorDict.

## Motivation and Context

Some users have pointed out that data classes offer advantages over dict-like classes:
- predictability: one knows in advance what fields to expect, the code is transparent although less generic
- autocomplete: IDEs can keep track of the attributes through type hinting
- dedicated behaviour: a class can have a set of dedicated methods that act on its attributes.

These kinds of data structures are easier to handle in production settings or within apps, although less flexible.

See [issue #102](https://github.com/pytorch-labs/tensordict/issues/102) for mode details.

## Types of changes

## Checklist


vmoens:
cc @tcbegley check this out!

you can run this to get a sense of what it does
```python
import torch
from tensordict.prototype.tensordictclass import tensordictclass
@tensordictclass
class MyClass:
    a: torch.Tensor

c = MyClass(batch_size=[3, 4], a=torch.zeros(3, 4, 5))
print(c)
print(c.reshape(-1))
print(c.view(-1))
print(c.permute(1, 0))
print(c.permute(1, 0).contiguous())
print(c[1])
print(c.batch_size)
print(c.device)
print(c.cpu())
print(c[torch.zeros(3, 4, dtype=torch.bool)])
```

tensordict/prototype/tensordictclass.py
```diff
@@ -0,0 +1,68 @@
...
+
+            self.tensordict = TensorDict(
+                {attr: getattr(self, attr, None) for attr in attributes},
+                batch_size=getattr(self, "batch_size", None),
+            )
+
+        def __getattr__(self, attr):
+            if attr in self.__dict__:
+                return getattr(self, attr)
+            else:
```
tcbegley, line 41: `__getattr__` is only called if attempting to access the attribute results in an `Attribute` error, so you can get rid of this if / else block and just have the body of else.

tensordict/prototype/tensordictclass.py
```diff
@@ -0,0 +1,68 @@
...
+                            return res
+
+                    return wrapped_func
+
+        def __getitem__(self, item):
+            res = self.tensordict[item]
+            return TensorDictClass(
+                **res,
+                batch_size=res.batch_size,
+            )  # device=res.device)
```
tcbegley, line 66: Why do we wrap with a class again? In any case I think the `**` isn't right here as the result of `self.tensordict[item]` will generally be a tensor or tensordict, not a dictionary. E.g. this gives me an error


```python
import torch
from tensordict.prototype.tensordictclass import tensordictclass


@tensordictclass
class MyClass:
    a: torch.Tensor


c = MyClass(batch_size=[3, 4], a=torch.zeros(3, 4, 1), )
print(c["a"])
```

I think you just want to return res right?

vmoens: It's more about indexing along dimensions, keys should be prohibited
Eg
c[1]

tcbegley: Ah ok, I didn't realise that `fn(**tensordict)` worked! That's neat. 

So my comment is incorrect, but if I understand correctly we should raise an error if the user supplies a key and direct them to access entries in the tensordict as attributes?

tensordict/prototype/tensordictclass.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, *args, _tensordict=None, **kwargs):
                 assert isinstance(_tensordict, TensorDict)
                 if args or kwargs:
                     raise ValueError("Cannot pass both args/kwargs and _tensordict.")
+                if not (EXPECTED_KEYS - set(_tensordict.keys()) == {"batch_size"}):
```
vmoens, line 60: Say `EXPECTED_KEYS = {'a', 'b', 'c'}` and `set(_tensordict.keys()) = {'a', 'd'}`
then `(EXPECTED_KEYS - set(_tensordict.keys())` is `{'b', 'c'}` but if these keys are optional the real problem should be that there is a `'d'` key that is not expcted

tensordict/prototype/tensordictclass.py
```diff
@@ -180,11 +176,12 @@ def wrapped_func(*args, **kwargs):
                 return wrapped_func
 
         def __getitem__(self, item):
+            if isinstance(item, str) and item in self.tensordict.keys():
```
vmoens, line 183: what if it is a string but not in the tensordict?
Also a tuple of strings should be prohibited too

test/test_tensordictclass.py
```diff
@@ -67,6 +67,24 @@ def test_attributes():
...
+        X=torch.ones(3, 4, 5),
+        y=torch.zeros(3, 4, 5, dtype=torch.bool),
+        batch_size=[3, 4],
+    )
+
+    assert data[:2].batch_size == torch.Size([2, 4])
+    assert data[1].batch_size == torch.Size([4])
+    assert data[1][1].batch_size == torch.Size([])
+
+    with pytest.raises(Exception):
```
vmoens, line 81: let's be more explicit with the exception type and the message (e.g. using `pytest.raises(ValueError, match="etc")`)

test/test_tensordictclass.py
```diff
@@ -96,7 +114,9 @@ def test_reshape():
     y = torch.zeros(3, 4, 5, dtype=torch.bool)
     batch_size = [3, 4]
     data = MyData(X=X, y=y, batch_size=batch_size)
+    print("\na")
```
vmoens, line 117: to be removed

tensordict/prototype/tensordictclass.py
roccajoseph: Yes, I just spotted that also. It breaks the tests, sorry. Pushing a change to revert the errors I introduced.

tensordict/prototype/tensorclass.py
```diff
@@ -0,0 +1,357 @@
...
+                return NotImplemented
+            return TD_HANDLED_FUNCTIONS[func](*args, **kwargs)
+
+        def __getattribute__(self, item):
+            def check_td_out_type(field_def):
+                if PY37:
+                    field_def = str(field_def)
+                if isinstance(field_def, str):
+                    if re.search(r"Optional\[.*\]", field_def):
+                        args = [re.search(r"Optional\[(.*?)\]", field_def).group(1)]
```
tcbegley, line 137: These two searches are basically the same, might be better to do

```python
match = re.search(r"Optional\[(.*?)\]", field_def)
if match is not None:
    args = [match.group(1)]
```

We could also precompile these patterns at global scope say to save a little bit of work each time __getattribute__ is called

```python
# outside __getattribute__
OPTIONAL_PATTERN = re.compile(r"Optional\[(.*?)\]")

match = OPTIONAL_PATTERN.search(field_def)
```

tcbegley (commented):
This is looking great! Few minor comments inline.

tensordict/prototype/tensorclass.py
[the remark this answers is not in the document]
vmoens: It's true that we don't really account for this kind of thing.
Perhaps we could call the re.search recursively until the output is unchanged?
thoughts @roccajoseph?

vmoens: Want to take care of this @roccajoseph ?

roccajoseph: Yes, I'm on it.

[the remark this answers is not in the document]
roccajoseph: Hmm, I would tend to say that it's better to keep it as simple as possible, even if it means to cover a bit less of the cases. Make the code more complicated for very edge cases doesn't sound good to me. But maybe I miss something important here.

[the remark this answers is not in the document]
vmoens: I'm on board with that, let's keep it simple for now.

PR #108, opened 2022-12-09
[BugFix] patched nn.Module deserialization

vmoens:
## Description

Fixes a bug where nn.Module with no forward method fail to be serialized/deserialized if they have been functionalized.
For the record, the error reads:
```
AttributeError: 'ModuleList' object has no attribute '_forward_unimplemented'
```
This is due to the fact that an nn.Module.forward is defined as
```
def _forward_unimplemented(self, *args):
    etc.

class nn.Module():
    forward = _forward_unimplemented
```
We expect this bug to occur every time a forward method is written outside of the module and assigned subsequently.
The current patch only fixes the `_forward_unimplemented`-related error.

cc @tcbegley


tcbegley (approved):
Great work tracking that down! Everything seems to be working for me again with this change.

PR #110, opened 2022-12-09
[Feature] Add NormalParamSplitter

tcbegley:
## Description

This PR adds a new module `NormalParamSplitter` which behaves much like `NormalParamWrapper`, but is intended to be used as a component in a sequence rather than a wrapper around another module.

I haven't yet deprecated the `NormalParamWrapper` so as not to break TorchRL.

This is branching off `deprecate-old-probabilistic-module` as the changes there are quite substantially relevant to these changes, so this shouldn't be merged before #109

### DIFF
--- a/tensordict/nn/distributions/continuous.py
+++ b/tensordict/nn/distributions/continuous.py
@@ -12,7 +12,7 @@ import torch
 from tensordict.nn.utils import mappings
 from torch import distributions as D, nn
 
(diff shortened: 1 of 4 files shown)


tcbegley:
I'm not in love with the name `NormalParamSplitter`, so am very open to hearing alternatives.

vmoens (approved):
LGTM
Let's just iterate a bit over the name of the class

Issue #17, opened 2022-11-04
[Feature Request] Implement `tensordict.pop(key, default)`


vmoens:
## Motivation

`dict.pop()` is a handy feature that is missing from tensordict.

zmopen:
working on this.

tcbegley:
Resolved by #111

PR #111, opened 2022-12-12
[Feature] Add TensorDict.pop()

salaxieb:
## Description

Implement tensordict.pop(key, default) where key can be a nested key.

## Motivation and Context

[Feature Request] Implement tensordict.pop(key, default) https://github.com/pytorch-labs/tensordict/issues/17

## Types of changes

## Checklist


test/test_tensordict.py
```diff
@@ -1757,6 +1757,25 @@ def test_split(self, td_name, device, performer):
...
+        out = td.pop("b", default)
+        assert out != default
+        out = td.pop("z", default)
+        assert out == default
+
+        with pytest.raises(
+            KeyError,
+            match = "You are trying to pop key z which is not in dict and not providing default value"
+        ):
+            td.pop("z")
```
vmoens, line 1779: I agree this will run as expected but it's a bit tricky since we call pop with that key before (requires the reader to relate to the previous test, observe that the key wasn't there and expect this line to behave regardless)
Maybe using a distinct key would be more straightforward?

vmoens (approved):
Neat!
Can you have a look at the nit in the test?

test/test_tensordict.py
salaxieb: Yep, done

test/test_tensordict.py
```diff
@@ -1757,6 +1757,30 @@ def test_split(self, td_name, device, performer):
                                 expected_inner_tensor_size
                             )
 
+    def test_pop(self, td_name, device):
+        td = getattr(self, td_name)(device)
+        assert "a" in td.keys()
+        out = td.pop("a")
+        assert torch.eq(out, getattr(self, td_name)(device)['a'])
```
vmoens, line 1764: this re-creates a whole tensordict from scratch
why not
```
        a = td["a"].clone()
        assert "a" in td.keys()
        out = td.pop("a")
        assert torch.eq(out, a)
```

test/test_tensordict.py
```diff
@@ -1757,6 +1757,30 @@ def test_split(self, td_name, device, performer):
...
+        assert 'a' not in td.keys()
+
+        default = "some value"
+        assert "b" in td.keys()
+        out = td.pop("b", default)
+        assert out != default
```
vmoens, line 1770: `assert tensor` will fail for every tensor that has more than one element
you should do `assert tensor.all()`

test/test_tensordict.py
```diff
@@ -1757,6 +1757,30 @@ def test_split(self, td_name, device, performer):
                                 expected_inner_tensor_size
                             )
 
+    def test_pop(self, td_name, device):
+        td = getattr(self, td_name)(device)
+        assert "a" in td.keys()
+        out = td.pop("a")
+        assert torch.all(torch.eq(out, getattr(self, td_name)(device)["a"]))
```
vmoens, line 1764: Here i'd rather create the tensordict once
```
        a = td["a"].clone()
        assert "a" in td.keys()
        out = td.pop("a")
        assert torch.all(torch.eq(out, a))
```

test/test_tensordict.py
salaxieb: I see now:
ValueError: default should be None or a Tensor instance, got some value

vmoens: ah sorry, nevermind
I saw an error which looked like `assert tensor` but it was for the one above that you just corrected :)

test/test_tensordict.py
```diff
@@ -1757,6 +1757,34 @@ def test_split(self, td_name, device, performer):
...
 
+    def test_pop(self, td_name, device):
+        td = getattr(self, td_name)(device)
+        assert "a" in td.keys()
+        a = td["a"].clone()
+        out = td.pop("a")
+        assert (out == a).all()
+        assert "a" not in td.keys()
+
+        default = torch.Tensor([0])
```
vmoens, line 1768: having a string was cool actually! We can keep that and `assert out != default` as it was before

salaxieb: yes, it was.. but if I use string in get function I get this error:
ValueError: default should be None or a Tensor instance, got some value

vmoens: Ah right, well we could just do 
`default = torch.zeros_like(a)` then

test/test_tensordict.py
```diff
@@ -1757,6 +1757,36 @@ def test_split(self, td_name, device, performer):
...
 
+    def test_pop(self, td_name, device):
+        td = getattr(self, td_name)(device)
+        assert "a" in td.keys()
+        a = td["a"].clone()
+        out = td.pop("a")
+        assert (out == a).all()
+        assert "a" not in td.keys()
+
+        default = torch.zeros_like(a).to(device)
```
vmoens, line 1768: argh maybe we should do this (since a and b don't share the same shape)
```
default = torch.zeros_like(td['b']).to(device)
```

vmoens (approved):
LGTM thanks!

PR #113, opened 2022-12-15
[Refactor] Refactor dispatch_kwargs for easier usage

vmoens:
## Description

Refactors `dispatch_kwargs` for a completely `TensorDict` oblivious behviour.

Allows to call a TDModule, TDSequential or any other module decorated with `dispatch_kwargs` with a set of keyword arguments. The function will return a tuple or tensors or a single tensor, depending on the number of module outputs.

@tcbegley

### DIFF
--- a/docs/source/reference/nn.rst
+++ b/docs/source/reference/nn.rst
@@ -47,6 +47,7 @@ additional input from the tensordict as necessary. Here's an example:
 
 .. code-block::
 
+  >>> from tensordict.nn import TensorDictSequential
   >>> class Net(nn.Module):
   ...     def __init__(self, input_size=100, hidden_size=50, output_size=10):
   ...         super().__init__()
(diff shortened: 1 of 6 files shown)


docs/source/reference/nn.rst
```diff
@@ -3,6 +3,190 @@
 tensordict.nn package
 =====================
 
+The tensordict.nn package makes it possible to flexibly use TensorDict within
```
vmoens, line 6: @tcbegley can you have a quick look at this?
Some parts have been copy-pasted / simplified from out recent post

tensordict/nn/common.py
```diff
@@ -58,14 +59,78 @@ def _check_all_nested(list_of_keys):
...
+        ...         return tensordict
+        ...
+        >>> module = MyModuleNest()
+        >>> b, = module(a_c=torch.zeros(1, 2))
+        >>> assert (b == 1).all()
+
+
+    """
+
+    # sanity check
```
vmoens, line 107: @tcbegley what do you think of this?
One issue with tensordict is that we name the package like we name the objects, it's not ideal.
Besides enforcing that the first arg always has the same name can be confusing.
We could check that the first input is typed as being a TensorDictBase but even that won't solve the problem (we're not very restrictive on types, it would be surprising to do that here no?)

tensordict/nn/common.py
```diff
@@ -58,14 +59,78 @@ def _check_all_nested(list_of_keys):
...
             for key in self.in_keys:
-                if key in kwargs:
-                    tensordict_values[key] = kwargs.pop(key)
-            tensordict = make_tensordict(**tensordict_values)
+                if isinstance(key, tuple):
+                    expected_key = "_".join(key)
+                else:
+                    expected_key = key
+                if expected_key in kwargs:
+                    tensordict_values[key] = kwargs.pop(expected_key)
```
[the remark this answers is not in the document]
vmoens: good point

PR #114, opened 2022-12-15
[Feature] Add metaclass for TensorClass instances

tcbegley:
## Description

This PR creates a new `_TensorClassMeta` metaclass, which checks the dataclass field names against reserved TensorDict attributes / methods when the class is defined, rather than when the first instance is created.

This prevents users from being able to define a class that could never be instantiated without error, and moves the error closer to the source of the problem.

cc: @roccajoseph

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -27,6 +27,23 @@ OPTIONAL_PATTERN = re.compile(r"Optional\[(.*?)\]")
 UNION_PATTERN = re.compile(r"Union\[(.*?)\]")
 
 
+class InvalidAttributeName(ValueError):
+    pass
+
+
+class _TensorClassMeta(type):
+    def __new__(cls, clsname, bases, attrs):
+        datacls = bases[0]
+        for attr in datacls.__dataclass_fields__:
+            if attr == "batch_size":
+                continue
+            if attr in dir(TensorDict):
+                raise InvalidAttributeName(
+                    f"Attribute name {attr} can't be used with @tensorclass"
+                )
+        return super().__new__(cls, clsname, bases, attrs)
+
+
 def tensorclass(cls: T) -> T:
     """A decorator to create :obj:`tensorclass` classes.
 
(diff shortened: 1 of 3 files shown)


tensordict/prototype/tensorclass.py
```diff
@@ -27,6 +27,23 @@
 UNION_PATTERN = re.compile(r"Union\[(.*?)\]")
 
 
+class InvalidAttributeName(ValueError):
+    pass
```
tcbegley, line 31: This might be overkill, but I couldn't see a built-in exception that accurately described the problem of an invalid attribute name. I think `ValueError` is reasonable, because naming attributes in a class definition is equivalent to passing those names as `attrs` argument to some metaclass, but to the user it feels quite different to calling a function with arguments.

tcbegley, line 31: This might be overkill, but I couldn't see a built-in exception that accurately described the problem of an invalid attribute name. I think `ValueError` is reasonable, because naming attributes in a class definition is equivalent to passing those names as `attrs` argument to some metaclass, but to the user it feels quite different to calling a function with arguments.

vmoens: Have you thought about AttributeError (even if it's semantically quite the opposite)?

vmoens (approved):
LGTM

tensordict/prototype/tensorclass.py
vmoens: e.g.
```
>>> class A:
...     pass
...
>>> A.__dict__ = 1
  File "<stdin>", line 1, in <module>
AttributeError: attribute '__dict__' of 'type' objects is not writable
```

tcbegley: Yes, `AttributeError` was the main alternative I considered, though as you say it's more about accessing something that doesn't exist rather than writing over something that does.

Happy to change it if you think that would be better? Even if it's not the conventional usage of AttributeError, it is an error relating to attributes...

And what about the custom exception? Just us a bare AttributeError instead?

PR #117, opened 2022-12-16
[Refactor] Remove prototype import patch

tcbegley:
## Description

Since pytorch/rl#738 has been merged, we can now clean up these patch imports that were left to maintain compatibility.

I've run the TorchRL tests locally against this branch to make sure nothing is broken.


vmoens (approved):
LGTM

Issue #106, opened 2022-12-08
[BUG] Auto-nested tensordict bugs


vmoens:
## Describe the bug

Auto-nesting may be a desirable feature (e.g. to build graphs), but currently it is broken for multiple functions, e.g.
```
tensordict = TensorDict({}, [])
tensordict["self"] = tensordict
print(tensordict)  # fails
tensordict.flatten_keys()  # fails
list(tensordict.keys(include_nested=True))  # fails
```

## Consideration

This is something that should be included in the tests. We could design a special test case in `TestTensorDictsBase` with a nested self.

## Solution

IMO there is not a single solution to this problem. For repr, could find a way of representing a nested tensordict, something like

```
TensorDict(fields={
   "self": ...
}, 
batch_size=[])
```
For keys, we could avoid returning a key if it a key pointing to the same value has already been returned (same for values and items).
For flatten_keys, it should be prohibited for TensorDict. The options are (1) leave it as it is since the maximum recursion already takes care of it or (2) build a wrapper around `flatten_keys()` to detect if the same method (i.e. the same call to the same method from the same class) is occurring twice, something like
```
def detect_self_nesting(fun):
    def new_fun(self, *args, **kwargs):
        if fun in self._being_called:
             raise RuntimeError
        self._being_called.append(fun)
        out = fun(self, *args, **kwargs)
        self._being_called.pop()
        return out
    return new_fun
```

There are probably other properties that I'm missing, but i'd expect them to be covered by the tests if we can design the dedicated test pipeline mentioned earlier.

ruleva1983:
Working to fix this issue

vmoens:
I think there are essentially 3 things to consider:
(1) iteration over keys: iterating over nested keys should lead to an infinite recursion, this is fine. The only thing that should work really is asserting if a (nested) key is in the auto-nested tensordict. For this to work, the `__contains__` should keep track of what is being / has already been explored to avoid infinite recursion and to make sure that we explore every branch. As soon as the `__contains__` reaches a self-nested tensordict it should interrupt the query for that tensordict and pass to the next
(2) tensordict methods that return tensordicts (or better, tensor-to-tensor methods): all methods that return a tensor out of a tensor can be wrapped under a common method IMO. 

Here is an example of how to do it with `to_tensordict(self)` but I guess that a similar, generic solution could be found for `apply`, `__eq__`, `to(device)` etc
```python
        def to_tensordict(
[...]

PR #119, opened 2022-12-16
[BugFix] Allowing for auto-nested tensordict

Zooll:
## Description

Blocked self nested behaviour during __repr__ or any other recursive calls.

## Motivation and Context

[BUG] Auto-nested tensordict bugs #106

## Types of changes

Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -137,6 +137,7 @@ class _TensorDictKeysView:
         self.tensordict = tensordict
         self.include_nested = include_nested
         self.leaves_only = leaves_only
+        self.loop_set = set()
 
     def __iter__(self):
         if not self.include_nested:
@@ -153,12 +154,13 @@ class _TensorDictKeysView:
 
     def _iter_helper(self, tensordict, prefix=None):
         items_iter = self._items(tensordict)
-
+        self.loop_set.add(id(tensordict))
         for key, value in items_iter:
             full_key = self._combine_keys(prefix, key)
             if (
                 isinstance(value, (TensorDictBase, KeyedJaggedTensor))
                 and self.include_nested
+                and id(value) not in self.loop_set
             ):
                 subkeys = tuple(
                     self._iter_helper(
@@ -169,6 +171,7 @@ class _TensorDictKeysView:
                 yield from subkeys
             if not (isinstance(value, TensorDictBase) and self.leaves_only):
                 yield full_key
+        self.loop_set.remove(id(tensordict))
 
     def _combine_keys(self, prefix, key):
         if prefix is not None:
@@ -265,6 +268,8 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
 
     def __init__(self):
         self._dict_meta = KeyDependentDefaultDict(self._make_meta)
+        self._being_called = False
+        self._being_flattened = False
 
     @abc.abstractmethod
     def _make_meta(self, key: str) -> MetaTensor:
@@ -1717,7 +1722,11 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         )
 
     def __repr__(self) -> str:
+        if self._being_called:
+            return "Auto-nested"
+        self._being_called = True
         fields = _td_fields(self)
+        self._being_called = False
         field_str = indent(f"fields={{{fields}}}", 4 * " ")
         batch_size_str = indent(f"batch_size={self.batch_size}", 4 * " ")
         device_str = indent(f"device={self.device}", 4 * " ")
@@ -1791,6 +1800,10 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
     def flatten_keys(
         self, separator: str = ".", inplace: bool = False
     ) -> TensorDictBase:
+        if self._being_flattened:
+            return self
+
+        self._being_flattened = True
         to_flatten = []
         existing_keys = self.keys(include_nested=True)
         for key, meta_value in self.items_meta():
@@ -1815,6 +1828,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
                     self.set(separator.join([key, inner_key]), inner_item)
             for key in to_flatten:
                 del self[key]
+            self._being_flattened = False
             return self
         else:
             tensordict_out = TensorDict(
@@ -1830,10 +1844,14 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
                     inner_tensordict = self.get(key).flatten_keys(
                         separator=separator, inplace=inplace
                     )
-                    for inner_key, inner_item in inner_tensordict.items():
-                        tensordict_out.set(separator.join([key, inner_key]), inner_item)
+                    if inner_tensordict is not self.get(key):
+                        for inner_key, inner_item in inner_tensordict.items():
+                            tensordict_out.set(
+                                separator.join([key, inner_key]), inner_item
+                            )
                 else:
                     tensordict_out.set(key, value)
+            self._being_flattened = False
             return tensordict_out
 
     def unflatten_keys(
@@ -4623,6 +4641,7 @@ class SavedTensorDict(TensorDictBase):
         device: Optional[torch.device] = None,
         batch_size: Optional[Sequence[int]] = None,
     ):
+        super().__init__()
         if not isinstance(source, TensorDictBase):
             raise TypeError(
                 f"Expected source to be a TensorDictBase instance, but got {type(source)} instead."
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -268,7 +268,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
 
     def __init__(self):
         self._dict_meta = KeyDependentDefaultDict(self._make_meta)
-        self._being_called = False
+        self._being_repr = False
         self._being_flattened = False
 
     @abc.abstractmethod
@@ -1722,11 +1722,13 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         )
 
     def __repr__(self) -> str:
-        if self._being_called:
+        if self._being_repr:
             return "Auto-nested"
-        self._being_called = True
-        fields = _td_fields(self)
-        self._being_called = False
+        try:
+            self._being_repr = True
+            fields = _td_fields(self)
+        finally:
+            self._being_repr = False
         field_str = indent(f"fields={{{fields}}}", 4 * " ")
         batch_size_str = indent(f"batch_size={self.batch_size}", 4 * " ")
         device_str = indent(f"device={self.device}", 4 * " ")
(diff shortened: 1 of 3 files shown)


tensordict/tensordict.py
```diff
@@ -889,6 +914,7 @@ def values_meta(
         """
         if make_unset:
             for k in self.keys(include_nested=include_nested):
+                # if self._check_meta(k):
```
tcbegley, line 917: This can presumably be deleted?

tensordict/tensordict.py
```diff
@@ -144,18 +144,20 @@ def __init__(
...
     def _iter_helper(self, tensordict, prefix=None):
         items_iter = self._items(tensordict)
-
+        self.loop_set.add(id(tensordict))
         for key, value in items_iter:
             full_key = self._combine_keys(prefix, key)
             if (
                 isinstance(value, (TensorDictBase, KeyedJaggedTensor))
                 and self.include_nested
+                and id(value) not in self.loop_set
```
tcbegley, line 160: Is there a reason you opted for a set of ids here instead of setting some attribute on the TensorDict as in the case of `__repr__` and `flatten`?

tcbegley (changes requested):
This is looking good, thanks for working on it!

I've left a few comments inline.

The other main thing that we should add are some tests. We can add a couple of tests for specific functionality such as checking for the presence of `"Auto-nested"` in the repr, but as @vmoens suggested in #106 it would be nice to run all tests on a tensordict with autonesting and check that we haven't inadvertently broken something.

Check [this class](https://github.com/pytorch-labs/tensordict/blob/4318760e6f4a6b1f3c2e3936379091cbbe4a3a1e/test/_utils_internal.py#L29) that gets used in the tests, you can add a method for `auto_nested_td` or similar, and then add it to the tests [here](https://github.com/pytorch-labs/tensordict/blob/4318760e6f4a6b1f3c2e3936379091cbbe4a3a1e/test/test_tensordict.py#L602-L618). In principle it should work for anything that `nested_td` works for, but in any case it will be interesting to see what breaks!

tensordict/tensordict.py
```diff
@@ -568,6 +577,22 @@ def pop(
...
+                    return self.get(key[0])._check_meta(key[1:])
+                key = key[0]
+            return key in self._dict_meta
+        except KeyError:
+            # TODO: this error message will make more sense if `.keys()` returns nested
+            # keys rather than just the top level ones
+            raise KeyError(
+                f"key {key} not found in {self.__class__.__name__} with keys"
+                f" {sorted(self.keys(), key=str)}"
+            )
```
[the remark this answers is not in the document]
Zooll: The main problem here is when we call `_get_meta` we won't get `KeyError` because we will get into `__missing__` func and then `_make_meta` will be called. `_make_meta`  will try to create a new `MetaTensor`, but during initialisation `MetaTensor` will call `_requires_grad` and then it will iterate through all of nested objects and try to get their meta data (it will call `_get_meta` again) and here is the loop. I tried to find a simple way to stop this loop, and check option without [] operator fitted very well

tensordict/tensordict.py
```diff
@@ -301,7 +306,11 @@ def size(self, dim: Optional[int] = None):
 
     @property
     def requires_grad(self):
-        return any(v.requires_grad for v in self.values_meta())
+        for k in self.keys():
+            if self._check_meta(k):
```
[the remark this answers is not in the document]
Zooll: Oh, here is more complex case. Because I faced with different behaviour classical Tensordict and unsqueeze Tensordict (I realized it during tests running).  Some times it is ok not to have values in you `_dict_meta`, as I understand because of some lazy operations. `self.values_meta()` is always call `_get_meta ` and here we return to our loop problem

tensordict/tensordict.py
Zooll: Yes, looks like I lost it.

tensordict/tensordict.py
Zooll: Yes, I preferred set because `_TensorDictKeysView` is independent class for keys retrieving.  What if we create several _TensorDictKeysView from same TensorDictBase, generally they couldn't work simultaneously.

tensordict/tensordict.py
```diff
@@ -1703,7 +1729,12 @@ def permute(
         )
 
     def __repr__(self) -> str:
+        if self._being_called:
+            return "Auto-nested"
+        field_str = indent("Auto-nested", 4 * " ")
```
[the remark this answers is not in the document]
Zooll: Yes, looks like I've lost this line during modification.

tensordict/tensordict.py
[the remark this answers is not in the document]
tcbegley: I see. Well what if we added something like the following to `_get_meta`.

```python
if key not in self._dict_meta:
    raise KeyError
```

That should prevent us from falling back to `__missing__` as you say, but also means we don't have to iterate over the nested structure twice?

tensordict/tensordict.py
tcbegley: I see. I think it would be hard to use multiple key views simultaneously, but the set works just fine so let's leave it as it is 🙂

tensordict/tensordict.py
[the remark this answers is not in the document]
Zooll: Sorry, I'm not sure I understood you correctly. In current code it's right behaviour to create new object of MetaTensor if we do not have it in self._dict_meta.  If we add 
```Python
if key not in self._dict_meta:
    raise KeyError
```
we will broke some new instances creation. Right now we check it only during `_requires_grad` process and assume that if we do not have a `key` it doesn't necessary to check grad for this item.

Zooll:
> This is looking good, thanks for working on it!
> 
> I've left a few comments inline.
> 
> The other main thing that we should add are some tests. We can add a couple of tests for specific functionality such as checking for the presence of `"Auto-nested"` in the repr, but as @vmoens suggested in #106 it would be nice to run all tests on a tensordict with autonesting and check that we haven't inadvertently broken something.
> 
> Check [this class](https://github.com/pytorch-labs/tensordict/blob/4318760e6f4a6b1f3c2e3936379091cbbe4a3a1e/test/_utils_internal.py#L29) that gets used in the tests, you can add a method for `auto_nested_td` or similar, and then add it to the tests [here](https://github.com/pytorch-labs/tensordict/blob/4318760e6f4a6b1f3c2e3936379091cbbe4a3a1e/test/test_tensordict.py#L602-L618). In principle it should work for anything that `nested_td` works for, but in any case it will be interesting to see what breaks!

I added "Auto-nested" as you recommended, but because of nature. of auto-nesting I think we can't test it using existing test cases. For instance we have this test:
```python
 def test_items_values_keys(self, td_name, device):
        torch.manual_seed(1)
        keys = list(td.keys())
        values = list(td.values())
        items = list(td.items())
    
        # Test td.items()
        constructed_td1 = TensorDict({}, batch_size=td.shape)
        for key, value in items:
            constructed_td1.set(key, value)
    
>       assert (td == constructed_td1).all()
```
if we use this td:
```python
 def auto_nested_td(self, device):
        tensordict = TensorDict({
            "a": torch.randn(4, 3, 2, 1, 5),
            "b": torch.randn(4, 3, 2, 1, 10),
            }, 
            batch_size=[4, 3, 2, 1],
            device=device,)
        tensordict["self"] = tensordict
        return tensordict
```
when we construct `constructed_td1` properly, because out `self` property contains our "main" td. and when we copy it to  constructed_td1.self property it's not the same as we put constructed_td1['self'] = constructed_td1. More over when we try to compare them, out __eq__ function will try to do deep comparation. It mean we are not able to compare td with self nested key.
I don't know what to do in this case we can discuss it or omit it in this iteration because we need to have a strategy about working with auto(or self) nesting.

Zooll:
After https://github.com/pytorch-labs/tensordict/pull/121, It became easier to solve this auto-nesting bug. I updated pull request. But it still has a problem to add tests which I mentioned here https://github.com/pytorch-labs/tensordict/pull/119#issuecomment-1362954669

tensordict/tensordict.py
```diff
@@ -153,12 +154,13 @@ def __iter__(self):
 
     def _iter_helper(self, tensordict, prefix=None):
         items_iter = self._items(tensordict)
-
+        self.loop_set.add(id(tensordict))
```
vmoens, line 157: `_TensorDictKeysView` is already a compute bottleneck in many cases and I'm afraid that this set manipulation adds a consequent overhead to the whole thing.
We could make make a list instead and iterate over the reversed list:
```python
self.loop_list = []
...
self.loop_list.append(id(tensordict))
...
  and id(value) not in reversed(self.loop_set)
...
self.loop_list = self.loop_list[:-1]
```
(not sure I get the full picture or that would be significantly faster).

If you have ideas about how to make this faster or some tests that show that for deeply nested tensordicts we don't have a significant overhead with the set I'd be supportive of it.

tensordict/tensordict.py
```diff
@@ -1717,7 +1722,11 @@ def permute(
         )
 
     def __repr__(self) -> str:
+        if self._being_called:
```
vmoens, line 1725: `_being_called` is used in repr only? Then why not `_being_repr` or smth like it?

tensordict/tensordict.py
```diff
@@ -1791,6 +1800,10 @@ def __iter__(self) -> Generator:
     def flatten_keys(
         self, separator: str = ".", inplace: bool = False
     ) -> TensorDictBase:
+        if self._being_flattened:
```
vmoens, line 1803: One idea could be to decorate this with a `try / except / finally` to make sure that this is fault-tolerant:
```
try:
    ...
    self._being_flattened = True
    ...
finally:
    self._being_flattened = False
```

tensordict/tensordict.py
vmoens, line 1725: Same as `flatten_keys`, we can make this fault-tolerant with a `try / finally`

tensordict/tensordict.py
```diff
@@ -1830,10 +1844,14 @@ def flatten_keys(
                     inner_tensordict = self.get(key).flatten_keys(
                         separator=separator, inplace=inplace
                     )
-                    for inner_key, inner_item in inner_tensordict.items():
-                        tensordict_out.set(separator.join([key, inner_key]), inner_item)
+                    if inner_tensordict is not self.get(key):
```
vmoens, line 1847: I wonder if this is compatible with lazy tensordicts (eg `LazyStackedTensorDict`) where two calls to `get(key)` return items which ids are different (eg because they are the results of a call to `stack`)

vmoens (commented):
Regarding the tests, I'm happy with tests being independent of the others although I wonder if the fact that a call to `__eq__` is prohibited should not be addressed by this PR.
TBH I did not quite get what the issue was. Can you elaborate a bit more? 
You can also code it and push the code with the tests breaking, that would give me some visibility on what's happening.

tensordict/tensordict.py
Zooll: Oh, I've just used same var name as you proposed here https://github.com/pytorch-labs/tensordict/issues/106, but sure lets rename it properly.

tensordict/tensordict.py
Zooll: Got it. Make sense

tensordict/tensordict.py
Zooll: It's not just one way traversal. You should keep all states you went through. For ex: 
image
Each node is tensordict and arrows are drawn if one TD contains another TD. If we go through 1->2->3->4 and then return add remove from loop_list all this nodes we will go to node 5 and then again to node 4. But if we keep it in our set there will not be this problem.  Also because it's not one way traversal in terms of complexity it's cheaper to use set with (O(1) in amortized complexity).

tensordict/tensordict.py
Zooll: It's interesting point. I didn't research 'LazyStackedTensorDict' 's behaviour. But if 'get(key)' return new ids every time how can we detect self loop?

tensordict/tensordict.py
Zooll: > Regarding the tests, I'm happy with tests being independent of the others although I wonder if the fact that a call to `__eq__` is prohibited should not be addressed by this PR. TBH I did not quite get what the issue was. Can you elaborate a bit more? You can also code it and push the code with the tests breaking, that would give me some visibility on what's happening.

This issue was about infinite recursion during print, flatten_keys, and list(keys()). Issue # https://github.com/pytorch-labs/tensordict/issues/106. This is example of code how to verify this PR:
<details>
  <summary>Code</summary>

```python
tensordict = TensorDict({
     "key 1": torch.ones(3, 4, 5),
     "key 2": torch.zeros(3, 4, 5, dtype=torch.bool),
}, batch_size=[3, 4])

tensordict2 = TensorDict({
    "super key 1": torch.ones(3, 4, 6),
    "super key 2": torch.zeros(3, 4, 7, dtype=torch.bool),
}, batch_size=[3, 4])

tensordict["innerTD"] = tensordict2
tensordict2["self 2"] = tensordict2
tensordict["self"] = tensordict
tensordict["key 3"] = torch.zeros(3, 4, 7, dtype=torch.bool)

f_keys = tensordict.flatten_keys()  # Ok
print(f_keys)
td_list = list(tensordict.keys(include_nested=True))  # Ok
print(td_list)
print(tensordict)  # Ok
```
</details>

Zooll:
> Regarding the tests, I'm happy with tests being independent of the others although I wonder if the fact that a call to `__eq__` is prohibited should not be addressed by this PR. TBH I did not quite get what the issue was. Can you elaborate a bit more? You can also code it and push the code with the tests breaking, that would give me some visibility on what's happening.

This issue was about infinite recursion during print, flatten_keys, and list(keys()). Issue # https://github.com/pytorch-labs/tensordict/issues/106. This is example of code how to verify this PR:
<details>
  <summary>Code</summary>

```python
tensordict = TensorDict({
     "key 1": torch.ones(3, 4, 5),
     "key 2": torch.zeros(3, 4, 5, dtype=torch.bool),
}, batch_size=[3, 4])

tensordict2 = TensorDict({
    "super key 1": torch.ones(3, 4, 6),
    "super key 2": torch.zeros(3, 4, 7, dtype=torch.bool),
}, batch_size=[3, 4])

tensordict["innerTD"] = tensordict2
tensordict2["self 2"] = tensordict2
tensordict["self"] = tensordict
tensordict["key 3"] = torch.zeros(3, 4, 7, dtype=torch.bool)

f_keys = tensordict.flatten_keys()  # Ok
print(f_keys)
td_list = list(tensordict.keys(include_nested=True))  # Ok
print(td_list)
print(tensordict)  # Ok
```
</details>

vmoens:
Thanks
If you could write some tests it would be easier to comment by using the fit review process

tensordict/tensordict.py
vmoens: Makes sense! thanks

Zooll:
@vmoens I've added tests for keys() and __repr__(). Unfortunately, I didn't handle to add tests for `flatten_keys` in current `test_flatten_keys` for several reasons. 1) We can't just add 
```python
@pytest.mark.parametrize(
    "td_name",
    [...,
    auto_nested_td,
    ]
)
```
because there are many other tests which will be broken.
2) I tried to add `test_flatten_keys(self, td_name, device, inplace, separator)` with auto nested td only, but faced with infinite recursive unlocking of self nested tensordict which doesn't related to flatten_keys problem https://github.com/pytorch-labs/tensordict/blob/455b64adc16833567f45d917d40e4abf3b0a898c/test/test_tensordict.py#L1716

vmoens:
We can broke the tests, I'd rather have that and solve the bug. Otherwise it's hard to see what the problem is.

Zooll:
> We can break the tests, I'd rather have that and solve the bug. Otherwise it's hard to see what the problem is.

I've added auto_nested_td into tests

vmoens:
Some benchmarking first:
On this branch, common_ops:
```
TensorDict({}, [3, 4]) 0.054020772999999966
TensorDict({'a': a, 'b': b}, [3, 4]) 0.10693320900000014
TensorDict({'a': a, ('b', 'b1'): b}, [3, 4]) 0.26566142599999987
TensorDict({'a': a, 'b': {'b1': b}}, [3, 4]) 0.19251463799999957
tdc = td.clone() 0.23632501300000008
tdc = td.clone();tdc['c'] = c 0.2863006590000001
tdc = td.clone();tdc.set('c', c) 0.2764215059999984
tdc = td.clone();tdc.share_memory_() 1.781234584
tdc = td.clone();tdc.update(td2) 0.41446814300000057
tdc = td.clone();tdc.update(td2) 0.6884915679999999
tdc = td.clone();tdc['b', 'b1'] = b 0.3841283730000029
tdc = td.clone();tdc['c', 'c', 'c'] = c 0.6221515459999978
```
On main
```
TensorDict({}, [3, 4]) 0.06562997599999987
TensorDict({'a': a, 'b': b}, [3, 4]) 0.1289677100000004
TensorDict({'a': a, ('b', 'b1'): b}, [3, 4]) 0.29692919799999995
TensorDict({'a': a, 'b': {'b1': b}}, [3, 4]) 0.24163629199999992
tdc = td.clone() 0.26948069800000063
tdc = td.clone();tdc['c'] = c 0.3178999400000002
tdc = td.clone();tdc.set('c', c) 0.3091379649999997
tdc = td.clone();tdc.share_memory_() 1.806307208999998
tdc = td.clone();tdc.update(td2) 0.42458679399999966
tdc = td.clone();tdc.update(td2) 0.6898641609999991
tdc = td.clone();tdc['b', 'b1'] = b 0.4234232099999993
tdc = td.clone();tdc['c', 'c', 'c'] = c 0.6865369579999978
```

So we're good on that side.

Functional:
```
instantiation, functorch: 2.3080958469999997
instantiation, tensordict: 1.3075678029999995
exec, functorch: 4.814458004
exec, tensordict: 5.19893972
```

Other benchmarks seem to be roughly similar.

vmoens:
@Zooll I'm on it. Thanks for your hard work!

I tried to solve the `to_tensordict()` but it's a hard one. I need to sleep on this because all the solutions I can think of are either completely wrong or way to complex.

I'll keep you posted!

vmoens:
Closing as this has been addressed in #201

PR #122, opened 2022-12-19
[Feature] Tensorclass device

tcbegley:
## Description

This PR allows the user to specify `device` when instantiating a TensorClass (`@tensorclass` decorated dataclass).

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -102,7 +102,9 @@ def tensorclass(cls: T) -> T:
         fields=[("batch_size", torch.Size, field(default_factory=list))],
     )
 
-    EXPECTED_KEYS = set(datacls.__dataclass_fields__.keys())
+    EXPECTED_KEYS = {
+        k for k in datacls.__dataclass_fields__.keys() if k != "batch_size"
+    }
 
     class _TensorClass(datacls, metaclass=_TensorClassMeta):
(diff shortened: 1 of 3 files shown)


tensordict/prototype/tensorclass.py
```diff
@@ -102,7 +102,9 @@ def tensorclass(cls: T) -> T:
         fields=[("batch_size", torch.Size, field(default_factory=list))],
     )
 
-    EXPECTED_KEYS = set(datacls.__dataclass_fields__.keys())
+    EXPECTED_KEYS = {
+        k for k in datacls.__dataclass_fields__.keys() if k != "batch_size"
+    }
```
tcbegley, line 107: If we don't exclude `batch_size` here then assinging to `.batch_size` attribute attempts to place the supplied value in the underlying tensordict under the key `"batch_size"`.

tcbegley (commented):
Because of the logic in `__setattr__`, it's not currently possible to set either `device` or `batch_size` by assigning to those attributes. I understand why we want to prevent setting of arbitrary attributes, but do we want to make an exemption for these two?

vmoens (approved):
LGTM

vmoens:
> Because of the logic in `__setattr__`, it's not currently possible to set either `device` or `batch_size` by assigning to those attributes. I understand why we want to prevent setting of arbitrary attributes, but do we want to make an exemption for these two?

Yes! But for device you'd need to call .to no?

PR #124, opened 2022-12-19
[Feature] Tensorclass updates

tcbegley:
## Description

This PR makes a few updates to tensorclasses. An earlier version of these changes introduced a new base class which I have now removed again.

The main things to note
1. I've added properties corresponding to `batch_size` and `device` which allow them to be set as in the case of `TensorDict`.  Note the necessary change in `__setattr__` which unlike `__getattr__` is called whether or not the attribute exists. Also note that `batch_size` is no longer one of the dataclass fields. I think this makes sense as the fields should correspond to keys in the tensordict and hence not include `batch_size`.
2. I've added a new `is_tensorclass` helper function to enable type checking of tensorclasses and tensorclass instances. It is based on `dataclasses.is_dataclass` in the Python standard library (and in fact uses it in the implementation).
3. I've defined `__repr__` on the metaclass to make sure that information like where a tensorclass was defined is preserved in its representation, rather than being changed to `tensordict.prototype.tensorclass.tensorclass.<locals>.MyClass` or similar. 
4. I have `to_tensordict` return the result of `super().to_tensordict()` without wrapping it in a tensorclass (since in this case, we explicitly don't want that behaviour).

### DIFF
--- a/test/test_tensorclass_nofuture.py
+++ b/test/test_tensorclass_nofuture.py
@@ -10,6 +10,7 @@ from _utils_internal import get_available_devices
 
 from tensordict import LazyStackedTensorDict, TensorDict
 from tensordict.prototype import tensorclass
+from tensordict.prototype.tensorclass import _TensorClassBase
 from tensordict.tensordict import _PermutedTensorDict, _ViewedTensorDict, TensorDictBase
 from torch import Tensor
 
@@ -45,6 +46,7 @@ def test_type():
         batch_size=[3, 4],
     )
     assert isinstance(data, MyData)
+    assert isinstance(data, _TensorClassBase)
 
 
 @pytest.mark.parametrize("device", get_available_devices())
@@ -58,6 +60,11 @@ def test_device(device):
     assert data.device == device
     assert data.X.device == device
     assert data.y.device == device
+
+    with pytest.raises(
+        RuntimeError, match="device cannot be set using tensorclass.device = device"
+    ):
+        data.device = torch.device("cpu")
 
 
 def test_banned_types():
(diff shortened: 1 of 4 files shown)


tcbegley:
One thing I've noticed is that it's not really clear how to write type hints for functions that accept a tensorclass. Best I can think of atm is for us to define a custom `Protocol` for tensorclasses, but it's not perfect and would require us to add `typing-extensions` as a dependency to maintain Python 3.7 support.

vmoens:
> One thing I've noticed is that it's not really clear how to write type hints for functions that accept a tensorclass. Best I can think of atm is for us to define a custom `Protocol` for tensorclasses, but it's not perfect and would require us to add `typing-extensions` as a dependency to maintain Python 3.7 support.

Do you mean for us, maintainers, or for any developer using tensorclass?
Developers will use type hints with their classes. For us it's true that there is no clear way of doing so.
I would guess that the builtin dataclass utils also struggle with this?

test/test_tensorclass_nofuture.py
```diff
@@ -313,6 +364,19 @@ def test_permute():
...
+    @tensorclass
+    class MyDataNested:
+        X: torch.Tensor
+        y: "MyDataNested" = None
+
+    X = torch.ones(3, 4, 5)
+    batch_size = [3, 4]
+    data_nest = MyDataNested(X=X, batch_size=batch_size)
+    data = MyDataNested(X=X, y=data_nest, batch_size=batch_size)
+    assert isinstance(data.y, MyDataNested), type(data.y)
```
vmoens, line 377: oh that works! Brilliant!

vmoens (approved):
LGTM Thanks!!

PR #125, opened 2022-12-20
[BugFix] Fix items_meta, values_meta

tcbegley:
## Description

I noticed that `items_meta` and `values_meta` were missing the `leaves_only` keyword that their analogues all expose. While adding it, I realised that the implementation of each of these methods was in fact broken in the `make_unset=False` case, since we were doing checks like `if isinstance(item, TensorDictBase)` where `item` is expected to be a `MetaTensor`. 

This PR both adds the missing kwargs and fixes the bug, and adds a test to check for the correct behaviour.


vmoens (approved):
Wow that's a big fix! Thanks so much

PR #133, opened 2022-12-27
[BugFix] add skipif not _has_functorch

apbard:
Some test use functorch's vmap function. Currently, this is defined only if functorch is available. Thus, we need to skip all related tests if functorch is not available


vmoens (approved):
LGTM thanks!

PR #139, opened 2023-01-03
[Feature] Support `range` in indexing operations

tcbegley:
## Description

This PR adds support for `range` in indexing operations, both get and set.

E.g.

```python

>>> td = TensorDict({"a": [1, 2, 3, 4, 5]}, [5])
>>> print(td[range(2)])
TensorDict(
    fields={
        a: Tensor(torch.Size([2]), dtype=torch.int64)},
    batch_size=torch.Size([2]),
    device=None,
    is_shared=False)
>>> td[range(2)] = TensorDict({"a": [0, 0]}, [2])
>>> print(td["a"])
tensor([0, 0, 3, 4, 5])
```

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1952,12 +1952,16 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
             idx = torch.tensor(idx, device=self.device)
             return self._index_tensordict(idx)
 
+        if isinstance(idx, range):
+            idx = torch.tensor(idx, device=self.device)
+            return self._index_tensordict(idx)
+
         if isinstance(idx, tuple) and any(
-            isinstance(sub_index, list) for sub_index in idx
+            isinstance(sub_index, (list, range)) for sub_index in idx
         ):
             idx = tuple(
                 torch.tensor(sub_index, device=self.device)
-                if isinstance(sub_index, list)
+                if isinstance(sub_index, (list, range))
                 else sub_index
                 for sub_index in idx
             )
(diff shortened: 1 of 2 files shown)


tensordict/tensordict.py
```diff
@@ -1952,12 +1952,16 @@ def __getitem__(self, idx: INDEX_TYPING) -> TensorDictBase:
             idx = torch.tensor(idx, device=self.device)
             return self._index_tensordict(idx)
 
+        if isinstance(idx, range):
+            idx = torch.tensor(idx, device=self.device)
+            return self._index_tensordict(idx)
```
tcbegley, line 1957: There's some repetition here, but I figured it's deliberate to benefit from early return in the most common case of indexing with a list. I also assumed that `range` would be less common than `np.ndarray`.

As a side-note: this seems to be the only place we handle `np.ndarray`, it's not considered in other `__setitem__` or `__getitem__` definitions.

test/test_tensordict.py
```diff
@@ -1383,6 +1385,13 @@ def test_getitem_string(self, td_name, device):
         td = getattr(self, td_name)(device)
         assert isinstance(td["a"], (MemmapTensor, torch.Tensor))
 
+    def test_getitem_range(self, td_name, device):
+        torch.manual_seed(1)
+        td = getattr(self, td_name)(device)
+        assert_allclose_td(td[range(2)], td[[0, 1]])
+        assert_allclose_td(td[range(1), range(1)], td[[0], [0]])
+        assert_allclose_td(td[:, range(2)], td[:, [0, 1]])
```
vmoens, line 1393: let's add Ellipsis here

vmoens (approved):
LGTM
We can try with ellipsis too if that's ok

test/test_tensordict.py
tcbegley: Added a test for indexing with `Ellipsis`.

vmoens (approved):

PR #141, opened 2023-01-03
[BugFix] Stack dimension indexing

tcbegley:
## Description

This PR fixes the failing tests on main arising from the merging of #139 and #140.

#140 introduced a new test class `nested_stacked_td` with stack dimension `1` to `TestTensorDictsBase` which surfaced issues with stack dimension indexing in `LazyStackedTensorDict`. In brief, whatever was used to index the stack dimension was simply used to index the underlying list at `self.tensordicts`. Thus while numbers and slices worked fine, you would get an error if you tried to index the stack dimension with a list, tensor, or `range` object.

With these changes, things should now work as expected.

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -4475,7 +4475,16 @@ class LazyStackedTensorDict(TensorDictBase):
             )
 
             if self.stack_dim < len(item):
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks!

PR #143, opened 2023-01-03
[BugFix] Prevent calls to get_nestedtensor when stack_dim is not 0

tcbegley:
## Description

This PR prevents the user from calling `LazyStackedTensorDict.get_nestedtensor` unless `stack_dim` is `0`. 

I did wonder whether it should be allowed in the case where each item under that key had the same shape (in which case it would be possible to return the result of a call to `torch.stack`), but I figured that it would be surprising to get a regular tensor rather than a nested tensor from a call to `get_nestedtensor`.


vmoens (approved):
LGTM thanks!
cc @matteobettini

PR #146, opened 2023-01-05
[Draft] Coding trees with TensorDict

tcbegley:
## Description

This PR contains a draft implementation of support for trees using `TensorDict`.

We allow the user to create a tree structure where each node in the tree is a tensordict with the same keys. The entire contents of the tree are backed by a single source tensordict with pre-allocated memory. As nodes are added to the tree the data in the node is stored in the source. The primary advantage of this setup is that collecting the data from multiple nodes requires us to simply index the source which can be done very efficiently compared to a naive alternative which would require us to iterate over the nodes we want and stack the results.

The interface is very preliminary and up for debate. The tests should give an idea of usage, here's a very basic example.

```python
>>> import torch
>>> from tensordict.prototype import make_tree

>>> # create the root node from a tensordict
>>> # preallocate memory for 100 nodes in the source
>>> root = make_tree(TensorDict({"data": torch.rand(2, 3, 4)}, [2, 3]), n_nodes=100)

>>> # adding a new tensordict as a child creates a new node in the tree
>>> root["left"] = TensorDict({"data": torch.rand(2, 3, 4)}, [2, 3])

>>> # we can now gather the data for the two nodes in the tree
>>> root.get_multiple_items("data", ("left", "data"))
tensor(...)  # shape torch.Size([2, 2, 3, 4])

>>> # alternatively we can create a tree from a nested tensordict
>>> root = make_tree(
...     TensorDict(
...         {"data": torch.ones(2, 3, 4), "left": TensorDict({"data": torch.zeros(2, 3, 4)}, [2, 3])},
...         [2, 3]
...     ),
...     n_nodes=100,
... )
>>> root.get_multiple_items("data", ("left", "data"))
```

### DIFF
--- a/tensordict/prototype/tree.py
+++ b/tensordict/prototype/tree.py
@@ -10,34 +10,6 @@ from tensordict.tensordict import (
 )
 from tensordict.utils import _is_shared
 
-_has_functorch = False
-try:
-    try:
-        from functorch._C import is_batchedtensor
-    except ImportError:
-        from torch._C._functorch import is_batchedtensor
-
-    _has_functorch = True
-except ImportError:
-    _has_functorch = False
-
-    def is_batchedtensor(tensor):
-        """Placeholder for the functorch function."""
-        return False
-
-
-try:
-    from torchrec import KeyedJaggedTensor
-
-    _has_torchrec = True
-except ImportError as err:
-    _has_torchrec = False
-
-    class KeyedJaggedTensor:
-        pass
-
-    TORCHREC_ERR = str(err)
-
 KEY_ERR = (
     "All nodes must have the same leaf keys as the root node. The leaf keys on the "
     "root node are {leaf_keys}, but you are trying to set the key(s) {key}."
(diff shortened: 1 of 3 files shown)


tensordict/prototype/tree.py
```diff
@@ -0,0 +1,221 @@
...
+            key, subkey = key[0], key[1:]
+        else:
+            subkey = ()
+        if key in self._children:
+            if subkey:
+                return self._children[key].get(subkey)
+            return self._children[key]
+        if subkey:
+            raise KeyError("key not valid")
+        return super().get(key)
```
tcbegley, line 126: This `get` method was a quick fix for a broken `__repr__`. I think these could be consolidated since they are very similar. In fact it's possible we could just delete `__getitem__` since I think `SubTensorDict` should delegate strings and tuples of strings to `get` anyway.

tcbegley:
Added some simple benchmarks. lmk what other things you think we should benchmark

```
get_multiple_items
tensordict: 2.145204999993439
tree: 0.9129869579919614

get_multiple_items_deep
tensordict 2.4592998750013066
tree 0.8956774580001365
```

vmoens:
> 

impressive!

PR #150, opened 2023-01-06
[BugFix] Add __len__ method to tensorclass

tcbegley:
## Description

Previously `len(my_tensorclass)` would result in an error `TypeError: object of type 'MyTensorClass' has no len()`.

This PR adds a `__len__` method to generated tensorclasses that falls back to the underlying tensordicts `__len__` method.


vmoens (approved):
Thanks for spotting and fixing this

PR #157, opened 2023-01-09
[Feature] Added support to save non-tensor data in tensor class

sreevasthav:
## Description

This is a WIP PR for Issue 145 : [Feature Request] tensorclass with non-tensor data

Made the changes to accommodate the following:
1. Save non-tensor data
2. Access tensor class object
3. Indexing

## Types of changes

-  New feature (non-breaking change which adds core functionality)

## Checklist


vmoens:
@sreevasthav we have made some important changes to the tensorclass construction. If you need help resolving the conflicts don't hesitate to reach out!

sreevasthav:
@vmoens  Thanks for letting me know. Is this a functionality change or just code refactoring? Can you share me the issue for this?

tcbegley:
@sreevasthav PR is #165. There's no corresponding issue, but the main problem that we needed to fix was making tensorclasses picklable. That meant dropping the locally defined `_TensorClass`. 

No change in functionality, but lots of stuff got moved out of `tensorclass` decorator into globally defined functions etc. Those functions should broadly correspond to methods on the old `_TensorClass`.

sreevasthav:
@vmoens  Can you please review the code? The above code has one limitation with nested use case. I am trying to save the non-tesor data for a nested class using class type since this is the only information available while getting the attributes as we already saved the tensor class to tensor dict. 

The issue with the implementation is when the non-tensor data is optional and one of the nested class didn't use the attribute. Since we are saving at the class type level while getting the attributes we don't have this information we fetch the non-tensor data for all the nested class of same type as the same value. 

Let me know if you can suggest any other approach to save and extract this correctly

tensordict/prototype/tensorclass.py
```diff
@@ -166,7 +173,11 @@ def wrapper(self, *args, batch_size=None, device=None, _tensordict=None, **kwarg
                     f"Keys from the tensordict ({set(_tensordict.keys())}) must "
                     f"correspond to the class attributes ({expected_keys})."
                 )
-            input_dict = {key: None for key in _tensordict.keys()}
+            # Handling Non-Tensor data
+            if _non_tensordict:
```
vmoens, line 177: I prefer the more explicit `_non_tensordict is not None` otherwise there's a risk that someone inputs an empty dict or something like it and hopes that it will return True when it won't.

test/test_tensorclass.py
```diff
@@ -496,6 +587,9 @@ class MyDataParent:
     assert data.tensordict["X"] is X_clone
     data.z = td_clone
     assert data.tensordict["z"] is td_clone
+    v_new = 'test_bluff'
+    data.v = v_new
```
vmoens, line 591: What if `MyDataNest` has a non-tensor attribute and we do
```python
data.y.z = "x"
```
would that work?

vmoens (commented):
LGTM, just a couple of things:
for nested tensorclass, can you clarify the behaviour a bit?
Can we add specific tests for that, like
```python
    @tensorclass
    class MyDataNest:
        X: torch.Tensor
        v: str
    @tensorclass
    class MyDataParent:
        W: Any
        X: Tensor
        z: TensorDictBase
        y: MyDataNest
        v: str

    batch_size = [3, 4]
    if any_to_td:
        W = TensorDict({}, batch_size)
    else:
        W = torch.zeros(*batch_size, 1)
    X = torch.ones(3, 4, 5)
    X_clone = X.clone()
    td = TensorDict({}, batch_size)
    td_clone = td.clone()
    data_nest = MyDataNest(X=X, v="x", batch_size=batch_size)
    v = 'test_tensorclass'
    data = MyDataParent(X=X, y=data_nest, z=td, W=W, v=v, batch_size=batch_size)
    
    assert data.y.v == "x"
    data.y.v = "y"
    assert data.y.v == "y"
    data_nest.v = "x"
    assert data.y.v == "x"
```

(if that makes sense and it's feasible)

vmoens:
Also I think the lint is not working, check the CONTRIBUTING.md for more info

tensordict/prototype/tensorclass.py
sreevasthav: In the function _get_non_tensor_dict in the line 549, I had to initialize an empty dict and add key value pairs if the value is a non-tensor value and hence used without None

sreevasthav:
@vmoens  Thanks for pointing out on setting the attributes in nested use case, I have updated the code and updated the test cases to include the suggestions you pointed out for

tensordict/prototype/tensorclass.py
vmoens: so what are we testing precisely? That it is empty? 
In fact I know it's a very pythonish way of doing things (and very common) but IMO it doesn't make the code very readable. `len(dictionary) == 0` or something similar is more verbose but helps understand what we're testing.

test/test_tensorclass.py
```diff
@@ -527,6 +622,14 @@ class MyDataParent:
     assert data.tensordict["X"] is X_clone
     data.z = td_clone
     assert data.tensordict["z"] is td_clone
+    v_new = "test_bluff"
+    data.v = v_new
+    assert data.v == v_new
+    assert data.y.v == "test_nested"
+    data.y.v = "test_nested_new"
+    assert data.y.v == "test_nested_new"
```
vmoens, line 630: is `data_nest.v` changed too here?

vmoens (approved):
Superb
I left a couple of comments.
@apbard @tcbegley the constructor is modified by this PR, let me know if that's a problem for your PRs

tcbegley:
> the constructor is modified by this PR, let me know if that's a problem for your PRs

It does clash with what we were doing, but it's not clear immediately whether it's a problem. Will have a play around and understand better what you're doing here!

vmoens:
@sreevasthav could you update the doc too? 
https://github.com/pytorch-labs/tensordict/blob/main/docs/source/reference/prototype.rst

test/test_tensorclass.py
sreevasthav: No. data_nest.v would remain the same(" test_nested"). Do we want to change the behavior here?

tensordict/prototype/tensorclass.py
sreevasthav: Sure. Makes sense. Will update the code to add the explicit condition.

tensordict/prototype/tensorclass.py
```diff
@@ -27,6 +28,8 @@
 
 # For __future__.annotations, we keep a dict of str -> class to call the class based on the string
 CLASSES_DICT = {}
+# For nested usecase, we keep dict to save key,value pairs of non-tensor data
+TENSOR_CLASS_NON_TENSOR_DATA = collections.defaultdict(dict)
```
sreevasthav, line 32: @vmoens I had a chat with Alessandro and Tom offline. Alessandro suggested to consider "self" member instead of a global dict. I have tried this today. Looks like the test case behavior you mentioned here may not work. https://github.com/pytorch-labs/tensordict/pull/157#pullrequestreview-1253808373. What are your thoughts? 
CC: @tcbegley @apbard

vmoens: I agree that the global variable is not ideal, thanks @apbard for the suggestion.
What about this:
In a tensorclass, if a store field is a tensorclass, we keep the nested tensordict structure as it is now, but we also store a reference to the stored tensorclass. Very roughly put:
```python
@tesorclass 
class MyDataNested:
    a_tensor: torch.Tensor
    a_string: str

@tesorclass 
class MyDataGlob:
    inner_object: MyDataNested

data_inner = MyDataNested(torch.randn(3), "hello world")
data_outer = MyDataGlob(data_inner)

# from the outside, everything seems fine
assert data_outer.data_inner is data_inner

# we can look at the private attributes now
assert data_outer.tensordict["data_inner"] is data_inner.tensordict
assert data_outer._inner_classes["data_inner"] is data_inner

# this will: (1) look at data_inner, and realize that it is a tensoclass (not a tensor or tensordict), (2) access it from `_inner_classes` instead (3) assign the string to the inner class
data_outer.data_inner.a_string = "to infinity and beyond"
```

The need to keep 2 copies (one of the tensorclass, one of the tensordict) is to keep all the functionality from tensordict. This also means that all methods such as `__getitem__` etc should keep the `_inner_classes` dictionary updated.

@tcbegley @apbard wdyt? Am I overthinking it? Is there a simpler solution (that does not involve rewriting the entire tensordict for dataclasses)?

apbard: @vmoens my suggestion was quite similar with a slight variation in terms of semantics.
Each tensorclass internally will have 2 private members for storage:
self.tensordata (aka a tensordict) and self.otherdata (aka a standard dict). The former will store the tensor data as current implementation, the other all other objects including nested class  
then the get and set methods will redirect stuff accordingly.

p.s we could potentially introduce a third dict to map keys to type for faster indexing and to faster sanity checks (avoiding double key)...but needs benchmarking

sreevasthav:
> LGTM, just a couple of things: for nested tensorclass, can you clarify the behaviour a bit? Can we add specific tests for that, like
> 
> ```python
>     @tensorclass
>     class MyDataNest:
>         X: torch.Tensor
>         v: str
>     @tensorclass
>     class MyDataParent:
>         W: Any
>         X: Tensor
>         z: TensorDictBase
>         y: MyDataNest
>         v: str
> 
>     batch_size = [3, 4]
>     if any_to_td:
>         W = TensorDict({}, batch_size)
>     else:
>         W = torch.zeros(*batch_size, 1)
>     X = torch.ones(3, 4, 5)
>     X_clone = X.clone()
>     td = TensorDict({}, batch_size)
>     td_clone = td.clone()
>     data_nest = MyDataNest(X=X, v="x", batch_size=batch_size)
>     v = 'test_tensorclass'
>     data = MyDataParent(X=X, y=data_nest, z=td, W=W, v=v, batch_size=batch_size)
>     
>     assert data.y.v == "x"
>     data.y.v = "y"
>     assert data.y.v == "y"
>     data_nest.v = "x"
>     assert data.y.v == "x"
> ```
> 
> (if that makes sense and it's feasible)

@vmoens  I think the set attribute updates are specific to the object and hence if you change data_nest.v = "x" , data.y.v will still be "y"

sreevasthav:
@vmoens / @apbard / @vmoens Can you please review the PR again. Changed the most of the test cases to be nested ones so that we can validate nested behavior of all the defined functions

docs/source/reference/prototype.rst
```diff
@@ -7,7 +7,9 @@ The :obj:`@tensorclass` decorator helps you build custom classes that inherit th
 behaviour from :obj:`TensorDict` while being able to restrict the possible entries
 to a predefined set or implement custom methods for your class.
 Like :obj:`TensorDict`, :obj:`@tensorclass` supports nesting, indexing, reshaping,
-item assignment and many more features.
+item assignment and many more features. :obj:`@tensorclass` allows non-tensor entries,
+however all the tensor operations are strictly restricted to tensor attributes. One
```
apbard, line 11: I would expand this a bit. 
- how are supported
- an example of operation
- how nested tensordict and nested tensorclass work

tensordict/prototype/tensorclass.py
```diff
@@ -130,7 +134,7 @@ def __torch_function__(
             )
 
     cls.__init__ = _init_wrapper(cls.__init__)
-    cls.from_tensordict = classmethod(_from_tensordict_wrapper(expected_keys))
+    cls.to_tensorclass = classmethod(_to_tensorclass_wrapper(expected_keys))
```
apbard, line 137: I would probably stick with the original naming. @vmoens wdyt?

tensordict/prototype/tensorclass.py
```diff
@@ -209,18 +215,26 @@ def wrapper(self, *args, batch_size, device=None, **kwargs):
...
+                raise ValueError(
+                    f"Keys from the non-tensor data ({set(non_tensordict.keys())}) must "
+                    f"correspond to the class attributes ({expected_keys})."
+                )
         # bypass initialisation. this means we don't incur any overhead creating an
         # empty tensordict and writing values to it. we can skip this because we already
         # have a tensordict to use as the underlying tensordict
         tc = cls.__new__(cls)
         tc.__dict__["tensordict"] = tensordict
+        if non_tensordict is not None:
```
apbard, line 236: is this if needed? shouldn't be initialised in all cases. That is 
`tc.__dict__["non_tensordict"] = non_tensordict if non_tensordict else {}`

tensordict/prototype/tensorclass.py
```diff
@@ -209,18 +215,26 @@ def wrapper(self, *args, batch_size, device=None, **kwargs):
...
             raise ValueError(
                 f"Keys from the tensordict ({set(tensordict.keys())}) must "
                 f"correspond to the class attributes ({expected_keys})."
             )
+        if non_tensordict is not None:
+            if not all(key in expected_keys for key in non_tensordict.keys()):
+                raise ValueError(
+                    f"Keys from the non-tensor data ({set(non_tensordict.keys())}) must "
+                    f"correspond to the class attributes ({expected_keys})."
+                )
```
apbard, line 230: I think we miss the check that ensures a key is present either in tensordict or in non_tensordict.
plus we need to check the type.

tensordict/prototype/tensorclass.py
```diff
@@ -302,16 +343,32 @@ def _setitem(self, item, value):
         raise ValueError("Invalid indexing arguments.")
     if not isinstance(value, self.__class__):
         raise ValueError("__setitem__ is only allowed for same-class assignement")
+    for key, val in self.non_tensordict.items():
```
apbard, line 346: what are you trying to do here?

tensordict/prototype/tensorclass.py
```diff
@@ -356,24 +413,52 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
 def __eq__(self, other):
     if not isinstance(other, self.__class__):
         return False
-    return self.tensordict == other.tensordict
+    keys1 = set(self.non_tensordict.keys())
+    keys2 = set(other.non_tensordict.keys())
+    if len(keys1.difference(keys2)) or len(keys1) != len(keys2):
+        raise KeyError(f"keys in {self} and {other} mismatch, got {keys1} and {keys2}")
```
apbard, line 419: @vmoens do we want to raise in this case?

tensordict/prototype/tensorclass.py
```diff
@@ -356,24 +413,52 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+    non_tensor = {}
+    for key, value in self.non_tensordict.items():
+        non_tensor[key] = value == other.non_tensordict.get(key)
```
apbard, line 422: why using the `.get(key)` method? the `[key]` call should be safe as key presence has already been checked

tensordict/prototype/tensorclass.py
```diff
@@ -356,24 +413,52 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
     if not isinstance(other, self.__class__):
         return True
-    return self.tensordict != other.tensordict
+    keys1 = set(self.non_tensordict.keys())
+    keys2 = set(other.non_tensordict.keys())
+    if len(keys1.difference(keys2)) or len(keys1) != len(keys2):
+        raise KeyError(f"keys in {self} and {other} mismatch, got {keys1} and {keys2}")
+    non_tensor = {}
+    for key, value in self.non_tensordict.items():
+        non_tensor[key] = value != other.non_tensordict.get(key)
```
apbard, line 437: same as `__eq__`

test/test_tensorclass_nofuture.py
```diff
@@ -207,23 +221,34 @@ def test_batch_size():
...
-        X=torch.ones(3, 4, 5),
-        y=torch.zeros(3, 4, 5, dtype=torch.bool),
-        batch_size=[3, 4],
-    )
+    x = torch.ones(3, 4, 5)
+    y = torch.zeros(3, 4, 5, dtype=torch.bool)
+    z = ["a", "b", "c"]
+    data = MyData2(X=x, y=y, z=z, batch_size=[3, 4])
 
     assert data[:2].batch_size == torch.Size([2, 4])
```
apbard, line 247: we should also check the content

test/test_tensorclass_nofuture.py
```diff
@@ -207,23 +221,34 @@ def test_batch_size():
...
     assert data[1].batch_size == torch.Size([4])
     assert data[1][1].batch_size == torch.Size([])
+    assert data[1].z == data[2].z == data[:2].z == z
+    assert data[:2].X.shape == torch.Size([2, 4, 5])
```
apbard, line 252: we miss getitem/indexing test with a nested class.
@vmoens, it is not clear to me what should be the behaviour in this case

test/test_tensorclass_nofuture.py
```diff
@@ -533,20 +729,29 @@ class MyDataParent:
...
+    data_nest = MyDataNest(X=X, v="test_nested", batch_size=batch_size)
+    v = "test_tensorclass"
+    data = MyDataParent(X=X, y=data_nest, z=td, W=W, v=v, batch_size=batch_size)
     data_nest_clone = data_nest.clone()
     assert type(data_nest_clone) is type(data_nest)
     data.y = data_nest_clone
-    assert data.tensordict["y"] is not data_nest.tensordict
-    assert data.tensordict["y"] is data_nest_clone.tensordict, (
-        type(data.tensordict["y"]),
+    assert data.y.tensordict is not data_nest.tensordict
```
apbard, line 738: do not understand this change. shouldn't it be simply "data.y is data_nest_clone"?

test/test_tensorclass_nofuture.py
```diff
@@ -533,20 +729,29 @@ class MyDataParent:
...
+    assert data.y.tensordict is data_nest_clone.tensordict, (
+        type(data.y.tensordict),
         type(data_nest.tensordict),
     )
     data.X = X_clone
     assert data.tensordict["X"] is X_clone
```
apbard, line 744: why are we checking this through the tensordict member?

BTW why tensordict is not protected? (i.e. with a `_` prefix)?

test/test_tensorclass_nofuture.py
```diff
@@ -533,20 +729,29 @@ class MyDataParent:
     X_clone = X.clone()
     td = TensorDict({}, batch_size)
     td_clone = td.clone()
-    data_nest = MyDataNest(X=X, batch_size=batch_size)
-    data = MyDataParent(X=X, y=data_nest, z=td, W=W, batch_size=batch_size)
+    data_nest = MyDataNest(X=X, v="test_nested", batch_size=batch_size)
```
apbard, line 732: please move the nested-related stuff at the bottom (before checks on batch size)  to not split "X" test

apbard: please also add comments like:
`# check set attr for tensor`
`# check set attr nested class`
etc...

test/test_tensorclass_nofuture.py
```diff
@@ -533,20 +729,29 @@ class MyDataParent:
...
     data.z = td_clone
     assert data.tensordict["z"] is td_clone
+    v_new = "test_bluff"
+    data.v = v_new
+    assert data.v == v_new
+    assert data.y.v == "test_nested"
+    data.y.v = "test_nested_new"
+    assert data.y.v == "test_nested_new"
+    data_nest.v = "test_nested"
+    assert data_nest.v == "test_nested"
```
apbard, line 754: we miss test setting a tensor of a nested class

test/test_tensorclass_nofuture.py
```diff
@@ -718,6 +943,7 @@ class MyClass:
     assert (tc_dest == tc).all()
     assert tc_dest.y.batch_size == tc.y.batch_size
     assert isinstance(tc_dest.y.x, MemmapTensor)
+    assert tc_dest.z == z
```
apbard, line 948: I don't see tests for get_attr

tensordict/prototype/tensorclass.py
sreevasthav: We check if all the non_tensor values are same before setting the item

tensordict/prototype/tensorclass.py
```diff
@@ -209,18 +215,26 @@ def wrapper(self, *args, batch_size, device=None, **kwargs):
     return wrapper
 
 
-def _from_tensordict_wrapper(expected_keys):
-    def wrapper(cls, tensordict):
+def _to_tensorclass_wrapper(expected_keys):
+    def wrapper(cls, tensordict, non_tensordict=None):
```
apbard, line 219: missing docstring

tensordict/prototype/tensorclass.py
```diff
@@ -231,40 +245,54 @@ def wrapper(cls, tensordict):
...
+    return {"tensordict": self.tensordict, "non_tensordict": self.non_tensordict}
 
 
 def _setstate(self, state):
     self.tensordict = state.get("tensordict", None)
+    self.non_tensordict = state.get("non_tensordict", None)
 
 
 def _getattribute_wrapper(getattribute):
     def wrapper(self, item):
```
apbard, line 257: missing functool.wrap ?

tensordict/prototype/tensorclass.py
```diff
@@ -231,40 +245,54 @@ def wrapper(cls, tensordict):
...
+            ):
+                out = self.non_tensordict[item]
+                return out
         return getattribute(self, item)
 
     return wrapper
 
 
 def _setattr_wrapper(setattr_, expected_keys):
     def wrapper(self, key, value):
```
apbard, line 279: missing functool.wrap ?

tensordict/prototype/tensorclass.py
```diff
@@ -277,8 +305,20 @@ def _getattr(self, attr):
 
     def wrapped_func(*args, **kwargs):
```
apbard, line 306: missing functool.wrap ?

tensordict/prototype/tensorclass.py
```diff
@@ -291,8 +331,9 @@ def _getitem(self, item):
...
     ):
         raise ValueError("Invalid indexing arguments.")
-    res = self.tensordict[item]
-    return self.from_tensordict(res)  # device=res.device)
+    tensor_res = self.tensordict[item]
+    non_tensor_res = self.non_tensordict
+    return self.to_tensorclass(tensor_res, non_tensor_res)  # device=res.device)
 
 
 def _setitem(self, item, value):
```
apbard, line 339: missing docstring

tensordict/prototype/tensorclass.py
```diff
@@ -291,8 +331,9 @@ def _getitem(self, item):
         isinstance(item, tuple) and all(isinstance(_item, str) for _item in item)
```
apbard, line 331: missing docstring

tensordict/prototype/tensorclass.py
```diff
@@ -302,16 +343,32 @@ def _setitem(self, item, value):
...
+        if not isinstance(val, _accepted_classes):
+            if val and val != value.non_tensordict[key]:
+                raise ValueError(
+                    f"Expecting {repr(val)} for {key} instead got {repr(value.non_tensordict[key])}"
+                )
     self.tensordict[item] = value.tensordict
 
 
 def _repr(self) -> str:
     fields = _all_td_fields_as_str(self.tensordict)
```
apbard, line 356: missing docstring

tensordict/prototype/tensorclass.py
```diff
@@ -471,6 +605,35 @@ def _all_td_fields_as_str(td: TensorDictBase) -> str:
     )
 
 
+def _all_non_td_fields_as_str(src_dict) -> list:
```
apbard, line 608: missing docstring

tensordict/prototype/tensorclass.py
```diff
@@ -471,6 +605,35 @@ def _all_td_fields_as_str(td: TensorDictBase) -> str:
...
+    result = []
+    for key, val in src_dict.items():
+        if not is_tensordict(val):
+            result.append(f"{key}={repr(val)}")
+
+    return result
+
+
+def _validate_non_tensor_data(list_tds) -> bool:
+    list_tds_copy = list_tds.copy()
```
apbard, line 618: missing docstring

tensordict/prototype/tensorclass.py
```diff
@@ -471,6 +605,35 @@ def _all_td_fields_as_str(td: TensorDictBase) -> str:
...
+                if key in tds.non_tensordict and val != tds.non_tensordict[key]:
+                    raise ValueError(
+                        f"{repr(val)} and {repr(tds.non_tensordict[key])} for the attribute "
+                        f"{repr(key)} are not matching"
+                    )
+    return True
+
+
 def _check_td_out_type(field_def):
     if PY37:
```
apbard, line 638: missing docstring

docs/source/reference/prototype.rst
sreevasthav: Agreed ! Let's raise a separate issue to update the documentation with all the use cases?

tensordict/prototype/tensorclass.py
```diff
@@ -198,7 +202,9 @@ def wrapper(self, *args, batch_size, device=None, **kwargs):
         self.tensordict = TensorDict(
```
apbard, line 202: this should be private

tensordict/prototype/tensorclass.py
```diff
@@ -198,7 +202,9 @@ def wrapper(self, *args, batch_size, device=None, **kwargs):
...
             {}, batch_size=batch_size, device=device, _run_checks=False
         )
-        init(self, **{key: _get_typed_value(value) for key, value in kwargs.items()})
+        # To save non tensor data (Nested tensor classes also go here)
+        self.non_tensordict = {}
```
apbard, line 206: this should be private.

tensordict/prototype/tensorclass.py
apbard: ok, please add a comment.

tensordict/prototype/tensorclass.py
```diff
@@ -302,16 +343,32 @@ def _setitem(self, item, value):
         raise ValueError("Invalid indexing arguments.")
     if not isinstance(value, self.__class__):
         raise ValueError("__setitem__ is only allowed for same-class assignement")
+    for key, val in self.non_tensordict.items():
+        if not isinstance(val, _accepted_classes):
+            if val and val != value.non_tensordict[key]:
```
apbard, line 348: this would fail in case of a nestedclass

tensordict/prototype/tensorclass.py
sreevasthav: > > @vmoens I think the set attribute updates are specific to the object and hence if you change data_nest.v = "x" , data.y.v will still be "y"
> 
> are you sure? Since we are not copying by default it should reflect the changes. Anyway it is worth adding a test and document the behaviour
Please ignore the above comments. In the test case there was a clone happening. Yes the above behavior will hold

test/test_tensorclass_nofuture.py
sreevasthav: Added nested case for indexing

test/test_tensorclass_nofuture.py
```diff
@@ -251,21 +277,53 @@ def test_setitem():
     ):
         data[:, [1, 2]] = data.clone()
 
+    # Test cases for non-tensor data
```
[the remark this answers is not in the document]
sreevasthav: That's a good point. Currently, in the code we expect all the non-tensor and nested class data to be same to set the item. We can discuss and finalize on the behavior

test/test_tensorclass_nofuture.py
apbard: yep, exactly because you use the get_attr in the set_attr and, implicitly, assume it works correctly, which maybe not the case.

tensordict/prototype/tensorclass.py
apbard: you marked this as resolved but I don't see any change in the code

test/test_tensorclass_nofuture.py
sreevasthav: Added the assertion for setting a tensor

test/test_tensorclass_nofuture.py
```diff
@@ -43,12 +44,14 @@ def stuff(self):
 class MyData2:
     X: torch.Tensor
     y: torch.Tensor
+    z: list
 
 
 def test_dataclass():
```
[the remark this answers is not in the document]
sreevasthav: That's a good idea. Will raise a separate PR to refactor the test cases as this PR is already a big one to review?

test/test_tensorclass_nofuture.py
```diff
@@ -239,6 +264,7 @@ def test_setitem():
     data = MyData(
         X=torch.ones(3, 4, 5),
         y=torch.zeros(3, 4, 5, dtype=torch.bool),
+        z="test_tensorclass",
         batch_size=[3, 4],
     )
     data[:2] = data[:2].clone()
```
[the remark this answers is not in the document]
apbard: marked as resolved. don't see the fix

tensordict/prototype/tensorclass.py
sreevasthav: Changed the code from before. Nested case is handled above with if and the else only has non_tensor data

test/test_tensorclass_nofuture.py
```diff
@@ -726,32 +739,17 @@ class MyDataParent:
...
-        type(data_nest.tensordict),
-    )
-    data.X = X_clone
-    assert data.tensordict["X"] is X_clone
-    data.z = td_clone
-    assert data.tensordict["z"] is td_clone
+    assert isinstance(data.y, type(data_nest))
+    assert data.y.tensordict is data_nest.tensordict
+    data.X = torch.zeros(3, 4, 5)
+    assert (data.X == torch.zeros(3, 4, 5)).all()
```
apbard, line 749: what happens here to data.y.X?

test/test_tensorclass_nofuture.py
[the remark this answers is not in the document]
sreevasthav: Added the assertions from line 299 to 301. Will add here as well

tensordict/prototype/tensorclass.py
```diff
@@ -340,8 +354,16 @@ def _setitem(self, item, value):
         raise ValueError("Invalid indexing arguments.")
     if not isinstance(value, self.__class__):
         raise ValueError("__setitem__ is only allowed for same-class assignement")
+    # Validating the non-tensor data before setting the item
     for key, val in self.non_tensordict.items():
-        if not isinstance(val, _accepted_classes):
+        # Need the nested tensor class to be the same
+        if is_tensorclass(val):
+            if (val != value.non_tensordict[key]).all():
```
apbard, line 361: shouldn't this be .any()?

tensordict/prototype/tensorclass.py
```diff
@@ -340,8 +354,16 @@ def _setitem(self, item, value):
...
+                raise ValueError(
+                    f"Expecting {repr(val)} for {key} instead got {repr(value.non_tensordict[key])}"
```
apbard, line 363: give a bit more of context of what we are doing. repr could be huge and not so meaningful

test/test_tensorclass_nofuture.py
apbard: don't see `data.y.X=something` anywhere

test/test_tensorclass_nofuture.py
sreevasthav: It will remain torch.ones(3, 4, 5)

tensordict/prototype/tensorclass.py
sreevasthav: Currently, we try to access the tensordict in some of the test cases. @vmoens / @tcbegley Do we want to make them private ?

tcbegley: I personally would be in favour of making them private.

If we do want public access to the underlying tensordict it would be better to add a property and prevent users from writing directly to it.

tensordict/prototype/tensorclass.py
```diff
@@ -210,17 +216,49 @@ def wrapper(self, *args, batch_size, device=None, **kwargs):
...
+            if not isinstance(tensordict[key], _accepted_classes):
+                raise TypeError(
+                    f"{tensordict[key]} from the tensordict is invalid."
+                    f"Accepted values are ({_accepted_classes})."
+                )
+        
+        # Validating non-tensor keys and for key clash
+        tensor_keys = set(tensordict.keys())
+        if non_tensordict is not None:
+            for key in non_tensordict.keys():
```
sreevasthav, line 243: Not restricting the types for non-tensor dict

tensordict/prototype/tensorclass.py
sreevasthav: Yes. Thanks! Changed this

sreevasthav:
> > LGTM, just a couple of things: for nested tensorclass, can you clarify the behaviour a bit? Can we add specific tests for that, like
> > ```python
> >     @tensorclass
> >     class MyDataNest:
> >         X: torch.Tensor
> >         v: str
> >     @tensorclass
> >     class MyDataParent:
> >         W: Any
> >         X: Tensor
> >         z: TensorDictBase
> >         y: MyDataNest
> >         v: str
> > 
> >     batch_size = [3, 4]
> >     if any_to_td:
> >         W = TensorDict({}, batch_size)
> >     else:
> >         W = torch.zeros(*batch_size, 1)
> >     X = torch.ones(3, 4, 5)
> >     X_clone = X.clone()
> >     td = TensorDict({}, batch_size)
> >     td_clone = td.clone()
> >     data_nest = MyDataNest(X=X, v="x", batch_size=batch_size)
> >     v = 'test_tensorclass'
> >     data = MyDataParent(X=X, y=data_nest, z=td, W=W, v=v, batch_size=batch_size)
> >     
> >     assert data.y.v == "x"
> >     data.y.v = "y"
> >     assert data.y.v == "y"
> >     data_nest.v = "x"
> >     assert data.y.v == "x"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (if that makes sense and it's feasible)
> 
> @vmoens I think the set attribute updates are specific to the object and hence if you change data_nest.v = "x" , data.y.v will still be "y"

Please ignore this comment. The behavior  will be as you quoted above

tensordict/prototype/tensorclass.py
apbard: please add a test that covers that case

tensordict/prototype/tensorclass.py
apbard: I see now that we return false if `not isinstance(other, self.__class__)`.
shouldn't be two tensorclasses classes with different keys treated the same way?

vmoens:
I think `None` values should allow for pre-allocation, like a missing value can be pre-allocated with TensorDict:
```python
from typing import Any

import torch

# this works:
s1 = TensorDict({}, [4])
s2 = TensorDict(
    {"X": TensorDict({"X": TensorDict({"X": torch.randn(2)}, [])}, [])}, [])
s1[0] = s2
print(s1)


@tensorclass
class M1:
    X: Any


@tensorclass
class M2:
    X: Any


@tensorclass
class M3:
    X: Any


m1 = M1(M2(M3(X=None, batch_size=[4]), batch_size=[4]), batch_size=[4])
m2 = M1(M2(M3(X=torch.randn(2), batch_size=[]), batch_size=[]), batch_size=[])

print(m1)
print(m2)

# this breaks:
m1[0] = m2
print(m1)

```

sreevasthav:
> pre-allocation

@vmoens  Here for m1, the X will be stored in _non_tensordict as None and _tensordict will be initialized as an empty TensorDict.

```
m1.X.X._tensordict: TensorDict(
    fields={
    },
    batch_size=torch.Size([4]),
    device=None,
    is_shared=False)
 
m1.X.X._non_tensordict: {'X': None}

m1.X.X.X: None
```
 
Now if I set m1[0] = m2

```
m1.X.X._tensordict: TensorDict(
    fields={
        X: Tensor(shape=torch.Size([4, 2]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([4]),
    device=None,
    is_shared=False)
m1.X.X._non_tensordict: {}
m1.X.X.X: tensor([[-1.0844,  0.5641],
        [ 0.0000,  0.0000],
        [ 0.0000,  0.0000],
        [ 0.0000,  0.0000]])
```

What we are doing in set_item is we set the item in _tensordict and del key-value pair from _non_tensordict to make sure that there won't be duplicate keys saved

Let me know if this is the expected behavior?

vmoens:
Yes that's it!

sreevasthav:
> Yes that's it!

Great ! Will add a test case for pre-allocation

sreevasthav:
> This is a brilliant work, I love it! Thanks for taking care of it.
> 
> ## dispatching logic
> Many private tensor-to-tensor methods look the same to me, couldn't we reduce the amount of code by recycling the logic a bit? eg
> 
> ```python
> def _do_smth(tdc, *args, **kwargs):
>     tensordict = torch.do_smth(tdc._tensordict, , *args, **kwargs)
>     non_tensor_dict = tdc._non_tensordict
>     # Cloning nested Tensor classes if any
>     for key, value in non_tensor_dict.items():
>         if is_tensorclass(value):
>             non_tensor_dict[key] = _do_smth(value, , *args, **kwargs)
>     out = tdc.from_tensordict(tensordict, non_tensor_dict)
>     return out
> ```
> 
> Wouldn't a wrapper be capable of handling that?
> 
> I'm just afraid that if we change the logic some day we'll need to manually go through every function and re-code it.
> 
> ## For the docstrings:
> I did not do all the suggestions, I think you got the gist docstrings should look like
> 
> ```
>     """Fist line without break
> 
>     After the first line, an empty line
> 
>     Args:
>         arg1 (int, optional): something to do with it.
>         arg2 (TensorDictBase, optional): notice the empty line before the end.
> 
> """
> ```
> 
> At the risk of being picky I don't see the advantage of many of these docstrings: they will not appear in the doc because they're private methods and they add little to the understanding of the file (eg the behaviour of `_unsqueeze` is self-explanatory when looking at the code and it won't appear in the doc). Happy to leave them of course. (in other cases, like `_all_td_fields_as_str` a docstring, or even simply a comment, are useful).

Added helper functions to handle non_tensor and nested logic for tensor operations. Regarding the documentation, even though "_squeeze" is self explanatory, we handle tensor and non-tensor data differently and I want to add it to that.  I am okay to add comments instead of a docstring. Let me know if you want me to change this

tensordict/prototype/tensorclass.py
sreevasthav: Not relevant any more

vmoens:
> Let me know if you want me to change this

Docstrings are good no worry.
Is there anything outstanding?

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,132 @@ item assignment and many more features.
...
+        batch_size=torch.Size([3, 4]),
+        device=None,
+        is_shared=False),
+    batch_size=torch.Size([3, 4]),
+    device=None,
+    is_shared=False)
+
+
+:obj:`@tensorclass` supports indexing. Internally the tensor objects gets indexed,
+however the non-tensor data remains same
```
vmoens, line 75: remains the same

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,132 @@ item assignment and many more features.
...
+         [1., 1., 1., 1., 1.],
+         [1., 1., 1., 1., 1.]]])
 
+ >>> # Changing nested tensor data
+ >>> data.nested.non_tensordata = "nested_test_changed"
+ >>>print("data.nested.non_tensordata:", repr(data.nested.non_tensordata))
+ data.nested.non_tensordata: 'nested_test_changed'
+
+:obj:`@tensorclass` supports torch functions and the behavior is extened to
+nested tensor classes as well.Note that the operations are strictly restricted
```
vmoens, line 129: `classes as well. Note`

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,132 @@ item assignment and many more features.
...
+         [1., 1., 1., 1., 1.],
+         [1., 1., 1., 1., 1.],
+         [1., 1., 1., 1., 1.]]])
 
+ >>> # Changing nested tensor data
+ >>> data.nested.non_tensordata = "nested_test_changed"
+ >>>print("data.nested.non_tensordata:", repr(data.nested.non_tensordata))
+ data.nested.non_tensordata: 'nested_test_changed'
+
+:obj:`@tensorclass` supports torch functions and the behavior is extened to
```
vmoens, line 128: it does not support all torch functions
we should be clear that it's a subset of torch functions, mostly shape operations, and no algebraic operations.

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,132 @@ item assignment and many more features.
...
+nested tensor classes as well.Note that the operations are strictly restricted
+to tensor data types
```
vmoens, line 130: I think we should say something like
```
@tensorclass supports multiple torch operations over the shape and device of its content, such as `stack`, `cat`, `reshape` or `to(device)`. To get a full list of the supported operations, check the tensordict documentation.
```

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,132 @@ item assignment and many more features.
...
+        nested=None,
+        batch_size=torch.Size([2, 4]),
+        device=None,
+        is_shared=False),
+    batch_size=torch.Size([2, 4]),
+    device=None,
+    is_shared=False)
+
+:obj:`@tensorclass` also supports and get and set attributes. One can also access
+nested tensor class data
```
vmoens, line 98:
```suggestion
:obj:`@tensorclass` also supports setting and resetting attributes, even for nested objects.
```

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,132 @@ item assignment and many more features.
...
+        floatdata=Tensor(shape=torch.Size([6, 4, 5]), device=cpu, dtype=torch.float32, is_shared=False),
+        intdata=Tensor(shape=torch.Size([6, 4, 1]), device=cpu, dtype=torch.int64, is_shared=False),
+        non_tensordata='nested_test_changed',
+        nested=None,
+        batch_size=torch.Size([6, 4]),
+        device=None,
+        is_shared=False),
+    batch_size=torch.Size([6, 4]),
+    device=None,
+    is_shared=False)
```
vmoens, line 153: Let's create an "edge cases" section where we describe less intuitive behaviours, such as `__eq__`, what happens when a `None` value is assigned with a tensor and so forth.

docs/source/reference/prototype.rst
sreevasthav: > > Let me know if you want me to change this
> 
> Docstrings are good no worry. Is there anything outstanding?

Almost done. working on documenting the edge cases

docs/source/reference/prototype.rst
sreevasthav: > Let's create an "edge cases" section where we describe less intuitive behaviours, such as `__eq__`, what happens when a `None` value is assigned with a tensor and so forth.

@vmoens  Adding the following to the edge cases:
1. Equality and inequality operations ( non-tensor data not validated here)
2. Set_item (Non-tensor data is validated at identity instead of equality)
3. cat/stack operations (No validation for non-tensor data in the list  and first tensor class object non-tensor data is considered while returning)
4. Pre-allocation. 

Quick question on pre-allocation. Do we want to tell users how we save and change them internally? I think since its internal we can just add it to the normal operations saying that tensorclass supports pre-allocation instead of an edge case. Wdyt?

sreevasthav: Added it to Edge cases for now

tensordict/prototype/tensorclass.py
sreevasthav: Added the check

test/test_tensorclass_nofuture.py
sreevasthav: Added test case for getattr

docs/source/reference/prototype.rst
sreevasthav: Added few examples and the edge cases for the documentation

docs/source/reference/prototype.rst
```diff
@@ -21,47 +25,230 @@ item assignment and many more features.
...
+
+:obj:`@tensorclass` supports setting an item. However, while setting an item
+the identity check of non-tensor / meta data is done instead of equality to
+avoid performance issues. User needs to make sure that the non-tensor data
+of an item matches with the object to avoid discrepancies.
+
+Here is an example:
+
+While setting an item with different non_tensor data, a UserWarning will be
```
vmoens, line 189:
```suggestion
While setting an item with different ``non_tensor`` data, a :class:`UserWarning` will be
```

vmoens (approved):
LGTM two minor things and we can ship this

apbard (approved):

PR #162, opened 2023-01-10
[Draft][Feature] Simple symbolic tracer for TensorDictModule and TensorDictSequential

tcbegley:
## Description

This PR contains a prototype alternative to `torch.fx.symbolic_trace` that is compatible with `TensorDictModule` and `TensorDictSequential`.

It works by tracing underlying modules, creating a new placeholder for the tensordict input, and replacing existing placeholders with reads from the tensordict.

A nice feature to call out is that when tracing `TensorDictSequential` we trace the entire sequence into a single graph, which means that a forward pass does not result in intermediate reads / writes in the tensordict.

The main outstanding issue at the moment is that it's not clear we can trace / include writing outputs back to the tensordict in the graph, so currently a traced module will return a tuple of tensors corresponding to the out_keys (in the case of `TensorDictSequential` the order should be the order that the keys are written to, which might be non-trivial to predict without working it out). I think the best solution to this is likely going to be writing a custom version of `torch.fx.GraphModule`, which would let us resolve the output issue, and probably also let us handle `tensordict_out` argument, which this code currently makes no attempt to address.

## Example

Here's an example of tracing the `TensorDictSequential` sample code from the documentation

```python
import torch
import torch.nn as nn
from tensordict.nn import TensorDictModule, TensorDictSequential
from tensordict.prototype.fx import symbolic_trace


class Net(nn.Module):
    def __init__(self, input_size=100, hidden_size=50, output_size=10):
        super().__init__()
        self.fc1 = nn.Linear(input_size, hidden_size)
        self.fc2 = nn.Linear(hidden_size, output_size)

    def forward(self, x):
        x = torch.relu(self.fc1(x))
        return self.fc2(x)

class Masker(nn.Module):
    def forward(self, x, mask):
        return torch.softmax(x * mask, dim=1)
    

net = TensorDictModule(
    Net(), in_keys=[("input", "x")], out_keys=[("intermediate", "x")]
)
masker = TensorDictModule(
    Masker(),
    in_keys=[("intermediate", "x"), ("input", "mask")],
    out_keys=[("output", "probabilities")],
)
module = TensorDictSequential(net, masker)


# trace module into GraphModule
gm = symbolic_trace(module)
```

we can inspect the generated code

```python
>>> print(gm.code)


def forward(self, tensordict):
    getitem = tensordict[('input', 'x')]
    _0_fc1 = getattr(self, "0").module.fc1(getitem);  getitem = None
    relu = torch.relu(_0_fc1);  _0_fc1 = None
    _0_fc2 = getattr(self, "0").module.fc2(relu);  relu = None
    getitem_1 = tensordict[('input', 'mask')];  tensordict = None
    mul = _0_fc2 * getitem_1;  getitem_1 = None
    softmax = torch.softmax(mul, dim = 1);  mul = None
    return (_0_fc2, softmax)
```

and check that forward passes work as expected

```python
>>> tensordict = TensorDict(
...     {
...         "input": TensorDict(
...             {"x": torch.rand(32, 100), "mask": torch.randint(2, size=(32, 10))},
...             batch_size=[32],
...         )
...     },
...     batch_size=[32],
... )
>>> tensordict = gm(tensordict)
>>> assert (
...     (0 <= tensordict["output", "probabilities"]) 
...     & (tensordict["output", "probabilities"] <= 1)
... ).all()
```

Finally we can visualise the graph using `torch.fx.passes.graph_drawer.FxGraphDrawer` to see what's going on (requires `pydot`).

```python
g = FxGraphDrawer(gm, "sequential")
with open("graph.svg", "wb") as f:
    f.write(g.get_dot_graph().create_svg())
```

sequential


vmoens:
This is super cool! Thanks for exploring it!

Before we land this:
- [ ] Complete the doc with the examples you gave above
- [ ] test it in conjunction with torch.compile
- [ ] related to the above: write benchmarks

tcbegley:
I've added docs and some simple benchmarks and tests. (The benchmark shows some nice benefits in eliminating intermediate tensordict I/O in a sequential forward pass, though once we solve the issue of returning a repopulated tensordict I expect the advantage to diminish slightly)

Nested sequential tensordict modules are currently broken. I've been working on a fix, though it's a little fiddly. I'll keep at it.

I think we should see what it takes to support the probabilistic case. I haven't thought much about it yet, but I think the same sort of approach should work.

Finally I took a look at `torch.compile`, but since tracing appears to work quite differently there (even though ultimately I believe an FX graph is used as an intermediate representation), so I'm not quite sure how to proceed.

tcbegley:
I took another look at the probabilistic modules, I think it might be challenging to trace because there's a lot of control flow and context managers and so on in there that I don't think `symbolic_trace` can easily handle. Perhaps the best we can do is have the user trace the module they're using and then tag on the probabilistic layer at the end?

vmoens:
> I took another look at the probabilistic modules, I think it might be challenging to trace because there's a lot of control flow and context managers and so on in there that I don't think `symbolic_trace` can easily handle. Perhaps the best we can do is have the user trace the module they're using and then tag on the probabilistic layer at the end?

happy with that. In the end what we should aim for is (1) solving whatever can to make tracing and compilation efficient for users, (2) massage that to make it easily accessible, (3) document the recipes, (4) identify breakages and give solutions in the doc

vmoens:
@tcbegley I think this is mergeable, but it just needs a bit of cleanup. What's your opinion?

tcbegley:
Yeah, I think since it's clearly marked as prototype it's probably safe enough to merge.

I'll have to refamiliarise myself a little with it. Do you have a clear sense of what needs tidying?

PR #165, opened 2023-01-12
[BugFix] Create TensorClass by modifying decorated class in-place

tcbegley:
## Description

This PR updates how we create dataclasses. Instead of locally defining a class and returning that, we modify the decorated class in place. The motivation is:

1. Returning a locally defined class leads to errors when we try to pickle instances of those classes.
2. The new behaviour is closer to the behaviour of `dataclasses.dataclass` and hence less surprising to users.

While refactoring I noticed a bug with how default_factory values were handled which I have fixed.

## Test

```python
import pickle

import torch
from tensordict.tensordict import assert_allclose_td

@tensorclass
class Data:
    value: torch.Tensor

    def square_value(self):
        return self.value ** 2

d = Data(value=torch.rand(5, 10), batch_size=[5])

with open("test.pkl", "wb") as f:
    pickle.dump(d, f)

with open("test.pkl", "rb") as f:
    d2 = pickle.load(f)

assert_allclose_td(d.to_tensordict(), d2.to_tensordict())
assert isinstance(d2, Data)
```

### DIFF
--- a/test/test_tensorclass.py
+++ b/test/test_tensorclass.py
@@ -60,10 +60,11 @@ def test_type():
         y=torch.zeros(3, 4, 5, dtype=torch.bool),
         batch_size=[3, 4],
     )
-    assert isinstance(data, MyDataUndecorated)
     assert isinstance(data, MyData)
     assert is_tensorclass(data)
     assert is_tensorclass(MyData)
+    # we get an instance of the user defined class, not a dynamically defined subclass
+    assert type(data) is MyDataUndecorated
 
 
 @pytest.mark.parametrize("device", get_available_devices())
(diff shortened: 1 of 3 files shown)


test/test_tensorclass.py
```diff
@@ -46,6 +58,7 @@ def test_type():
         y=torch.zeros(3, 4, 5, dtype=torch.bool),
         batch_size=[3, 4],
     )
+    assert isinstance(data, MyDataUndecorated)
```
vmoens, line 61: wonderful that it works!

vmoens (approved):
Very cool stuff!
Do we want to add a test with multiprocessing too?

test/test_tensorclass.py
tcbegley: I realised that this actually works on `main`, because the undecorated class is a subclass of the dynamically generated class. Updated the test to check that `type(data) is MyDataUndecorated` which is what I was really trying to check for.

vmoens (approved):

PR #166, opened 2023-01-13
[Docs] Update to ImageNet tutorial

tcbegley:
Update the ImageNet tutorial in the docs, with benchmarks from the full data and using multiple workers in "standard" dataloaders for fairer comparison.


tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -117,30 +204,40 @@ class ImageNetData:
...
+        dl = DataLoader(dataset, batch_size=batch, num_workers=NUM_WORKERS)
+        i = 0
+        pbar = tqdm.tqdm(total=len(dataset))
+        for image, target in dl:
+            _batch = image.shape[0]
+            pbar.update(_batch)
+            data[i : i + _batch] = cls(
+                images=image, targets=target, batch_size=[_batch]
+            )
+            i += _batch
```
tcbegley, line 229: Does the `tqdm` stuff here make this too confusing?

tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -153,59 +250,132 @@ def __len__(self):
...
 
-t0 = time.time()
-for batch in val_dataloader_tc:
-    image, target = batch.images.contiguous(), batch.targets.contiguous()
-print(f"One iteration over tensorclass val data. Time: {time.time() - t0:4.4f}s")
+##############################################################################
+# Results from ImageNet
+# ---------------------
+#
+# Repeating the above on full-size ImageNet data, we get the following results
```
tcbegley, line 372: should we add information about the environment, e.g. 2x A100 GPUs or whatever?

vmoens: good idea

tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -153,59 +250,132 @@ def __len__(self):
...
+#
+#  | One iteration over dataloader done! Rate: 2984.8455 fps, time:  428.7951s
+#  | One iteration over tensorclass dataloader done! Rate: 6457.9865 fps, time:  198.1867s
```
vmoens, line 375: I have new results using a mp collector for the tensordict version:


```
One iteration over dataloader done! Rate: 2969.6382 fps, time:  430.9909s
One iteration over tensorclass dataloader done! Rate: 27273.2961 fps, time:  46.9282s
One iteration over val data done! Rate: 2367.5322 fps, time:  20.5784s
One iteration over tensorclass val data done! Rate: 19623.9993 fps, time:  2.4827s
```

We are 10x as fast!

vmoens (commented):
Excellent work!

Let's:
- change to mp collection for the tc version too
- change the global variables
- update the results
- check the GPU utilization (I will have a go at this)

It would be nice if we could produce a figure :)

tutorials/sphinx_tuto/tensorclass_imagenet.py
vmoens: Do you mean: do we want it there?
The thing is that otherwise you can wait for some minutes without a progress bar which is a bit mysterious for the users...

tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -153,59 +250,132 @@ def __len__(self):
...
 
-batch_size = 64
-train_dataloader = DataLoader(train_data, batch_size=batch_size)
-val_dataloader = DataLoader(val_data, batch_size=batch_size)
+
+##############################################################################
+# ``DataLoader`` has support for multiple workers loading data in parallel. The
+# tensorclass dataloader will use just one worker, but load data in batches.
+
+batch_size = 8
```
vmoens, line 279: On the cluster I used 128 and it worked fine

tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -153,59 +250,132 @@ def __len__(self):
...
+    val_data,
+    batch_size=batch_size,
+    num_workers=NUM_WORKERS,
+)
 
 train_dataloader_tc = DataLoader(
-    train_data_tc, batch_size=batch_size, collate_fn=lambda x: x
+    train_data_tc,
+    batch_size=batch_size,
+    collate_fn=Collate(collate_transform, device),
```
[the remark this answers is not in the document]
vmoens: Also I think that here we should not miss the occasion of pointing at how convenient it is: our collate fn is essentially an nn.module that will work just fine. Collate fn can be nasty to implement!
Ours could even be moved to the training loop as a module that would be part of the model (eg for deployment).

tutorials/sphinx_tuto/tensorclass_imagenet.py
vmoens: Just saw that you corrected that below :)

vmoens:
@tcbegley working on a distributed version of this usin torch RPC. Perf is roughly the same as the multiprocess + tensorclass.

We should make a figure with the collection speed with 
- torchvision + torch
- tensorclass, single proc
- tensorclass multiproc
- tensorclass distributed

The distributed example is a bit different from this tutorial, and perhaps harder to fit in a tutorial. (since it's very rough I made this a draft PR to make sure we're not rushing into anything).
I append it in the `benchmarks`, but we can move it to tutorials if you think appropriate.

benchmarks/distributed/dataloading.py
```diff
@@ -0,0 +1,363 @@
...
+            pbar.update(BATCH_SIZE)
+
+        t0 = time.time()
+        iteration = 0
+        total = 0
+        while total < len_data:
+            batch = _prefetch_queue.popleft().wait()
+            i = iteration % (self.world_size - 1)
+            iteration += 1
+            _prefetch_queue.append(
```
vmoens, line 231: TODO: put this under a condition (we don't want to append if there is no data to collect)

benchmarks/distributed/dataloading.py
```diff
@@ -0,0 +1,363 @@
...
+        print("Creating DataNode object")
+        self.rank = rank
+        self.id = rpc.get_worker_info().id
+        train_ref = rpc.get_worker_info(f"{TRAINER_NODE}")
+        self.train_ref = rpc.remote(
+            train_ref,
+            get_trainer,
+        )
+        self.batch_size = batch_size
+        device = f"cuda:{rank}"
```
vmoens, line 282: @tcbegley Each proc is on its own cuda. It's mainly to avoid OOM errors that we don't have in the multiproc case.
So, I guess there's something a bit fishy going on here, since the amount of computation is the same.

vmoens:
Cc @H-Huang for the distributed on device collection and transforms.
Cc @shagunsodhani, @fmassa, @NicolasHug, @Nicolas-Dufour and @FrancescoSaverioZuppichini for advice and feedback

tutorials/sphinx_tuto/tensorclass_imagenet.py
tcbegley: Yeah, 128 works great for the full imagenet, but caused some problems in the benchmarking code on the hymenoptera data (`t0` is only defined after certain number of batches which was not reached).

tutorials/sphinx_tuto/tensorclass_imagenet.py
tcbegley: Yeah that's fair. Let's leave it then. I guess most people are probably pretty familiar with it anyway.

tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -17,28 +17,45 @@
...
 
+NUM_WORKERS = int(os.environ.get("NUM_WORKERS", "4"))
+# sphinx_gallery_start_ignore
+# this example can be run locally or in the tensordict CI on the small hymenoptera
+# subset of imagenet, but we also want to be able to compare to runs on larger subsets
+# of imagenet which would be impractical to run in CI. If this script is run with the
+# environment variable RUN_ON_CLUSTER set, then we set everything to run on a larger
+# subset of imagenet. the fraction of images can be set with the FRACTION environment
+# variable, we use the first `len(dataset) // FRACTION` images. Default is 10.
+RUN_ON_CLUSTER = os.environ.get("RUN_ON_CLUSTER", False)
```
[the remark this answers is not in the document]
tcbegley: I see you already added smth, but thought I'd flag [`distutils.util.strtobool`](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool). A bit weird that we'd add a distutils import I guess, but it does exactly what we want.

tutorials/sphinx_tuto/tensorclass_imagenet.py
```diff
@@ -50,13 +51,7 @@
...
 # subset of imagenet. the fraction of images can be set with the FRACTION environment
 # variable, we use the first `len(dataset) // FRACTION` images. Default is 10.
-RUN_ON_CLUSTER = os.environ.get("RUN_ON_CLUSTER", False)
-if RUN_ON_CLUSTER in ("0", "False", False):
-    RUN_ON_CLUSTER = False
-elif RUN_ON_CLUSTER in ("1", "True", True):
-    RUN_ON_CLUSTER = True
-else:
-    raise NotImplementedError(RUN_ON_CLUSTER)
+RUN_ON_CLUSTER = strtobool(os.environ.get("RUN_ON_CLUSTER", "False"))
```
vmoens, line 54: Sure much better

benchmarks/distributed/dataloading.py
```diff
@@ -0,0 +1,363 @@
...
+                    ),
+                )
+                print(f"Connected to data node {data_info}")
+                time.sleep(5)
+                self.datanodes.append(data_rref)
+                return
+            except Exception as e:
+                print(f"Failed to connect to data node: {e}")
+                time.sleep(RETRY_DELAY_SECS)
+
```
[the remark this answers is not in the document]
vmoens: I'm ok with a local external dependency yes

benchmarks/distributed/dataloading.py
```diff
@@ -0,0 +1,379 @@
...
+"""
+
+##############################################################################
+# This file provides an example of distributed dataloading with tensorclass.
+# It can be run with :obj:`torchrun`. Here is an example of the slurm config to
+# use, although torchrun is compatible with any other scheduler too:
+#
+#  .. code-block::
+#
+#     #!/bin/bash
```
vmoens, line 18: thinking about it, it would be clearer to have a `dataloading.sh` no? It wouldnt hurt to keep it there

tcbegley: Yeah, that would be good, and then it's easier to run / test too.

vmoens:
I see 2 outstanding issues for this PR:
- [ ] logging the new distributed results 
- [ ] writing a little something on how to save the cached dataset with torchsnapshot

benchmarks/distributed/dataloading.py
```diff
@@ -53,38 +43,31 @@
...
+parser.add_argument(
+    "--wandb_key",
+    type=str,
+    default="",
+)
 torch.cuda.set_device(0)
 
 RETRY_LIMIT = 2
-NUM_WORKERS = 16
+NUM_WORKERS = 64
```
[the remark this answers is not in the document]
vmoens: These cores are used only to build/preprocess the dataset "once and for all" using plain torch dataloader (ie you would do that once, and hopefully only once per paper :p ). We include only the deterministic transforms, and leave the random ones for the "real" training loop.
Then you have a contiguous mem representation of your data on disk that you can play with.

Later (during the "real" training loop), you use only as many procs as your world size.

We use 64 because we run it often and it's slightly faster than 16 or 32, albeit not significantly. 

Now that being said we could lower this number even to 1 or 0, it would just make running the benchmark longer but the later stage would still be super fast. FYI we get to 30K fps with this (>10x faster than traditional transforms).
Does that make sense?

With @tcbegley we wanted to add a comment about the "cost" of using gpus for preproc: here we have 1GPU / subcollector. I'm a bit surprised by how much mem each proc consumes, but even there: if your training loop with imagenet is dataloading-bound, you could speed it up by -- say -- 5x using what we present here, thereby reducing the overall cost. Intuitively, even if your exp costs more, you run it for a shorter time (because you use your resources better), and you also gain in dev/ops time.

benchmarks/distributed/dataloading.py
```diff
@@ -10,20 +10,10 @@
...
-#
-#     #!/bin/bash
-#     #SBATCH --gpus 8
-#     #SBATCH -p train
-#     #SBATCH -o slurm-%A_%a.out
-#     #SBATCH --exclusive
-#     #SBATCH -c 96
-#     NUM_TRAINERS=5
-#     srun torchrun --standalone --nnodes=1 --nproc_per_node=$NUM_TRAINERS dataloading.py --world_size $NUM_TRAINERS
+# It can be run by simply calling `python dataloading.py --world_size=5` (optionally
```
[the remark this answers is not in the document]
vmoens: Yes i can run it with 2 gpus, one for data collection and one for training. Let me try that

benchmarks/distributed/dataloading.py
```diff
@@ -0,0 +1,446 @@
...
+            init_rpc, ((rank, name, world_size) for rank, name in enumerate(names))
+        )
+        pool.starmap(
+            func,
+            (
+                (rank, world_size, args, train_data_tc)
+                for rank, name in enumerate(names)
+            ),
+        )
+        pool.apply_async(shutdown)
```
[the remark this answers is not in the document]
tcbegley: Yeah, that would be neat!

benchmarks/distributed/dataloading.py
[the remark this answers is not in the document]
shagunsodhani: Thanks for the clarification

vmoens:
I added the saving option for the distributed case. Now i will run the benchmarks, update the fig and we're good to go!

docs/source/reference/tensordict.rst
```diff
@@ -16,6 +16,39 @@ regular pytorch tensors.
     SubTensorDict
     LazyStackedTensorDict
 
+Memory-mapped tensors
+---------------------
+
+:obj:`tensordict` offers the :obj:`MemmapTensor` primitive which allows you to work
+with tensors stored in physical memory in a handy way. The main advantages of :obj:`MemmapTensor`
+are its easiness of construction (no need to handle the storage of a tensor), tbe possibility to
```
tcbegley, line 24:
```suggestion
are its easiness of construction (no need to handle the storage of a tensor), the possibility to
```

docs/source/reference/tensordict.rst
```diff
@@ -16,6 +16,39 @@ regular pytorch tensors.
...
+efficient indexing of stored tensors.
+
+If all workers have access to the same storage, passing a :obj:`MemmapTensor` will just consist in passing
+a reference to a file on disk plus a bunch of extra meta-data for reconstructing it when
+sent across processes or workers on a same machine (both in multiprocess and distributed settings).
+The same goes with indexed memory-mapped tensors.
+
+Indexing memory-mapped tensors is much faster than loading several independent files from
+the disk and does not require to load the full content of the array in memory.
+However, physical storage of pytorch tensors should not be any different:
```
tcbegley, line 35:
```suggestion
However, physical storage of PyTorch tensors should not be any different:
```

PR #167, opened 2023-01-16
[Feature] TorchSnapshot compatibility

vmoens:
## Description

Ensures compatibility with torchsnapshot, especially with memmap-tensors. 

To test this PR:
```python
import torchsnapshot
from tensordict import TensorDict, MemmapTensor
import torch
from torchsnapshot import StateDict, Snapshot
import uuid

td = TensorDict({"a": torch.randn(3), "b": TensorDict({"c": torch.randn(3, 1)}, [3, 1])}, [3])
td.memmap_()
assert isinstance(td["b", "c"], MemmapTensor)

app_state = {
    "state": StateDict(tensor=td.state_dict())
}
snapshot = Snapshot.take(app_state=app_state, path=f"/tmp/{uuid.uuid4()}")


td_dest = TensorDict({"a": torch.zeros(3), "b": TensorDict({"c": torch.zeros(3, 1)}, [3, 1])}, [3])
td_dest.memmap_()
assert isinstance(td_dest["b", "c"], MemmapTensor)
app_state = {
    "state": StateDict(tensor=td_dest.state_dict())
}
snapshot.restore(app_state=app_state)

print(td, td_dest)
assert (td_dest == td).all()
assert (td_dest["b"].batch_size == td["b"].batch_size)
assert isinstance(td_dest["b", "c"], MemmapTensor)
```

TODO:
- [x] Document the feature
- [ ] Test the feature


tcbegley (commented):
LGTM so far. Added a few minor comments.

vmoens:
@tcbegley I added some doc.
Also I had to add `__eq__` and `__ne__`. Do you think I should do a separate PR with them?

docs/source/saving.rst
```diff
@@ -0,0 +1,174 @@
...
+=========================================
+
+While we can just save a tensordict with :obj:`torch.save(my_tensordict)`, this
+will create a single file with the whole content of the data structure.
+One can easily imagine situations where this is sub-optimal!
+
+In general, we recommend using `torchsnapshot <https://github.com/pytorch/torchsnapshot>`_.
+TorchSnapshot will save each of your tensors independently, with a data structure that
+mimics the one of your tensordict or tensorclass. Moreover, TensorDict has naturally
+buit-in the tools necessary for saving and loading huge datasets on disk without
```
tcbegley, line 11: buit-in -> built-in

tensordict/prototype/tensorclass.py
```diff
@@ -332,6 +334,18 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
     self.tensordict._batch_size_setter(new_size)
 
 
+def __eq__(self, other):
+    if is_tensorclass(other):
```
tcbegley, line 338: Should we check that `other` is in fact an instance of the current class, i.e.

```python
if isinstance(other, self.__class__):
    ...
```

Otherwise it's possible that you could have two different dataclasses with the same underlying tensordict that passes this equality check.

tensordict/prototype/tensorclass.py
```diff
@@ -332,6 +334,18 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+        return self.tensordict.__eq__(other.tensordict)
+    return self == other
+
+
+def __ne__(self, other):
+    if is_tensorclass(other):
```
tcbegley, line 344: ditto

tcbegley (approved):
LGTM!

I think it's fine to include the `__eq__` and `__ne__` here, since they're needed by these changes and they're relatively simple, small changes.

tensordict/prototype/tensorclass.py
vmoens: good point

vmoens:
@tcbegley we get a `No source for code: '/tmp/tmplhl4mh7r/_remote_module_non_scriptable.py'.` at the end of each run, which I think comes from torchsnapshot (since it's the only thing I added in the installation scripts).
I can ask those guys what they think of this bug, but since it breaks our CI: do we want to land this?

tcbegley:
Let's wait until you hear back I guess. I think it's low risk that there's a problem with the library rather than simply something wrong with the test environment, but I think this change is not super urgent, and anything we merge now will make it into a couple of nightly releases over the weekend, so we may as well double check that everything is fine.

vmoens:
> Let's wait until you hear back I guess. I think it's low risk that there's a problem with the library rather than simply something wrong with the test environment, but I think this change is not super urgent, and anything we merge now will make it into a couple of nightly releases over the weekend, so we may as well double check that everything is fine.

seems like `coverage xml -i` can get rid of this

https://stackoverflow.com/questions/2386975/no-source-for-code-message-in-coverage-py

PR #172, opened 2023-01-19
[Feature] TensorClass `__post_init__` support

tcbegley:
## Description

This PR makes it possible to write a `__post_init__` method on a tensorclass, as one can with [dataclasses](https://docs.python.org/3/library/dataclasses.html?highlight=dataclass#post-init-processing).

Note however that we do not currently support setting derived values, atm this is since the tensorclass permits only setting attributes in `__dataclass_fields__` which is determined from the class definition. The main challenge with lifting this restriction is that we would not have type information for the derived attributes, which is required elsewhere in the code to determine the types of return values. I don't personally see a way around this atm...

Example

```python
import torch


@tensorclass
class MyDataPostInit:
    X: torch.Tensor
    y: torch.Tensor

    def __post_init__(self):
        assert (self.X > 0).all()
        assert self.y.abs().max() <= 10
        # modifying existing fields is fine
        self.y = self.y.abs()


y = torch.clamp(torch.randn(3, 4), min=-10, max=10)
data = MyDataPostInit(X=torch.rand(3, 4), y=y, batch_size=[3, 4])
assert (data.y == y.abs()).all()

# this results in an assertion error
MyDataPostInit(X=-torch.ones(2), y=torch.rand(2), batch_size=[2])
```

cc @apbard


test/test_tensorclass.py
```diff
@@ -506,22 +506,40 @@ class MyDataParent:
...
+
+
 def test_default():
     @tensorclass
     class MyData:
         X: torch.Tensor = None  # TODO: do we want to allow any default, say an integer?
         y: torch.Tensor = torch.ones(3, 4, 5)
 
     data = MyData(batch_size=[3, 4])
-    assert data.__dict__["y"] is None
```
tcbegley, line 516: This was ultimately an implementation detail that changed and caused the tests to fail. `"y"` no longer exists in `data.__dict__`.

tensordict/prototype/tensorclass.py
```diff
@@ -238,7 +230,6 @@ def wrapper(self, key, value):
         if type(value) in CLASSES_DICT.values():
             value = value.__dict__["tensordict"]
         self.__dict__["tensordict"][key] = value
-        assert self.__dict__["tensordict"][key] is value
```
tcbegley, line 241: This check was failing both for `None` values, but also tensors that were moved to a new device when assigned to the tensordict, so I decided to just delete it. lmk though if you think we need to retain some kind of input validation.

apbard: I agree with you. we can drop that check.

vmoens:
@tcbegley sorry, but I did not get when things would break in your comment. Can you give an example?

tcbegley:
> Can you give an example?

Sure

```python
import torch


@tensorclass
class Data:
    x: torch.Tensor
    y: torch.Tensor

    def __post_init__(self):
        self.z = self.x + self.y

d = Data(x=torch.rand(10), y=torch.rand(10), batch_size=[10])
```

That will fail because `"z"` is not in `expected_keys`, but even if we were to solve that problem, I wonder if there's going to be problems around not having a field definition / type annotation which we use in various places, e.g. [here](https://github.com/pytorch-labs/tensordict/blob/main/tensordict/prototype/tensorclass.py#L397)

vmoens:
> That will fail because `"z"` is not in `expected_keys`

But this will work right?

```python
import torch


@tensorclass
class Data:
    x: torch.Tensor
    y: torch. Tensor
    z: Any

    def __post_init__(self):
        self.z = self.x + self.y

d = Data(x=torch.rand(10), y=torch.rand(10), batch_size=[10])

```

tcbegley:
I gives me an error, but I think it could be made to work. Currently it's failing because we try to put `dataclasses._MISSING_TYPE` into the tensordict and it doesn't know what to do with it. But we could handle that.

vmoens:
> I gives me an error, but I think it could be made to work. Currently it's failing because we try to put `dataclasses._MISSING_TYPE` into the tensordict and it doesn't know what to do with it. But we could handle that.

My opinion (do you guys agree?) is that the purpose of tensorclass is to have an explicit list of content before instantiation. If someone violates that, we just need to handle the error in an informative way (essentially telling people that assigning undefined values is not permitted). 
@apbard what do you think? Is this something that you'd find intuitive enough?

apbard:
> My opinion (do you guys agree?) is that the purpose of tensorclass is to have an explicit list of content before instantiation. If someone violates that, we just need to handle the error in an informative way (essentially telling people that assigning undefined values is not permitted). 

Agree, your example with `z: Any` should work. Declaration in __post_init_ not. 

@tcbegley we should add both use-cases to unit-tests

tcbegley:
> My opinion (do you guys agree?) is that the purpose of tensorclass is to have an explicit list of content before instantiation. If someone violates that, we just need to handle the error in an informative way (essentially telling people that assigning undefined values is not permitted).

I also agree with this. Though in that case I'm not sure we should support this pattern

```python
@tensorclass
class Data:
    x: torch.Tensor
    y: torch. Tensor
    z: Any

    def __post_init__(self):
        self.z = self.x + self.y

d = Data(x=torch.rand(10), y=torch.rand(10), batch_size=[10])
```

An analogous dataclass would throw a `TypeError` on instantiation because the user hasn't supplied an argument for `z` which has no default. I think if we require all features to be listed up-front, then they should all be passed to the constructor also.

If the user really doesn't want to pass `z` to the constructor because it's supposed to be derived from `x` and `y` then there is always the option to do the following

```python
@tensorclass
class Data:
    x: torch.Tensor
    y: torch.Tensor

    @property
    def z(self):
        return self.x + self.y
```

Which I think is much clearer and has the added benefit of being up to date if `x` and `y` change. You could even use `functools.cached_property` if they won't change and computation is something more expensive than addition

tcbegley:
Yeah, sort of. That example would run under my proposed changes, but it's not a pattern I would necessarily encourage. I think passing in an arbitrary object only to have it get overwritten no matter the value is not ideal.

But yes, my main argument is that we should try to stick to dataclass-like behaviour where possible.

test/test_tensorclass.py
```diff
@@ -67,6 +68,14 @@ def test_type():
     assert type(data) is MyDataUndecorated
 
 
+def test_signature():
+    sig = inspect.signature(MyData)
+    assert list(sig.parameters) == ["X", "y", "batch_size", "device", "_tensordict"]
+
+    with pytest.raises(TypeError, match="missing 2 required positional arguments"):
```
apbard, line 75: I would probably add also other cases:
with just y
with tensordict
and with/without batch_size

tcbegley: Good idea, added!

apbard: you missed the x+y but without one of batch_size and device that should not raise in all cases

tcbegley: I figured that those working cases should be adequately covered (implicitly at least) by all the other test cases. Can add an explicit one if you think it's required.

apbard: I would say "better explicit than implicit" and I would add them. But I agree that from the coverage point of view they should be indirectly tested in other tests

apbard:
thanks to #175 I noticed that we miss one test: a class with __post_init_ initialised by tensordict. Currently it would fail as tensordict is filled *after* the init

vmoens:
> thanks to #175 I noticed that we miss one test: a class with custom __post_init_ that gets initialised with a tensordict. Currently this test would fail as self.tensordict is filled _after_ the init

Is this something we need to address before landing this?

tensordict/prototype/tensorclass.py
vmoens: + we don't want `assert` in the code base, these should be kept for tests only

test/test_tensorclass.py
```diff
@@ -67,6 +68,34 @@ def test_type():
     assert type(data) is MyDataUndecorated
 
 
+def test_signature():
```
vmoens, line 71: Nice!

vmoens (approved):
LGTM
@apbard let me know if you want your last comment to be addressed before merging

apbard:
> LGTM @apbard let me know if you want your last comment to be addressed before merging

yes, I think we should. In fact, we lack coverage for that use-case and it would also fail.

tcbegley:
We should definitely add a test. 

It seems to me @apbard that #175 resolves that issue right? I think the bigger problem then is that merging #175 is going to cause havoc with the non-tensor data PR that is close to being finished.

Perhaps @vmoens if you're happy with #175 in principle, we can merge into this branch, add the test, and then work on adapting to the pending non-tensor data changes.

tensordict/prototype/tensorclass.py
```diff
@@ -158,56 +159,75 @@ def __torch_function__(
...
+        # bypass initialisation. this means we don't incur any overhead creating an
+        # empty tensordict and writing values to it. we can skip this because we already
+        # have a tensordict to use as the underlying tensordict
+        tc = cls.__new__(cls)
+        tc.__dict__["tensordict"] = tensordict
+        # since we aren't calling the dataclass init method, we need to manually check
+        # whether a __post_init__ method has been defined and invoke it if so
+        if hasattr(tc, "__post_init__"):
+            tc.__post_init__()
+        return tc
```
tcbegley, line 228: I added this to reduce overhead when constructing from `TensorDict`. We bypass the regular constructor which means we don't create a tensordict or have to set the attributes. The only thing we need to make sure we do is run the `__post_init__` method if it exists because we're no longer invoking the dataclass' init method.

I think this is ok, but if you can think of any edge cases I might have missed please let me know!

See relevant discussion here: https://github.com/pytorch-labs/tensordict/pull/175#discussion_r1084278607

vmoens: Happy with that solution

PR #175, opened 2023-01-20
[Refactor] TensorClass drop _tensordict argument in constructor

tcbegley:
## Description

This PR refactors the generated init method of tensorclasses. The motivation was to try and make the signature less confusing. 

Previously `batch_size` was a positional or keyword argument with default value `None`, however, usually if `batch_size` was not specified, this would lead to a `ValueError`, so during normal instantiation `batch_size` was a required argument.

However, `batch_size` was **not** required if the user constructed a tensorclass by passing a `TensorDict` via the `_tensordict` argument.  Hence it was not possible for us to make `batch_size` a required argument.

The changes:
- `batch_size` is now a required keyword-only argument. If you don't specify it, Python will raise a `TypeError`.
- There is no longer a `_tensordict` argument in tensorclass constructors.
- Constructing a tensorclass from a tensordict is still possible, but like this: `tc = MyTensorClass.from_tensordict(td)`.

I've opened this as a separate PR to get specific feedback on whether we want to adopt this approach.

CC @apbard

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -124,8 +124,8 @@ def tensorclass(cls: T) -> T:
                 f"Attribute name {attr} can't be used with @tensorclass"
             )
 
-    cls.__init__ = _init_wrapper(cls.__init__, expected_keys)
-    cls.from_tensordict = classmethod(_from_tensordict)
+    cls.__init__ = _init_wrapper(cls.__init__)
+    cls.from_tensordict = classmethod(_from_tensordict_wrapper(expected_keys))
     cls.__torch_function__ = classmethod(__torch_function__)
     cls.__getstate__ = _getstate
     cls.__setstate__ = _setstate
@@ -158,7 +158,7 @@ def tensorclass(cls: T) -> T:
     return cls
 
 
-def _init_wrapper(init, expected_keys):
+def _init_wrapper(init):
     init_sig = inspect.signature(init)
     params = list(init_sig.parameters.values())
     # drop first entry of params which corresponds to self and isn't passed by the user
@@ -200,10 +200,18 @@ def _init_wrapper(init, expected_keys):
     return wrapper
 
 
-def _from_tensordict(cls, tensordict):
-    tc = cls(**tensordict, batch_size=tensordict.batch_size)
-    tc.__dict__["tensordict"] = tensordict
-    return tc
+def _from_tensordict_wrapper(expected_keys):
+    def wrapper(cls, tensordict):
+        if not all(key in expected_keys for key in tensordict.keys()):
+            raise ValueError(
+                f"Keys from the tensordict ({set(tensordict.keys())}) must "
+                f"correspond to the class attributes ({expected_keys})."
+            )
+        tc = cls(**tensordict, batch_size=tensordict.batch_size)
+        tc.__dict__["tensordict"] = tensordict
+        return tc
+
+    return wrapper
 
 
 def _getstate(self):
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -125,7 +125,7 @@ def tensorclass(cls: T) -> T:
             )
 
     cls.__init__ = _init_wrapper(cls.__init__, expected_keys)
-    cls._build_from_tensordict = classmethod(_build_from_tensordict)
+    cls.from_tensordict = classmethod(_from_tensordict)
     cls.__torch_function__ = classmethod(__torch_function__)
     cls.__getstate__ = _getstate
     cls.__setstate__ = _setstate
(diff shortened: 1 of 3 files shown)


apbard (commented):
documentation needs to be updated with constructor signature and the arguments

tensordict/prototype/tensorclass.py
```diff
@@ -165,65 +165,45 @@ def _init_wrapper(init, expected_keys):
...
+        inspect.Parameter("device", inspect.Parameter.KEYWORD_ONLY, default=None),
     ]
     wrapper.__signature__ = init_sig.replace(parameters=params + new_params)
 
     return wrapper
 
 
-def _build_from_tensordict(cls, tensordict):
-    return cls(_tensordict=tensordict)
+def _from_tensordict(cls, tensordict):
```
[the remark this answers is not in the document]
tcbegley: Can you say a bit more about how you imagine this working?

tensordict/prototype/tensorclass.py
```diff
@@ -165,65 +165,45 @@ def _init_wrapper(init, expected_keys):
...
+    tc = cls(**tensordict, batch_size=tensordict.batch_size)
+    tc.__dict__["tensordict"] = tensordict
+    return tc
```
apbard, line 206:
```suggestion
def _from_tensordict(cls, tensordict, copy=True):
    #TODO assert tensordict is tensordict
    if copy:
        tc = cls(**tensordict, batch_size=tensordict.batch_size)
    else:
            input_dict = {key: None for key in tensordict.keys()}
            init(self, **input_dict, batch_size=tensordict.batch_size)
            self.tensordict = _tensordict
    tc.__dict__["tensordict"] = tensordict
    return tc
```

tensordict/prototype/tensorclass.py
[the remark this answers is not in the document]
apbard: see [here](https://github.com/pytorch-labs/tensordict/pull/175/files#r1082814475).

tensordict/prototype/tensorclass.py
tcbegley: `init` here is the dataclass' `__init__`?

I see where you're doing with this, just a bit awkward because currently the internals of `_from_tensordict` don't have access to `init`. We'll have to rewrite it as a wrapper of some sort.

apbard: wouldn't calling `cls.__init__` be enough?

tcbegley: But that's the tensorclass' init method right? So that's effectively what I'm currently doing anyway with

```python
tc = cls(**tensordict, batch_size=tensordict.batch_size)
```

apbard: oh, my bad...in this PR we are still overwriting `cls` instead of using `datacls`.
you are right then, we need to do a wrapper like other functions

tensordict/prototype/tensorclass.py
```diff
@@ -158,72 +158,60 @@ def __torch_function__(
...
-def _build_from_tensordict(cls, tensordict):
-    return cls(_tensordict=tensordict)
+def _from_tensordict_wrapper(expected_keys):
+    def wrapper(cls, tensordict):
+        if not all(key in expected_keys for key in tensordict.keys()):
+            raise ValueError(
+                f"Keys from the tensordict ({set(tensordict.keys())}) must "
+                f"correspond to the class attributes ({expected_keys})."
+            )
+        tc = cls(**tensordict, batch_size=tensordict.batch_size)
```
vmoens, line 210: isn't this a bit overkilling it?
IIUC we deconstruct the tensordict, reconstruct another one, then replace it with the old one. Since the construction is quite time consuming, do you think we can find another solution?

apbard: my very same objection, but @tcbegley measured the overhead and seems negligible. Do you have a specific test case in mind that we could try to double check the benchmark?

vmoens: The overhead is mainly due to the checks, so if we could do the checks only once it'd be cool (`TensorDIct(dictionary, batch_size, device, _run_checks=False)`).
Given that the tensordict that is provided is presumably already checked we should be good no?

apbard: off-topic: why the `_` in a parameter name? It should be used only for protected methods/variables

vmoens: Exactly for that: we don't want users to call that. We may refactor this in the future to hide it a bit more.
You should use that only if you build a tensordict out of another tensordict and the checks have been done, which will only occur in with dedicated methods.
I'm open to suggestions to make it cleaner :)

vmoens: See this example:
```python
td = TensorDict({"a": torch.randn(3), "b": torch.randn(2)}, [3], _run_checks=False) # passes 
td[2]  # breaks
```
the code breaks where it should not.
If we make that arg public, we're telling the users that this would be ok-ish to construct the TD like this, which isn't.

tcbegley: Since we initialise an empty tensordict in the constructor, and then populate it via `__setattr__` from inside the dataclass' constructor, I think we will also need to set `_run_checks=False` inside there to really reduce overhead. Is it ok to always have that turned off though? Presumably we still want checks when running something like

```python
tc.X = torch.rand(100, 10)
```

vmoens: > Since we initialise an empty tensordict in the constructor, and then populate it via `__setattr__` from inside the dataclass' constructor, I think we will also need to set `_run_checks=False` inside there to really reduce overhead. Is it ok to always have that turned off though? Presumably we still want checks when running something like
> 
> ```python
> tc.X = torch.rand(100, 10)
> ```

yep in that case we need to run it

apbard: > Exactly for that: we don't want users to call that. We may refactor this in the future to hide it a bit more. You should use that only if you build a tensordict out of another tensordict and the checks have been done, which will only occur in with dedicated methods. I'm open to suggestions to make it cleaner :)

couldn't we simply store a `self._checked` boolean attribute that gets set to true the first time, and get rid of that argument constructor?

vmoens: That could be an option, I should go through the tensordict.py file and check if that would work in all cases

PR #177, opened 2023-01-23
[BugFix] Typo in tensorclass.__ne__

tcbegley:
## Description

Noticed a small bug in the definition of tensorclass' `__ne__` method. I don't think it was actually reachable though...

If we fail the first if statement, we are guaranteed to pass the second (because `other` must be an instance of `self.__class__` and hence `is_tensorclass` will be `True`). So we'll always hit one of the first two return statements.

In some sense then, the bugfix is irrelevant. I think we could probably refactor this to

```python
def __eq__(self, other):
    if not isinstance(other, self.__class__):
        return False
    return self.tensordict == other.tensordict
```

or the even more compact

```python
def __eq__(self, other):
    return isinstance(other, self.__class__) and self.tensordict == other.tensordict
```

We also could drop `__ne__` completely because default behaviour of `object.__ne__` is to return `not object.__eq__(self, other)`.

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -337,17 +337,13 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
 def __eq__(self, other):
     if not isinstance(other, self.__class__):
         return False
-    if is_tensorclass(other):
-        return self.tensordict.__eq__(other.tensordict)
-    return self == other
+    return self.tensordict == other.tensordict
 
 
 def __ne__(self, other):
     if not isinstance(other, self.__class__):
         return True
-    if is_tensorclass(other):
-        return self.tensordict != other.tensordict
-    return self != other
+    return self.tensordict != other.tensordict
 
 
 def _unbind(tdc, dim):
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -347,7 +347,7 @@ def __ne__(self, other):
         return True
     if is_tensorclass(other):
         return self.tensordict != other.tensordict
-    return self == other
+    return self != other
 
 
 def _unbind(tdc, dim):


tensordict/prototype/tensorclass.py
```diff
@@ -347,7 +347,7 @@ def __ne__(self, other):
         return True
     if is_tensorclass(other):
         return self.tensordict != other.tensordict
-    return self == other
+    return self != other
```
vmoens, line 350: oops 😅 🙊

vmoens (approved):
LGTM

vmoens:
> We also could drop `__ne__` completely because default behaviour of `object.__ne__` is to return `not object.__eq__(self, other)`.

I agree, although we may fall back on `TensorDict.__ne__` which we don't want no?

tcbegley:
> although we may fall back on TensorDict.__ne__ which we don't want no?

I'm not sure this can happen since the tensorclass is not a subclass of a tensordict, it just contains one?

vmoens:
> I'm not sure this can happen since the tensorclass is not a subclass of a tensordict, it just contains one?

Not sure either, we have a fallback mechanism which checks if something "exists" for your tensorclass. If it doesn't it try to get it from TensorDict. In this case I'm not sure which fallback would prevail (the `__ne__` = not `__eq__` or `TensorDict.__ne__`)

tcbegley:
Ah ok, I see what you mean.

I think for the avoidance of doubt, and in the interest of being explicit, let's keep it then. 

Though fwiw I'm still fairly confident we wouldn't fall back on it, both because of [this line](https://github.com/pytorch-labs/tensordict/blob/main/tensordict/prototype/tensorclass.py#L224) and because `__getattr__` shouldn't be invoked because `__ne__` will always be defined courtesy of `object`.

vmoens:
Cool
Do you want to adapt `__eq__` too or shall we keep this PR as it is?

tcbegley:
I've adapted both. Just double check my reasoning that that code was truly unreachable, but I think it's ok.

PR #185, opened 2023-01-24
[WIP, POC] Variable size representation of tensors

vmoens:
## Description

Proposes a variable sized tensor to be used in distributed / mp settings. Writing and accessing is currently slow, a "good" `__setitems__` and `__getitems__` could be designed entirely in python I believe, but this optimization work is secondary to the design decisions.

Eventually we'd like to use nested tensors with disk storage, but I don't see how to initialize a nested tensor without knowing in advance the shape of each tensor and allocate the storage in RAM of cuda, which is exactly what we don't want to do.

To test:

```python
from tensordict.dynamic_shaped_tensor import DynamicShapeTensor
import torch

if __name__ == "__main__":
    N = 10
    data = []
    d = DynamicShapeTensor([3, 4, -1], 3 * 4 * 12 * N, N, 0)
    for i in range(N):
        data.append(torch.randn(3, 4, torch.randint(1, 12, (1,)).item()))
        d[i] = data[-1]
    for i in range(N):
        assert (d[i] == data[i]).all()

```

cc @tcbegley FYI
cc @ @FrancescoSaverioZuppichini @hristo-vrigazov: any advice?

### DIFF
--- a/tensordict/dynamic_shaped_tensor.py
+++ b/tensordict/dynamic_shaped_tensor.py
@@ -10,7 +10,7 @@ class DynamicShapeTensor:
         shape: torch.Size,  # -1 indicates variable size
         total_elts: int,  # size of the storage
         total_index: int,  # number of indexable items along stack dim
-        stack_dim: 0,
+        stack_dim: int = 0,
     ):
         if stack_dim != 0:
             raise NotImplementedError(stack_dim)
--- a/tensordict/dynamic_shaped_tensor.py
+++ b/tensordict/dynamic_shaped_tensor.py
@@ -1,0 +1,74 @@
+import numpy as np
+import torch
+
+from tensordict import MemmapTensor
+
+
+class DynamicShapeTensor:
+    def __init__(
+        self,
+        shape: torch.Size,  # -1 indicates variable size
+        total_elts: int,  # size of the storage
+        total_index: int,  # number of indexable items along stack dim
+        stack_dim: 0,
+    ):
+        if stack_dim != 0:
+            raise NotImplementedError(stack_dim)
+        self.offsets = MemmapTensor(total_index, dtype=torch.int64).as_tensor()
+        self.total_elts = total_elts
+        assert sum(s == -1 for s in shape) == 1
+        assert sum(s > 0 for s in shape) == len(shape) - 1
+        self.shape = shape
+        self.shapes = MemmapTensor(total_index, dtype=torch.int64).as_tensor()
+        self.storage = MemmapTensor(total_elts).as_tensor()
+        self.variable_shape = sum(np.cumprod([s > 0 for s in shape]))
+        self.total_index = total_index
+
+    def __setitem__(self, idx: int, value):
+        assert isinstance(idx, int)
+        assert value.ndim == len(self.shape)
+        assert (s1 == s2 for (s1, s2) in zip(value.shape, self.shape) if s2 > 0)
+        prev_offset = self.offsets[idx]
+        offset = value.numel()
+        if idx < 0:
+            idx = self.total_index + idx
+        if idx > 0:
+            start = self.offsets[idx - 1].item()
+            stop = self.offsets[idx] = start + offset
+        else:
+            start = 0
+            stop = self.offsets[idx] = offset
+        if idx < self.total_index - 1 and self.offsets[idx + 1].item() > 0:
+            self.offsets[idx:] += offset - prev_offset
+            if self.offsets[-1] > self.total_elts:
+                raise RuntimeError
+        print(start, stop)
+        self.storage[start:stop] = value.view(-1)
+        shape = value.shape
+        out_shape = list(self.shape)
+        out_shape[self.variable_shape] = shape[self.variable_shape]
+        print(shape, out_shape)
+        self.shapes[idx] = shape[self.variable_shape]
+
+    def __getitem__(self, item):
+        if item < 0:
+            item = self.total_index + item
+        start = self.offsets[item - 1].item() if item > 0 else 0
+        stop = self.offsets[item].item()
+        if stop == 0:
+            raise RuntimeError
+        shape = list(self.shape)
+        shape[self.variable_shape] = self.shapes[item].item()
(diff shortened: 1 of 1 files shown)


tensordict/dynamic_shaped_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+from tensordict import MemmapTensor
+
+
+class DynamicShapeTensor:
+    def __init__(
+        self,
+        shape: torch.Size,  # -1 indicates variable size
+        total_elts: int,  # size of the storage
+        total_index: int,  # number of indexable items along stack dim
+        stack_dim: 0,
```
tcbegley, line 13:
```suggestion
        stack_dim: int = 0,
```

tensordict/dynamic_shaped_tensor.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
```
tcbegley, line 1: Should we move this to `tensordict.prototype`?

tensordict/dynamic_shaped_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+import torch
+
+from tensordict import MemmapTensor
+
+
+class DynamicShapeTensor:
```
tcbegley, line 7: This is slightly at odds with the name of the submodule `dynamic_shaped_tensor` vs. `DynamicShapeTensor`.

tcbegley (commented):
This looks cool!

Currently it will overwrite data and break the offsets if you write to an index with a different shape. E.g.

```python
import torch
from tensordict.dynamic_shaped_tensor import DynamicShapeTensor

t = DynamicShapeTensor([2, -1], 2 * 5 * 2, 2, 0)
t[0] = torch.rand(2, 3)
t[1] = torch.rand(2, 1)
t[0] = torch.rand(2, 4)  # writes over t[1]
t[1]  # RuntimeError as `start` and `stop` offsets are both 8.
```

Not sure how you could fix that without doing a lot of copies whenever a different shape is used. 

Possibly you could just enforce that the shape is fixed for each index once it has been used, but I don't know if that's going to be overly restrictive if e.g. you have random batches of data that you want to store in one of these things.

tensordict/dynamic_shaped_tensor.py
vmoens: ah right! let's go for shaped no?
Any other idea for the name by the way?

vmoens:
> This looks cool!
> 
> Currently it will overwrite data and break the offsets if you write to an index with a different shape. E.g.
> 
> ```python
> import torch
> from tensordict.dynamic_shaped_tensor import DynamicShapeTensor
> 
> t = DynamicShapeTensor([2, -1], 2 * 5 * 2, 2, 0)
> t[0] = torch.rand(2, 3)
> t[1] = torch.rand(2, 1)
> t[0] = torch.rand(2, 4)  # writes over t[1]
> t[1]  # RuntimeError as `start` and `stop` offsets are both 8.
> ```
> 
> Not sure how you could fix that without doing a lot of copies whenever a different shape is used.
> 
> Possibly you could just enforce that the shape is fixed for each index once it has been used, but I don't know if that's going to be overly restrictive if e.g. you have random batches of data that you want to store in one of these things.

oh gosh in that case we would need to shift the entire dataset, that's right. Maybe we should constrain re-writing to keeping the same shape?

hristo-vrigazov:
> oh gosh in that case we would need to shift the entire dataset, that's right. Maybe we should constrain re-writing to keeping the same shape?

I think for sure this has to be restrained, otherwise the user of the library should re-initialize the tensor.

By the way, are we sure about the name? Because it looks like it allows dynamic values only along a given axis/dimension, not a fully dynamic shape. For example, if the user of  the library wants to store images (which have two axes of varying values), I don't think this class would be suited for this purpose. Maybe `DynamicDimTensor` as an alternative would be slightly closer to what the class is doing?

tensordict/dynamic_shaped_tensor.py
tcbegley: That's fair. I guess we need to hold ourselves to moving stuff out once we're relatively confident the API is stable and there's no major bugs. I think there is some value in having the prototype submodule so that we can merge and try out new features while indicating to the user that they might be brittle.

Maybe tensorclass' time has come!

tensordict/dynamic_shaped_tensor.py
tcbegley: I'm happy with "shaped".

Only other thing that sprung to mind was something like "jagged" or "ragged", but I think you just run the risk of confusion with `KeyedJaggedTensor` etc.?

tensordict/dynamic_shaped_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+        self.variable_shape = sum(np.cumprod([s > 0 for s in shape]))
+        self.total_index = total_index
+
+    def __setitem__(self, idx: int, value):
+        assert isinstance(idx, int)
+        assert value.ndim == len(self.shape)
+        assert (s1 == s2 for (s1, s2) in zip(value.shape, self.shape) if s2 > 0)
+        prev_offset = self.offsets[idx]
+        offset = value.numel()
+        if idx < 0:
```
hristo-vrigazov, line 33: I think you can re-use Numpy's indexing behavior without reimplementing it here. This can be achieved by initializing a field with `np.arange` with the number of elements, and apply `idx` to this field, this will give you the indices you need, always positive. This will also make sure that slicing also works

vmoens: I wonder what's the overhead of doing that? What if you have >100M elements for instance?

hristo-vrigazov: There will be memory overhead, yes. Similar to the offsets array. And since here we don't need to support advanced indexing in getitem, then it does not make sense, it's just unnecessary overhead. Only if in the future we want to for example to be able to pass a multi-dimensional tensor to getitem (in the way it can be passed in numpy), it would make sense

PR #187, opened 2023-01-25
[Refactor] Generate test_tensorclass.py from test_tensorclass_nofuture.py

tcbegley:
## Description

Currently we maintain two sets of almost identical tests in `test_tensorclass.py` and `test_tensorclass_nofuture.py` to test the impact of the directive `from __future__ import annotations`.

This PR proposes using `conftest.py` to generate `test_tensorclass.py` from `test_tensorclass_nofuture.py` in order to ensure that the two files stay in sync.

Feedback very welcome!


vmoens (approved):
LGTM
I like it that we only keep the one marked "nofuture"

tumblr_n8qx4tNF6f1tv3v8ao1_500

PR #188, opened 2023-01-25
[BugFix] Relax error matching in tensorclass tests

tcbegley:
## Description

Relax error matching in TensorClass tests. Largely motivated by FBCode compatibility, but since #165 the need to check the representation of types in error messages is less important anyway.


vmoens (approved):
LGTM

Issue #178, opened 2023-01-23
[Feature Request] More arguments for `td.memmap_()`


vmoens:
## Motivation

Since `TensorDict.memmap_()` creates a copy on physical mem of the content of a tensordict, it would be nice to have some more options:
- [ ] `MemmapTensors` are temporary files. It could be nice to disable this when calling `memmap_()` (like `temporary=False`). In that way, one would know that the content of the memmap will be kept in memory after the process exits.
- [ ] We should also be able to pass the path to where the files have to be saved (`path="/path/to/my_tensordict"). This would work nicely with #176 
- [ ] Similarly, it could be neat to be able to control if a `memmap_` tensor can be written or not (by passing `mode="r+"` for instance). Of course, `memmap_()` would always write the content but then modify the permissions accordingly to make sure that the content is somewhat `locked` for in-place modifs.

vmoens:
cc @sreevasthav

PR #189, opened 2023-01-25
[Feature] More arguments for MemmapTensor construction

tcbegley:
## Description

This PR addresses #178 by adding more arguments to the `MemmapTensor` constructor, and making the behaviour of `TensorDict.memmap_` richer when `prefix` is specified.

In brief:
- `MemmapTensor` now accepts two additional arguments: `filename` and `mode`.
- `filename` determines the location on disk of the underlying `memmap`.
- `mode` determines the opening mode of the underlying `memmap`. Options are `"r+"`, `"w+"` and `"c"` as well as their verbose equivalents. This is what `numpy.memmap` supports with the exception of `"r"` as `torch.Tensor` gets quite unhappy if the underlying NumPy array is not writable.
- `TensorDict.memmap_` with a specified `prefix` will now locate the contents of the tensordict in the specified directory, with a nested structure determined by the keys in the tensordict. E.g.

    ```python
    import torch
    td = TensorDict({"a": torch.rand(10), "b": {"c": torch.rand(10)}}, [10])
    td.memmap_(prefix="tensordict")
    ```

    yields the following structure

    ```
    tensordict
    ├── a.memmap
    ├── a.meta.pt
    ├── b
    │   ├── c.memmap
    │   ├── c.meta.pt
    │   └── meta.pt
    └── meta.pt
    ```

- Other variants of tensordict work similarly (only `SubTensorDict` is not supported as they already don't support `.memmap_`.
- I've added a classmethod `load_memmap` to `TensorDict` and variants which can be used to read this directory structure back into a new tensordict.
    ```python
    import torch
    td = TensorDict({"a": torch.rand(10), "b": {"c": torch.rand(10)}}, [10])
    td.memmap_(prefix="tensordict")
    td2 = TensorDict.load_memmap("tensordict")
    equal = td == td2
    assert all(v.all() for v in equal.values(include_nested=True))    
    ```

Note that we ended up not adding the proposed `temporary` kwarg to `TensorDict.memmap_` as it was felt that supplying `prefix` was an implicit request that the saved file should not be ephemeral.

### DIFF
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -10,6 +10,7 @@ import os
 import tempfile
 import warnings
 from copy import copy, deepcopy
+from pathlib import Path
 from tempfile import _TemporaryFileWrapper
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -98,7 +99,10 @@ class MemmapTensor:
             serialization. If False, the current process keeps the ownership
             of the temporary file.
             Default: False.
-        prefix (str or path, optional): prefix of the file location.
+        prefix (str or path, optional): *Deprecated* prefix of the file location. Should
+            not be specified together with prefix.
+        filename (str or path, optional): location of the underlying memory-map. Should
+            not be specified together with prefix.
 
     Examples:
         >>> x = torch.ones(3,4)
@@ -135,12 +139,42 @@ class MemmapTensor:
         dtype: torch.dtype = None,
         transfer_ownership: bool = False,
         prefix: Optional[str] = None,
+        filename: Optional[str] = None,
+        mode: str = "r+",
     ):
         self.idx = None
         self._memmap_array = None
         self.prefix = prefix
         self.is_meta = False
-        self.file = tempfile.NamedTemporaryFile(prefix=prefix, delete=False)
+
+        if mode in ("r+", "w+", "c", "copyonwrite", "readwrite", "write"):
+            self.mode = mode
+        else:
+            raise ValueError(
+                'Accepted values for mode are "r+", "readwrite", "w+", "write", "c" or '
+                '"copyonwrite". PyTorch does not support tensors backed by read-only '
+                'NumPy arrays, so "r" and "readonly" are not supported.'
+            )
+
+        if prefix is not None:
+            warnings.warn(
+                "prefix has been deprecated. If you want to control the location of "
+                "the MemmapTensor on disk, consider using filename instead."
+            )
+            if filename is not None:
+                raise ValueError("filename and prefix should not both be specified")
+
+        # open the files in r+ mode so as to not overwrite any data that might exist
+        # there. the actual memmap will be instantiated with user-supplied mode
+        if filename is None:
+            self.file = tempfile.NamedTemporaryFile(
+                prefix=prefix, delete=False, mode="r+"
+            )
+        else:
+            # if filename doesn't exist we must create it
+            Path(filename).touch(exist_ok=True)
+            self.file = open(filename, mode="r+")
+
         self.filename = self.file.name
         self.file.close()  # we close the file for now, but don't delete it
 
@@ -177,6 +211,8 @@ class MemmapTensor:
         tensor: Union[torch.Tensor, MemmapTensor, np.ndarray],
         transfer_ownership=False,
         prefix=None,
+        filename: Optional[str] = None,
+        mode: str = "r+",
     ) -> MemmapTensor:
         if isinstance(tensor, MemmapTensor):
             if transfer_ownership:
@@ -206,6 +242,8 @@ class MemmapTensor:
             dtype=dtype,
             prefix=prefix,
             transfer_ownership=transfer_ownership,
+            filename=filename,
+            mode=mode,
         )
         out.copy_(tensor)
         return out
@@ -242,7 +280,6 @@ class MemmapTensor:
         self._dtype = dtype
         self._ndim = len(shape)
         self._numel = prod(shape)
-        self.mode = "r+"
         self._has_ownership = True
         self._had_ownership = True
 
@@ -479,7 +516,10 @@ MemmapTensor of shape {self.shape}."""
 
     def __del__(self) -> None:
         if "_has_ownership" in self.__dir__() and self._has_ownership:
-            os.unlink(self.filename)
+            if isinstance(self.file, tempfile._TemporaryFileWrapper):
+                # only delete file if we created a temporary file. Otherwise file should
+                # persist on disk
+                os.unlink(self.filename)
             del self.file
 
     def __eq__(self, other: Any) -> torch.Tensor:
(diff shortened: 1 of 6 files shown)


tcbegley:
Not quite sure why the tests are failing. I thought maybe a device issue, but I ran on cluster with GPUs available and they all passed. Will investigate more tomorrow.

vmoens:
This is not bc-breaking, right?

tensordict/tensordict.py
```diff
@@ -2656,23 +2662,75 @@ def memmap_(self, prefix=None) -> TensorDictBase:
...
+                    self._tensordict[key] = value.memmap_(prefix=prefix / key)
+                    torch.save(
+                        {"batch_size": value.batch_size, "device": value.device},
+                        prefix / key / "meta.pt",
+                    )
+                else:
+                    self._tensordict[key] = value.memmap_()
                 continue
-            self._tensordict[key] = MemmapTensor(value, prefix=prefix)
+            self._tensordict[key] = MemmapTensor(
```
vmoens, line 2676: what happens if value is already a MemmapTensor?
Do we create a new one with a copied content?

I'm thinking: maybe if it is a MemmapTensor placed at a different location, we could raise an exception here, telling the user that this would otherwise have a weird behaviour

```
td = TensorDict({"one": MemmapTensor(..., filename="two.memmap")}, [])
td.memmap_("./")
```
ie would we expect that  `mv two.memmap one.memmap` or that `cp two.memmap one.memmap`? Or just nothing?

tensordict/tensordict.py
```diff
@@ -2656,23 +2662,75 @@ def memmap_(self, prefix=None) -> TensorDictBase:
...
+                value,
+                filename=str(prefix / f"{key}.memmap") if prefix is not None else None,
+            )
+            if prefix is not None:
+                torch.save(
```
vmoens, line 2681: OOC could we save all the metadata at once?
load/save overhead is not too much a worry here though (I wouldn't suspect it to constitute a bottleneck in a typical training loop)

vmoens (approved):
LGTM
Can you comment on `memmap_()` when the content is already memmap?

tensordict/tensordict.py
tcbegley: Yes, the current behaviour if the element is already a `MemmapTensor` is to create a new `MemmapTensor` with copied content. Relevant logic is here:

https://github.com/pytorch-labs/tensordict/blob/1ae3066d104ed615836c8a088d0b02f93d3ee29f/tensordict/memmap.py#L226-L228

I think an error is maybe a bit restrictive, though I do think it could be good to allow user to choose between copy and move. Maybe smth like `copy_existing=False` would either move or raise an error if the current memmap is in the wrong place.

tensordict/tensordict.py
tcbegley: I think in principle there's no particular reason we couldn't, it would just be a little bit more fiddly most likely. But agreed, there is some overhead from doing multiple reads / writes each time.

One thing though that is nice about the current approach though is that you can load a nested tensordict by pointing `TensorDict.load_memmap` at the nested directory. It can be oblivious to the parent directories.

tensordict/tensordict.py
vmoens: > Maybe smth like copy_existing=False would either move or raise an error if the current memmap is in the wrong place.

copy_existing=False (default): raises an exception if already memmap and not in the right place
copy_existing=True: creates a copy at desired loc

What if it's in `/tmp`? We should just use the current location and not raise an error neither move the data right?

tensordict/tensordict.py
vmoens: Yeah I agree let's leave it as is. 
Can we add to the doc that this is "supposed" to be slow, and should not be used to move things around during training?

tensordict/tensordict.py
tcbegley: This sounds good.

> What if it's in `/tmp`?

As in, what if the user calls `td.memmap_` without specifying `prefix`? I think in that case indeed, we don't care about where the files are saved, so if they already exist we can just leave them where they are. You do run the risk of the user deleting the file from elsewhere in their code, but that's technically a risk even if you copy.

vmoens: Cool! So `memmap_()` with memmap files created without specific path is a no-op (just locks the tensordict)

vmoens:
Oh sorry Tom
Can we also update the doc with this? In the section about saving tensordicts.
Thanks!

tcbegley:
> This is not bc-breaking, right?

Only just noticed this question, apologies. I believe everything here should be fully backwards compatible, though note I have declared that `prefix` will be deprecated from `MemmapTensor` which of course would be bc-breaking.

If you call `memmap_` on a `TensorDict` with a specified `prefix`, those memmaps will be created in a different location to before, so in that sense this is not strictly bc, however previously all memmaps would be thrown in the specified directory with randomly generated filename, so the only way to extract things again would be to iterate over the `MemmapTensors` and access `.filename` then do something with it. That should continue to work, even though those memmaps are no longer all in the same directory.

tensordict/tensordict.py
tcbegley: I've added the `copy_existing` argument as discussed.

tensordict/tensordict.py
tcbegley: Added a note on slowness to docs and docstring

tcbegley:
I've added the discussed changes and documentation to the "saving" page. I think this is all good to go now.

tensordict/memmap.py
```diff
@@ -177,6 +211,8 @@ def from_tensor(
         tensor: Union[torch.Tensor, MemmapTensor, np.ndarray],
         transfer_ownership=False,
         prefix=None,
+        filename: Optional[str] = None,
+        mode: str = "r+",
     ) -> MemmapTensor:
         if isinstance(tensor, MemmapTensor):
```
vmoens, line 217: what if I pass a memmap tensor, but i say that the filename is something different?
I would guess that the new memmap tensor is at the desired location.
Should we (1) raise an exception, (2) move the memmap tensor or (3) create a copy?

vmoens (approved):
LGTM
Just a question about from_tensor

tensordict/memmap.py
tcbegley: At the moment it creates a copy at the new location.

We could certainly add a keyword argument to allow the file to be moved instead of copied though? Or an exception to be more consistent with `TensorDict.memmap_`?

vmoens: I'm not sure it does. I think it would if we'd be doing
```suggestion
        if isinstance(tensor, MemmapTensor) and filename is None:
            ...
```
but maybe i'm missing smth

vmoens (approved):

tensordict/memmap.py
tcbegley: Hmm, I think you're right actually. Let me fix that and understand why my test didn't catch it.

tcbegley: Thanks for catching! Should now be fixed and I've made sure the tests catch this case.

vmoens (approved):

PR #191, opened 2023-01-26
[BugFix] Using set_interaction_mode as decorator

vmoens:
## Description

Using `set_interaction_mode` as a decorator did not have the desired behaviour bc the `clone` method had to be adapted.


tcbegley (approved):
LGTM!

PR #192, opened 2023-01-26
[BugFix] MemmapTensor from MemmapTensor __dir__ issue

tcbegley:
## Description

This PR fixes a bug that arises when you construct one `MemmapTensor` from another which I noticed while working on #189 

Repro:

```python
>>> import torch
>>> from tensordict import MemmapTensor
>>> mt = MemmapTensor(4, 3, 2, 1)
>>> assert mt.squeeze(-1).shape == torch.Size([4, 3, 2])  # this is fine
>>> mt2 = MemmapTensor(mt)
>>> mt2.squeeze(-1)  # AttributeError: squeeze not found
```

Issue is that internal attribute `self._tensor_dir` was constructed from `elem.__dir__()`, but this leads to attributes not being found.

### DIFF
--- a/test/test_memmap.py
+++ b/test/test_memmap.py
@@ -49,15 +49,7 @@ def test_grad():
         torch.bool,
     ],
 )
-@pytest.mark.parametrize(
-    "shape",
-    [
-        [
-            2,
-        ],
-        [1, 2],
-    ],
-)
+@pytest.mark.parametrize("shape", [[2], [1, 2]])
 def test_memmap_data_type(dtype, shape):
     """Test that MemmapTensor can be created with a given data type and shape."""
     t = torch.tensor([1, 0], dtype=dtype).reshape(shape)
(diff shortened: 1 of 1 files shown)


test/test_memmap.py
```diff
@@ -49,15 +49,7 @@ def test_grad():
...
-@pytest.mark.parametrize(
-    "shape",
-    [
-        [
-            2,
-        ],
-        [1, 2],
-    ],
-)
+@pytest.mark.parametrize("shape", [[2], [1, 2]])
```
tcbegley, line 52: No changes here, just reformatting some [magic trailing commas](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma)

vmoens (commented):
Good catch!
I tried this and it breaks:
```python
import torch
from tensordict import MemmapTensor
mt = MemmapTensor(4, 3, 2, 1)
mt2 = mt[0]
assert isinstance(mt2, MemmapTensor)
assert mt2.squeeze(-1).shape == torch.Size([3, 2])  # this is fine
mt3 = MemmapTensor(mt2)
mt3.squeeze(-1)  # AttributeError: squeeze not found
```

Worse, this breaks for a whole different reason:
```python
from tensordict import MemmapTensor
mt = MemmapTensor(4, 3, 2, 1)
mt = mt[0]
mt + 1
```
because we use the same name so the original `mt` disappears and clears the temporary file!
Not sure how to fix this

tcbegley:
> I tried this and it breaks:

Your first example runs fine for me on this branch. At least, when I copy paste the snippet I don't see an `AttributeError`.

I can reproduce the second example though, that does indeed seem nasty. Let me have a think about that...

tcbegley:
This PR is kind of obsolete with the changes from #193.

We can either merge to add the new test or abandon.

PR #193, opened 2023-01-26
[Feature] apply across multiple TDs

### LINKED ISSUE #181: [Feature Request] Operations between TensorDicts
## Motivation

It would be really useful if there would be the possibility to perform operations such as _sum, multiplication_ etc... between two TensorDicts ( key-wise). Is this a feature that you have in mind to introduce in future?

Right now I noticed that is not possible.

## Checklist

- [ X] I have checked that there is no similar issue in the repo (**required**)

vmoens:
## Description

Allows to call `apply` across multiple TDs.

EDIT:

As pointed in the conversation below, this highlighted a bug with `MemmapTensor` created from other `MemmapTensor`s. The whole thing was becoming messy and I changed the way `MemmapTensor` are built as a result of this.
This IS BC-breaking:
```python
MemmapTensor(tensor)  # not allowed anymore
MemmapTensor.from_tensor(tensor)  # proper way to do it
MemmapTensor.from_tensor(memmap_tensor)  # returns memmap_tensor (without copy)
```

We would need to see what to do with #189: I would guess that `from_tensor` could take other arguments, such as the destination etc. If any of these suggests that a copy must be done, then it will be.

Wdyt @tcbegley?

Closes #181 

cc @Daxo32

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -2699,7 +2699,7 @@ class TensorDict(TensorDictBase):
             if isinstance(value, TensorDictBase):
                 self._tensordict[key] = value.memmap_()
                 continue
-            self._tensordict[key] = MemmapTensor(value, prefix=prefix)
+            self._tensordict[key] = MemmapTensor.from_tensor(value, prefix=prefix)
         self._is_memmap = True
         self.lock()
         return self
@@ -3455,7 +3455,7 @@ torch.Size([3, 2])
             if self.is_shared() and self.device.type == "cpu":
                 tensor_expand.share_memory_()
             elif self.is_memmap():
-                tensor_expand = MemmapTensor(tensor_expand)
+                tensor_expand = MemmapTensor.from_tensor(tensor_expand)
         parent.set(key, tensor_expand, _run_checks=_run_checks)
         self.set_(key, tensor)
         return self
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -663,7 +663,7 @@ class TensorDictBase(Mapping, metaclass=abc.ABCMeta):
         kwargs = {}
         if not isinstance(self, SubTensorDict):
             kwargs["_process"] = False
-        is_locked = self.is_locked
+        is_locked = out.is_locked
         if not inplace and is_locked:
             out.unlock()
         for key, item in self.items():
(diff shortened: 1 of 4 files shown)


tensordict/tensordict.py
```diff
@@ -1162,15 +1199,19 @@ def unbind(self, dim: int) -> Tuple[TensorDictBase, ...]:
...
+                    lambda tensor, idx=_idx: tensor[idx],
+                    batch_size=batch_size,
+                )
             )
-            for _idx in idx
-        )
+            if self.is_shared():
+                out[-1].share_memory_()
+            elif self.is_memmap():
+                out[-1].memmap_()
```
vmoens, line 1213: @tcbegley 
Here we index the memmap tensor, hence we get a memmap tensor.
But the tensordict is not marked as memmap.
so we call memmap_() if needed, to "mark" it as memmap. But we don't expect anything to happen to the tensors (they should stay where they are). What would happen with #189?

tcbegley: The way I have things set up at the moment, if the entries of the tensordict are already `MemmapTensors`, _and_ either the user doesn't specify a `prefix` in `memmap_`, or the memmaps are already located in the right place, then `memmap_` is a no-op on the entries, it simply sets `self._is_memmap = True` and locks the tensordict.

So if I'm understanding you right, there shouldn't be any conflict here I think.

tcbegley (dismissed):
LGTM!

tcbegley:
Tests appear to be failing because of this line

https://github.com/pytorch-labs/tensordict/blob/c50a1a6f18413378d8f66efbde8a5add0b1b361a/test/test_tensordict.py#L927

There's something really weird going that I don't understand. Look at this example:

```python
>>> import torch
>>> td = TensorDict({"a": torch.arange(12).reshape(3, 4)}, [3]).memmap_()
>>> td_unbind = torch.unbind(td, dim=0)
>>> td_unbind[0]["a"].contiguous()
tensor([0, 1, 2, 3])
>>> td_unbind[1]["a"].contiguous()  # row 1 contains values from row 0...
tensor([0, 1, 2, 3])
>>> td[0]["a"].contiguous()
tensor([0, 1, 2, 3])
>>> td[1]["a"].contiguous()
tensor([4, 5, 6, 7])
```

This doesn't happen if you don't call `memmap_` on `td` initially, so it's not the apply statement, must be something to do with the line

```python
elif self.is_memmap():
    out[-1].memmap_()
```

But I can't figure out why this goes wrong...

vmoens:
that ain't good let me see

vmoens:
> Tests appear to be failing because of this line
> 
> https://github.com/pytorch-labs/tensordict/blob/c50a1a6f18413378d8f66efbde8a5add0b1b361a/test/test_tensordict.py#L927
> 
> There's something really weird going that I don't understand. Look at this example:
> 
> ```python
> >>> import torch
> >>> from tensordict import TensorDict
> >>> td = TensorDict({"a": torch.arange(12).reshape(3, 4)}, [3]).memmap_()
> >>> td_unbind = torch.unbind(td, dim=0)
> >>> td_unbind[0]["a"].contiguous()
> tensor([0, 1, 2, 3])
> >>> td_unbind[1]["a"].contiguous()  # row 1 contains values from row 0...
> tensor([0, 1, 2, 3])
> >>> td[0]["a"].contiguous()
> tensor([0, 1, 2, 3])
> >>> td[1]["a"].contiguous()
> tensor([4, 5, 6, 7])
> ```
> 
> This doesn't happen if you don't call `memmap_` on `td` initially, so it's not the apply statement, must be something to do with the line
> 
> ```python
> elif self.is_memmap():
>     out[-1].memmap_()
> ```
> 
> But I can't figure out why this goes wrong...

This was a problem with the way we were creating new `MemmapTensor`s from others.
A `MemmapTensor` contains an `_index` that represents what elements must be returned from the original array.
When copying the old onto the new memmap, we were not copying the index.

tcbegley (approved):
Nice! This looks a lot cleaner.

Issue #194, opened 2023-01-27
[BUG] Improvements on MemmapTensor


vmoens:
## Describe the bug

We need to do some woodwork on MemmapTensor:
- [ ] Remove the ownership and replace it by a check over the number of times the array is being opened (for temporary files only)
- [ ] Solve a bug where calling `MemmapTensor(memmap_tensor[idx])` returns the wrong view of the `MemmapTensor` #193 
- [ ] Improve `MemmapTensor` creation args, including: creating a `Memmap` on a specific location (#189)
- [ ] Improve the way we copy a `Memmap` from one place to another, if we want and if we don't want to change the filename

cc @tcbegley

tcbegley:
- [ ] `AttributeError` on `Tensor` attributes when creating one `MemmapTensor` from another. (#192 <- needs to be adapted to recently merged #193)

I'll also update #189, though the desired no-op behaviour there is blocked by #192.

PR #197, opened 2023-01-30
[Benchmarks] Add MemmapTensor benchmarks

tcbegley:
## Description

Adding some basic `MemmapTensor` benchmarks to assist work on #194

### DIFF
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -1,7 +1,2 @@
-import sys
-
-sys.path.append("../test")
-
-
 def pytest_addoption(parser):
     parser.addoption("--rank", action="store")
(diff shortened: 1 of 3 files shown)


benchmarks/conftest.py
```diff
@@ -1,2 +1,7 @@
+import sys
+
+sys.path.append("../test")
```
tcbegley, line 3: I added this to let me import `get_available_devices` from `_utils_internal`.

We could alternatively accept a small amount of duplication and just copy / paste it?

vmoens: We can copy yes, no worry
I'd rather have that than messing with relative imports (fbcode will complain)

vmoens (approved):
LGTM 
you can copy the utils

PR #199, opened 2023-02-03
Added Test for Iteration behavior in TensorDictKeysView

ruleva1983:
## Description

Added a test case to check new changes in __iter__ function for TensorDictKeysView

## Motivation and Context

This test was to generated to clarify acceptable behavior in iteration of KeysView

## Types of changes

- [x] Unittest

## Checklist

### DIFF
--- a/tensordict/__init__.py
+++ b/tensordict/__init__.py
@@ -9,6 +9,7 @@ from .tensordict import (
     merge_tensordicts,
     SubTensorDict,
     TensorDict,
+    detect_loop,
 )
 
 try:
--- a/tensordict/__init__.py
+++ b/tensordict/__init__.py
@@ -10,6 +10,7 @@ from .tensordict import (
     SubTensorDict,
     TensorDict,
     detect_loop,
+    _TensorDictKeysView,
 )
 
 try:
@@ -24,4 +25,5 @@ __all__ = [
     "TensorDict",
     "merge_tensordicts",
     "set_transfer_ownership",
+    "_TensorDictKeysView"
 ]
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -5470,6 +5470,30 @@ def _clone_value(value, recurse):
 
 
 def detect_loop(tensordict: TensorDict) -> bool:
+    """
+        This helper function detects the presence of an auto nesting loop inside
+        a TensorDict object. Auto nesting appears when a key of TensorDict references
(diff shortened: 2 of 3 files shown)


tensordict/__init__.py
```diff
@@ -9,6 +9,8 @@
     merge_tensordicts,
     SubTensorDict,
     TensorDict,
+    detect_loop,
+    _TensorDictKeysView,
```
tcbegley, line 13: I think we don't want to expose this, if someone really wants to use it then they can import from `tensordict.tensordict`.

tensordict/tensordict.py
```diff
@@ -5400,3 +5467,22 @@ def _clone_value(value, recurse):
...
+
+def detect_loop(tensordict: TensorDict) -> bool:
+    visited = set()
+    visited.add(id(tensordict))
+
+    def detect(t_d: TensorDict):
+        for k, v in t_d.items():
+            if id(v) in visited:
+                return True
+            visited.add(id(v))
```
tcbegley, line 5480: Can move the `visited.add` and `visited.remove` inside the if statement, as it's only relevant in the tensordict case.

test/test_tensordict.py
```diff
@@ -3714,6 +3715,187 @@ def test_tensordict_prealloc_nested():
     assert buffer["agent.obs"].batch_size == torch.Size([B, N, T])
 
 
+def test_tensordict_view_iteration():
```
tcbegley, line 3718: Some of this can likely be unified with [`test_keys_view`](https://github.com/pytorch-labs/tensordict/blob/3f20f804dc36e58c8fc20f69ceab526f49d5df86/test/test_tensordict.py#L2963)

tcbegley (commented):
Thanks! A couple of inline comments. This also needs a quick rebase on `auto-nested-tensordicts` branch as the commit history has got a bit confused.

How do you imagine `detect_loop` being used?

tensordict/__init__.py
ruleva1983: Alright but we need to import it to the test file in order to run the tests

test/test_tensordict.py
ruleva1983: We should actually create a full class for all tests regarding TensorDictKeysView same way as for TensorDict tests

ruleva1983:
New PR created that takes this over. Closing

PR #200, opened 2023-02-06
Loop check and iter check for TensorDictKeysView

ruleva1983:
## Description

- Added helper function for detection of loops in autonested case
- Detailed test for the __iter__ function of TensorDictKeysView

## Motivation and Context

Test required to check the iteration functionality, specifically in the case of autonested that
each value is actually referenced only once

## Types of changes

- [x] Tests
- [ ] 
## Checklist


tcbegley (approved):
LGTM

PR #201, opened 2023-02-06
[WIP] Auto-nested TensorDict

tcbegley:
## Description

This PR adds support for auto-nesting inside `TensorDict`. Work-in-progress, will update this comment with details and examples in future.

### DIFF
--- a/tensordict/__init__.py
+++ b/tensordict/__init__.py
@@ -5,7 +5,6 @@
 
 from .memmap import MemmapTensor, set_transfer_ownership
 from .tensordict import (
-    _TensorDictKeysView,
     detect_loop,
     LazyStackedTensorDict,
     merge_tensordicts,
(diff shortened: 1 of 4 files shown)


ruleva1983:
test_chunk: fails for auto-nested due to usage of cat function. Tentatively disable it for that specific case

vmoens:
We're putting this PR on hold for now.
The changes are extensive and substantially reduce code readability.
The plan is either to finish this PR at some point in the future, or adopt another strategy (e.g. a specialized class for auto-nesting).

tcbegley:
For the benefit of anyone who picks this up in future, copying my comment from #220 about the reasons for some outstanding test failures:

The following tests fail because `assert_allclose_td` does not support auto-nested values
- `test_from_empty`
- `test_masking`
- `test_getitem_ellipsis`
- `test_getitem_range`

The following tests fail because we can't instantiate a TensorDict from a Python `dict` with auto-nested values
- `test_broadcast`
- `test_equal_dict`
- `test_nested_dict_init`

Finally `test_nestedtensor_stack` is failing because `LazyStackedTensorDict.contiguous` is broken. I think a few other methods for `LazyStackedTensorDict` could be broken but not caught by the tests. The issues here largely stem from the fact that values are computed lazily, and hence use of `id` to check for repeated values is brittle.

PR #205, opened 2023-02-08
[Test, BugFix] add regression test for PR #203

apbard:
#203 fixed a bug but didn't include a dedicated regression test

### DIFF
--- a/test/test_tensorclass_nofuture.py
+++ b/test/test_tensorclass_nofuture.py
@@ -368,14 +368,24 @@ def test_setitem():
 
     data1 = MyDataMemMap1(
         x=torch.zeros(3, 4, 5),
-        y=MemmapTensor.from_tensor(torch.zeros(3, 4, 5), transfer_ownership=True),
+        y=MemmapTensor.from_tensor(torch.zeros(3, 4, 5)),
         batch_size=[3, 4],
     )
-    data2 = MyDataMemMap2(
-        x=MemmapTensor.from_tensor(torch.ones(3, 4, 5), transfer_ownership=True),
+    data2_wrong = MyDataMemMap2(
+        x=MemmapTensor.from_tensor(torch.ones(3, 4, 5)),
         y=torch.ones(3, 4, 5),
         batch_size=[3, 4],
     )
+    with pytest.raises(ValueError, match="__setitem__ is only allowed for same-class assignement"):
+        data1[:2] = data2_wrong[:2]
+    with pytest.raises(ValueError, match="__setitem__ is only allowed for same-class assignement"):
+        data2_wrong[2:] = data1[2:]
+        
+    data2 = MyDataMemMap1(
+        x=MemmapTensor.from_tensor(torch.ones(3, 4, 5)),
+        y=torch.ones(3, 4, 5),
+        batch_size=[3, 4],
+    )    
     data1[:2] = data2[:2]
     assert (data1[:2] == 1).all()
     assert (data1.x[:2] == 1).all()
--- a/test/test_tensorclass_nofuture.py
+++ b/test/test_tensorclass_nofuture.py
@@ -376,16 +376,19 @@ def test_setitem():
         y=torch.ones(3, 4, 5),
         batch_size=[3, 4],
     )
-    with pytest.raises(ValueError, match="__setitem__ is only allowed for same-class assignement"):
+    with pytest.raises(ValueError, 
+                       match="__setitem__ is only allowed for same-class assignement"):
         data1[:2] = data2_wrong[:2]
-    with pytest.raises(ValueError, match="__setitem__ is only allowed for same-class assignement"):
+    with pytest.raises(ValueError, 
+                       match="__setitem__ is only allowed for same-class assignement"):
         data2_wrong[2:] = data1[2:]
         
     data2 = MyDataMemMap1(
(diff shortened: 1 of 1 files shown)


test/test_tensorclass_nofuture.py
```diff
@@ -354,6 +354,47 @@ class MyDataNested:
...
+        batch_size=[3, 4],
+    )
+    data2_wrong = MyDataMemMap2(
+        x=MemmapTensor.from_tensor(torch.ones(3, 4, 5)),
+        y=torch.ones(3, 4, 5),
+        batch_size=[3, 4],
+    )
+    with pytest.raises(ValueError, match="__setitem__ is only allowed for same-class assignement"):
+        data1[:2] = data2_wrong[:2]
+    with pytest.raises(ValueError, match="__setitem__ is only allowed for same-class assignement"):
```
apbard, line 381: The fact that this raises and the following case not is a bit strange. @vmoens wdyt?

test/test_tensorclass_nofuture.py
```diff
@@ -354,6 +354,47 @@ class MyDataNested:
...
+        data2_wrong[2:] = data1[2:]
+        
+    data2 = MyDataMemMap1(
+        x=MemmapTensor.from_tensor(torch.ones(3, 4, 5)),
+        y=torch.ones(3, 4, 5),
+        batch_size=[3, 4],
+    )    
+    data1[:2] = data2[:2]
+    assert (data1[:2] == 1).all()
```
apbard, line 390: this currently returns False. I think we should treat scalars differently and perform a broadcast @vmoens, @tcbegley, @sreevasthav wdyt?

vmoens: Yep
Happy to implement some broadcasting for `TensorDIct` first with float, integers, tensors and memmap-tensors in `__setitem__`.

test/test_tensorclass_nofuture.py
vmoens: I just tried and on my side both tests pass

test/test_tensorclass_nofuture.py
vmoens: This actually works as of today
```python
td = TensorDict>>> td = TensorDict({"a": [1, 1, 1]}, [])
>>> td==1
TensorDict(
    fields={
        a: Tensor(shape=torch.Size([3]), device=cpu, dtype=torch.bool, is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
>>> (td==1)["a"]
tensor([True, True, True])
```
but this does not 
```
>>> td==torch.zeros(3)
False
```
I can make a quick patch

vmoens: I also think it makes sense to compare two tensorclass of a different type, but I might get this wrong.
Our type checks work with `isinstance`, hence we'll get some weird behaviours if one class inherits from another on the lhs or rhs (ie `a==b` is not `b==a`)

test/test_tensorclass_nofuture.py
apbard: they pass *without* the pytest.raise?

test/test_tensorclass_nofuture.py
apbard: > I also think it makes sense to compare two tensorclass of a different type, but I might get this wrong. Our type checks work with `isinstance`, hence we'll get some weird behaviours if one class inherits from another on the lhs or rhs (ie `a==b` is not `b==a`)

we can relax the constraint to be just a tensorclass with all keys equal

test/test_tensorclass_nofuture.py
vmoens: I'm confused, the first should work with `pytest.raises` but not the second, which should not raise any exception. Is it what you're saying? Why would it be the case?

apbard: no, i am comparing this two "raises" checks with the case at lines 389 and following. The only difference between data2_wrong and data2 is the class name. I think that is similar to what you mention about inheritance in the other comment thread.

test/test_tensorclass_nofuture.py
```diff
@@ -354,6 +354,52 @@ class MyDataNested:
...
+    )
+    data2_wrong = MyDataMemMap2(
+        x=MemmapTensor.from_tensor(torch.ones(3, 4, 5)),
+        y=torch.ones(3, 4, 5),
+        batch_size=[3, 4],
+    )
+    with pytest.raises(
+        ValueError, match="__setitem__ is only allowed for same-class assignement"
+    ):
+        data1[:2] = data2_wrong[:2]
```
vmoens, line 382: it's not super clear to me why we don't want to allow this?
We should be able to write a tensor on a memmap-tensor and inversely

apbard: Here we are actually writing a `MyDataMemMap2` on `MyDataMemMap1` and it is currently forbidden (hence the pytest.raises), but I am totally in favour to change this behaviour as per previous discussion.

vmoens: Yeah I would expect it to work, or at least not be disabled by design

vmoens: Should I do it? Do we want to merge this first?

apbard: I would merge. I can have a look at that

vmoens (approved):

PR #208, opened 2023-02-08
[CI] Temporarily disable torchrec tests

apbard:
torchrec tests consistently fail due to upstream library. We should disable them until upstream is fixed and they can pass


vmoens (approved):
LGTM
feels good to see a green CI

PR #210, opened 2023-02-08
[BugFix] `__eq__` refactoring

vmoens:
## Description

Allows for tensor, int, float and tensordict/class comparison.

cc @apbard @tcbegley

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -538,6 +538,38 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
     self._tensordict._batch_size_setter(new_size)
 
 
+def _state_dict(self):
+    """Returns a state_dict dictionary that can be used to save and load data from a tensorclass."""
+    state_dict = self._tensordict.state_dict()
+    state_dict["_non_tensordict"] = {
+        key: value if not is_tensorclass(value) else value.state_dict()
+        for key, value in self._non_tensordict.items()
+    }
+    return state_dict
+
+
+def _load_state_dict(self, state_dict):
+    """Loads a state_dict attemptedly in-place on the destination tensorclass."""
+    for key, item in state_dict.items():
+        # keys will never be nested which facilitates everything, but let's
+        # double check in case someone does something nasty
+        if not isinstance(key, str):
+            raise TypeError("Only str keys are allowed when calling load_state_dict.")
+        if key in self._tensordict.keys():
+            self._tensordict.set(key, item, inplace=True)
+        elif key == "_non_tensordict":
+            for sub_key, sub_item in item.items():
+                if is_tensorclass(self._non_tensordict.get(sub_key, None)):
+                    self._non_tensordict[sub_key].load_state_dict(sub_item)
+                elif isinstance(sub_item, dict) and "_non_tensordict" in sub_item:
+                    raise RuntimeError(
+                        "Loading a saved tensorclass on a uninitialized tensorclass is not allowed"
+                    )
+                else:
+                    self._non_tensordict[sub_key] = sub_item
+    return self
+
+
 def _any(self, dim: int = None):
     """A recursive implementation of `any()` over the tensorclass leaves.
 
@@ -606,38 +638,6 @@ def _any(self, dim: int = None):
     )
 
 
-def _state_dict(self):
-    """Returns a state_dict dictionary that can be used to save and load data from a tensorclass."""
-    state_dict = self._tensordict.state_dict()
-    state_dict["_non_tensordict"] = {
-        key: value if not is_tensorclass(value) else value.state_dict()
-        for key, value in self._non_tensordict.items()
-    }
-    return state_dict
-
-
-def _load_state_dict(self, state_dict):
-    """Loads a state_dict attemptedly in-place on the destination tensorclass."""
-    for key, item in state_dict.items():
-        # keys will never be nested which facilitates everything, but let's
-        # double check in case someone does something nasty
-        if not isinstance(key, str):
-            raise TypeError("Only str keys are allowed when calling load_state_dict.")
-        if key in self._tensordict.keys():
-            self._tensordict.set(key, item, inplace=True)
-        elif key == "_non_tensordict":
-            for sub_key, sub_item in item.items():
-                if is_tensorclass(self._non_tensordict.get(sub_key, None)):
-                    self._non_tensordict[sub_key].load_state_dict(sub_item)
-                elif isinstance(sub_item, dict) and "_non_tensordict" in sub_item:
-                    raise RuntimeError(
-                        "Loading a saved tensorclass on a uninitialized tensorclass is not allowed"
-                    )
-                else:
-                    self._non_tensordict[sub_key] = sub_item
-    return self
-
-
 def _all(self, dim: int = None):
     """A recursive implementation of `all()` over the tensorclass leaves.
 
@@ -684,20 +684,23 @@ def _all(self, dim: int = None):
         >>> assert c.all(-1).y.batch_size == c.all(1).y.batch_size
 
     """
-    if dim is not None and dim < 0:
+    if dim is None:
+        return self._tensordict.all() and all(
+            value.all()
+            for value in self._non_tensordict.values()
+            if is_tensorclass(value)
+        )
+
+    if dim < 0:
         dim = self.batch_dims + dim
 
-    if dim is not None:
-        non_tensor = {
-            key: None if not is_tensorclass(obj) else obj.all(dim=dim)
-            for key, obj in self._non_tensordict.items()
-        }
-        return self._from_tensordict(
-            self._tensordict.all(dim=dim),
-            non_tensor,
-        )
-    return self._tensordict.all() and all(
-        value.all() for value in self._non_tensordict.values() if is_tensorclass(value)
+    non_tensor = {
+        key: None if not is_tensorclass(obj) else obj.all(dim=dim)
+        for key, obj in self._non_tensordict.items()
+    }
+    return self._from_tensordict(
+        self._tensordict.all(dim=dim),
+        non_tensor,
     )
 
 
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -540,7 +540,7 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
 
 def _state_dict(self):
     """Returns a state_dict dictionary that can be used to save and load data from a tensorclass."""
-    state_dict = self._tensordict.state_dict()
+    state_dict = {"_tensordict": self._tensordict.state_dict()}
     state_dict["_non_tensordict"] = {
         key: value if not is_tensorclass(value) else value.state_dict()
         for key, value in self._non_tensordict.items()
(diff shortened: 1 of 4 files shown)


tensordict/prototype/tensorclass.py
```diff
@@ -536,21 +536,38 @@ def __eq__(self, other):
...
+            tensordict or any compatible type (int, float or tensor), in
+            which case the equality check will be propagated to the leaves.
 
     Returns:
         False if the objects are of different class types, Tensorclass of boolean values for tensor attributes and None for non-tensor attributes
 
     """
-    if not isinstance(other, self.__class__):
+    if not is_tensorclass(other) and not isinstance(
+        other, (TensorDictBase, int, float, Tensor)
```
apbard, line 548: maybe we can use np.isscalar?

vmoens: sure

test/test_tensorclass_nofuture.py
```diff
@@ -1058,6 +1058,54 @@ class MyClass:
...
+    assert (a == a.clone()).all()
+    assert (a != 1.0).any()
+    assert (a.y == 1).all()
+    assert (a != torch.ones([])).any()
+    assert (a.y == torch.ones([])).all()
+    assert (a == b).all()
+    assert (b == a).all()
+    assert (a == c).all()
+    assert (c == a).all()
+
```
apbard, line 1107: add either here or in the indexing/getitem test a check with sliced tensorclass

test/test_tensorclass_nofuture.py
```diff
@@ -1058,6 +1058,54 @@ class MyClass:
...
+
```
apbard, line 1108: add a check on ne

tensordict/prototype/tensorclass.py
vmoens: Nope:
```
>>> np.isscalar("x")
True
```

apbard: from numpy doc of isscalar:
If you need a stricter way to identify a numerical scalar, use isinstance(x, numbers.Number), as that returns False for most non-numerical elements such as strings.

vmoens:
TODO:
- [ ] Test all()
- [ ] test any()
- [ ] document all and any

test/test_tensorclass_nofuture.py
vmoens: I had to implement `all` and `any` in the process

vmoens:
@apbard do you mind reviewing this? Thanks!

tensordict/prototype/tensorclass.py
apbard: i don't see the updated code with the isinstance call. forgot to commit?

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+        return self._from_tensordict(
+            self._tensordict.any(dim=dim),
+            non_tensor,
+        )
+    return self._tensordict.any() or any(
+        value.any() for value in self._non_tensordict.values() if is_tensorclass(value)
+    )
+
+
+def _state_dict(self):
```
apbard, line 586: why in this PR? I don't see tests for these

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+    state_dict = self._tensordict.state_dict()
+    state_dict["_non_tensordict"] = {
+        key: value if not is_tensorclass(value) else value.state_dict()
+        for key, value in self._non_tensordict.items()
+    }
+    return state_dict
+
+
+def _load_state_dict(self, state_dict):
+    """Loads a state_dict attemptedly in-place on the destination tensorclass."""
```
apbard, line 597: we are not checking that state dict contains all the keys needed

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+        elif key == "_non_tensordict":
+            for sub_key, sub_item in item.items():
+                if is_tensorclass(self._non_tensordict.get(sub_key, None)):
+                    self._non_tensordict[sub_key].load_state_dict(sub_item)
+                elif isinstance(sub_item, dict) and "_non_tensordict" in sub_item:
+                    raise RuntimeError(
+                        "Loading a saved tensorclass on a uninitialized tensorclass is not allowed"
+                    )
+                else:
+                    self._non_tensordict[sub_key] = sub_item
```
apbard, line 614: this way we are allowing key creation.

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+        # double check in case someone does something nasty
+        if not isinstance(key, str):
+            raise TypeError("Only str keys are allowed when calling load_state_dict.")
+        if key in self._tensordict.keys():
+            self._tensordict.set(key, item, inplace=True)
+        elif key == "_non_tensordict":
+            for sub_key, sub_item in item.items():
+                if is_tensorclass(self._non_tensordict.get(sub_key, None)):
+                    self._non_tensordict[sub_key].load_state_dict(sub_item)
+                elif isinstance(sub_item, dict) and "_non_tensordict" in sub_item:
```
apbard, line 609: add a comment to clarify this if. Is not immediate to understand. E.g.
```suggestion
                # sub_item is the state dict of a tensorclass
                elif isinstance(sub_item, dict) and "_non_tensordict" in sub_item:
```

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+        ...     torch.randn(3, 4),
+        ...     MyClass(torch.randn(3, 4, 1), None, batch_size=[3, 4, 1]),
+        ...     batch_size=[3, 4])
+        >>> _ = c.all()  # bool
+        >>> # one can work with negative dimensions, which will be converted to positive dims for the root tensorclass
+        >>> assert (c.all(-1) == c.all(1)).all()
+        >>> assert c.all(-1).y.batch_size == c.all(1).y.batch_size
+
+    """
+    if dim is not None and dim < 0:
```
apbard, line 647: see suggestion above

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
     Args:
-        other: object to compare to this object
+        other: object to compare to this object. Can be a tensorclass, a
+            tensordict or any compatible type (int, float or tensor), in
+            which case the equality check will be propagated to the leaves.
 
     Returns:
         False if the objects are of different class types, Tensorclass of boolean values for tensor attributes and None for non-tensor attributes
 
     """
```
apbard, line 675: maybe add an example?

tensordict/prototype/tensorclass.py
```diff
@@ -532,25 +537,166 @@ def _batch_size_setter(self, new_size: torch.Size) -> None:
...
+        >>> @tensorclass
+        ... class MyClass:
+        ...      X: Tensor
+        ...      y: "MyClass"
+        ...
+        >>> c = MyClass(
+        ...     torch.randn(3, 4),
+        ...     MyClass(torch.randn(3, 4, 1), None, batch_size=[3, 4, 1]),
+        ...     batch_size=[3, 4])
+        >>> _ = c.all()  # bool
```
apbard, line 641: add examples of returned values

tensordict/tensordict.py
```diff
@@ -959,10 +959,16 @@ def __ne__(self, other: object) -> TensorDictBase:
             tensors of the same shape as the original tensors.
 
         """
-        if not isinstance(other, (TensorDictBase, dict, float, int)):
-            return False
+        from tensordict.prototype import is_tensorclass
```
apbard, line 962: why import here? to avoid circular import?

tensordict/tensordict.py
```diff
@@ -959,10 +959,16 @@ def __ne__(self, other: object) -> TensorDictBase:
...
+
+        if not isinstance(
+            other, (TensorDictBase, dict, float, int, torch.Tensor, MemmapTensor)
+        ) and not is_tensorclass(other):
+            return True
         if not isinstance(other, TensorDictBase) and isinstance(other, dict):
```
apbard, line 968: I would probably switch the order:
- tensordictbase
- dict
- tensorclass
- numeric + tensor
- other (aka return True) 

reading "positive" ifs is easier than all these negations

tensordict/tensordict.py
```diff
@@ -988,10 +994,16 @@ def __eq__(self, other: object) -> TensorDictBase:
             tensors of the same shape as the original tensors.
 
         """
-        if not isinstance(other, (TensorDictBase, dict, float, int)):
+        from tensordict.prototype import is_tensorclass
+
```
apbard, line 998: same as above

vmoens:
@apbard i made the changes, have a look

tensordict/prototype/tensorclass.py
vmoens: because when introducing the non-tensor data feature, we broke `all` which was working out of the box, but now was causing some silent error in the state-dict test
Basically, before all the tensor data was stored in a single tensordict which made it simple to fall back on tensordict methods.
Now we will have to check one-by-one that all methods have a proper fallback since they are not automatically routed towards `self._non_tensordict`

tensordict/prototype/tensorclass.py
vmoens: we can't do the exact same thing since any works with an `or` and all with and `and`
but i can flip the order

tensordict/prototype/tensorclass.py
apbard: exactly

tensordict/prototype/tensorclass.py
apbard: imagine you have a "filled" tensorclass. then you load it from a "partial" state dict. what would you expect as a result?
- a raise
- a tensorclass with the "partial" state + others None
- a tensorclass with the "partial" state + others old value

vmoens: To me the latter would make more sense -- at least with the use cases I have in mind.

tensordict/tensordict.py
```diff
@@ -959,32 +960,31 @@ def __ne__(self, other: object) -> TensorDictBase:
...
 
         """
+        # avoiding circular imports
         from tensordict.prototype import is_tensorclass
 
-        if not isinstance(
-            other, (TensorDictBase, dict, float, int, torch.Tensor, MemmapTensor)
-        ) and not is_tensorclass(other):
-            return True
         if not isinstance(other, TensorDictBase) and isinstance(other, dict):
```
apbard, line 966: move this after the TensordictBase
it will be just: `if isinstance(other, dict)`

tensordict/tensordict.py
```diff
@@ -994,30 +994,29 @@ def __eq__(self, other: object) -> TensorDictBase:
...
 
         """
+        # avoiding circular imports
         from tensordict.prototype import is_tensorclass
 
-        if not isinstance(
-            other, (TensorDictBase, dict, float, int, torch.Tensor, MemmapTensor)
-        ) and not is_tensorclass(other):
-            return False
         if not isinstance(other, TensorDictBase) and isinstance(other, dict):
```
apbard, line 1000: as before

tensordict/tensordict.py
vmoens: it needs to be before since there is no return enclosed

vmoens: but only the `isinstance(other, dict)` is necessary

apbard: right, I overlooked it, but why you need make_tensordict? the code that checks eq/ne seems to perform actions valid also for dict.

vmoens: indeed it works without it, let's remove it then

apbard (approved):
LGTM

PR #211, opened 2023-02-08
[BugFix] Fix `_getitem_batch_size` in various edge cases.

tcbegley:
## Description

This PR fixes various bugs I encountered in `_getitem_batch_size` where for particular choices of index the following test would fail

```python
assert tensor[index].shape == _getitem_batch_size(tensor, index)
```

I first noticed this in the case where the index was a tuple of tensor-like indices. Indeed something as simple as

```python
tensor = torch.rand(2, 2)
index = ([0], [0])
assert tensor[index].shape == _getitem_batch_size(tensor, index)
```

currently fails on `main`. The correct behaviour when there are tuples of tensor-like indices present is to broadcast them onto a common shape, and then extract the diagonal entry. So `tensor[list1, list1]` is roughly equivalent to `torch.stack([tensor[i, j] for i, j in zip(list1, list2)])`. The final shape depends slightly on whether those tensor-like indices were contiguous or not.

Handling Boolean masks is also a bit delicate. From the [NumPy docs on advanced indexing](https://numpy.org/doc/stable/user/basics.indexing.html#advanced-indexing) it turns out that masking is equivalent to converting the mask to a tuple of integer indices with `mask.nonzero()` which can equivalently be written `mask.nonzero().unbind(-1)` if `mask` is a PyTorch tensor rather than a NumPy array.

I've added tests which check thousands of indexing combinations (these can probably be relaxed a bit), though it's possible I've missed something so feedback is welcome!

### DIFF
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -101,7 +101,7 @@ def _getitem_batch_size(shape: torch.Size, items: INDEX_TYPING) -> torch.Size:
     for _item in items:
         if isinstance(_item, (list, np.ndarray)):
             _item = torch.tensor(_item)
-        if isinstance(_item, torch.Tensor):
+        elif isinstance(_item, torch.Tensor):
             # np.broadcast will complain if we give it CUDA tensors
             _item = _item.cpu()
         if isinstance(_item, torch.Tensor) and _item.dtype is torch.bool:
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -101,6 +101,9 @@ def _getitem_batch_size(shape: torch.Size, items: INDEX_TYPING) -> torch.Size:
     for _item in items:
         if isinstance(_item, (list, np.ndarray)):
             _item = torch.tensor(_item)
+        if isinstance(_item, torch.Tensor):
+            # np.broadcast will complain if we give it CUDA tensors
+            _item = _item.cpu()
         if isinstance(_item, torch.Tensor) and _item.dtype is torch.bool:
             # when using NumPy's advanced indexing patterns, any index containing a
             # boolean array can be equivalently replaced with index.nonzero()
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        pytest.skip(
+            "Not enough dimensions in test tensor for this combination of parameters"
+        )
+    mask_shape = (2, 3, 4, 5)[idx : idx + ndim]
+    mask = torch.randint(2, mask_shape, dtype=torch.bool)
+    if slice_leading_dims:
+        index = (slice(None),) * idx + (mask,)
+    else:
+        index = (0,) * idx + (mask,)
+    assert tensor[index].shape == _getitem_batch_size(tensor.shape, index)
```
vmoens, line 84: can we add a 
```
if __name__ == "__main__":
   argparse.blabla
```

here to be runnable with python too (helps me debug)?

test/test_utils.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
```
vmoens, line 1: can you add headers?

tensordict/utils.py
```diff
@@ -99,32 +96,69 @@ def _getitem_batch_size(
...
+            if prev is not None and i != prev + 1:
+                contiguous = False
+            prev = i
+        elif isinstance(_item, Number) and prev is not None and i == prev + 1:
+            prev = i
+
     bs = []
+    if tensor_indices:
+        try:
+            b = np.broadcast(*tensor_indices)
```
vmoens, line 136: this is the offending bit in the tests
np expects cpu tensors, non cuda.

vmoens (approved):
LGTM thanks for working on this!
There's just the cuda errors to fix and we're good to go

tcbegley:
Thanks! I've addressed your comments.

PR #212, opened 2023-02-09
[Test] use temp_path instead of tmpdir

apbard:
I am switching from tmpdir to the new tmp_path fixture.
@tcbegley, @vmoens is there a reason we were defining our custom tmpdir instead of using the regular pytest one?


tcbegley (approved):
LGTM

>  is there a reason we were defining our custom tmpdir instead of using the regular pytest one?

The reason would be that I didn't know about the PyTest one 😅.

PR #223, opened 2023-02-17
[Feature] MemmapTensor.empty_like and TensorDict.memmap_like

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -254,6 +254,37 @@ class MemmapTensor:
         out.copy_(tensor)
         return out
 
+    @classmethod
+    def empty_like(
+        cls,
+        tensor: Union[torch.Tensor, MemmapTensor, np.ndarray],
+        transfer_ownership=False,
+        prefix=None,
+        filename: Optional[str] = None,
+        mode: str = "r+",
+    ) -> MemmapTensor:
+        if isinstance(tensor, np.ndarray):
+            raise TypeError(
+                "Convert input to torch.Tensor before calling MemmapTensor."
+            )
+        device = tensor.device if hasattr(tensor, "device") else torch.device("cpu")
+        dtype = (
+            tensor.dtype
+            if isinstance(tensor, (torch.Tensor, MemmapTensor))
+            else numpy_to_torch_dtype_dict[tensor.dtype.name]
+        )
+        shape = tensor.shape
+        out = cls(
+            shape,
+            device=device,
+            dtype=dtype,
+            prefix=prefix,
+            transfer_ownership=transfer_ownership,
+            filename=filename,
+            mode=mode,
+        )
+        return out
+
     @staticmethod
     def _create_memmap_with_index(memmap_tensor, index):
         memmap_copy = copy(memmap_tensor)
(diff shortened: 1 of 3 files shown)


tensordict/memmap.py
```diff
@@ -254,6 +254,37 @@ def from_tensor(
...
+        if isinstance(tensor, np.ndarray):
+            raise TypeError(
+                "Convert input to torch.Tensor before calling MemmapTensor."
+            )
+        device = tensor.device if hasattr(tensor, "device") else torch.device("cpu")
+        dtype = (
+            tensor.dtype
+            if isinstance(tensor, (torch.Tensor, MemmapTensor))
+            else numpy_to_torch_dtype_dict[tensor.dtype.name]
+        )
```
tcbegley, line 275: If `tensor` is an instance of `np.ndarray` then this code isn't actually reachable (we would have raised a `TypeError` on line 267) so I think the ternary operator isn't necessary.

tensordict/memmap.py
```diff
@@ -254,6 +254,37 @@ def from_tensor(
...
+        transfer_ownership=False,
+        prefix=None,
+        filename: Optional[str] = None,
+        mode: str = "r+",
+    ) -> MemmapTensor:
+        if isinstance(tensor, np.ndarray):
+            raise TypeError(
+                "Convert input to torch.Tensor before calling MemmapTensor."
+            )
+        device = tensor.device if hasattr(tensor, "device") else torch.device("cpu")
```
tcbegley, line 270: Similarly here, given the type hints and the fact that `np.ndarray` would have already result in error, I think we can just do `tensor.dtype`.

tensordict/tensordict.py
```diff
@@ -1156,6 +1156,74 @@ def memmap_(self, prefix=None, copy_existing=False) -> TensorDictBase:
         """
         raise NotImplementedError(f"{self.__class__.__name__}")
 
+    def memmap_like(self, prefix=None) -> TensorDictBase:
```
tcbegley, line 1159: There's quite a bit of shared logic here with `memmap_`. I wonder if there's some way to reuse.

I guess a dumb version of this function would be something like

```python
return self.clone().memmap_(...)
```

But that results in unnecessary copies.

tensordict/tensordict.py
```diff
@@ -2735,7 +2803,9 @@ def detach_(self) -> TensorDictBase:
             value.detach_()
         return self
 
-    def memmap_(self, prefix=None, copy_existing=False) -> TensorDictBase:
+    def memmap_(
+        self, prefix=None, copy_existing=False, create_empty=False
```
tcbegley, line 2807: What is `create_empty` for?

tcbegley (approved):
LGTM

As I said in a comment, it feels like it should be possible to deduplicate some code but I don't think it's trivial, so repeating is fine for now imo.

tensordict/tensordict.py
vmoens: removed :)

tensordict/tensordict.py
vmoens: no not in the case of `expand` above
`clone()` will allocate the memory on device / ram! That's a disaster if you have a tensor that is small but expanded millions of times

PR #226, opened 2023-02-20
[Feature] enable broadcast on set_item

apbard:


test/test_tensorclass_nofuture.py
```diff
@@ -419,6 +434,35 @@ class MyDataMemMap3:
         data_wrong_cls[2:] = data1[2:]
 
 
+@pytest.mark.parametrize(
+    "val2broadcast",
+    [
+        0,
+        torch.zeros(4, 5),
+        TensorDict({"X": torch.zeros(2, 4, 5)}, batch_size=[2, 4]),
+        MemmapTensor.from_tensor(torch.zeros(4, 5)),
```
vmoens, line 443: Here we're creating objects from the library on the global space, which means that if memmap is broken we won't even launch the tests (since instantiation will occur at launch time, not during the test). Do you thing we can handle this differently?

tensordict/tensordict.py
```diff
@@ -2086,45 +2087,65 @@ def __getitem__(self, idx: INDEX_TYPING) -> TensorDictBase:
...
+                except AttributeError as err:
+                    if "for populating tensordict with new key-value pair" in str(err):
+                        raise RuntimeError(
+                            "Trying to replace an existing nested tensordict with "
+                            "another one with non-matching keys. This leads to "
+                            "unspecified behaviours and is prohibited."
+                        )
+                    raise err
+                return
+        # else:
```
vmoens, line 2146: we can remove this

tensordict/tensordict.py
```diff
@@ -2086,45 +2087,65 @@ def __getitem__(self, idx: INDEX_TYPING) -> TensorDictBase:
...
-                else sub_index
-                for sub_index in index
-            )
-        if isinstance(index, tuple) and sum(
-            isinstance(_index, str) for _index in index
-        ) not in [len(index), 0]:
-            raise IndexError(_STR_MIXED_INDEX_ERROR)
+            try:
+                # try to cast value to tensor:
+                value = self._convert_to_tensor(value)
```
vmoens, line 2101: we had a long discussion at the time with @zeenolife [here](https://github.com/pytorch/rl/pull/438) IIRC and we decided that it was better to leave `set` do the convertion rather than having `__setitem__` do it. 
What is the reason to move this to `__setitiem__`?

tensordict/tensordict.py
```diff
@@ -2021,9 +2021,10 @@ def _index_tensordict(self, idx: INDEX_TYPING):
     def __getitem__(self, idx: INDEX_TYPING) -> TensorDictBase:
         """Indexes all tensors according to the provided index.
 
-        Returns a new tensordict where the values share the storage of the original tensors (even
-        when the index is a torch.Tensor). Any in-place modification to the
-        resulting tensordict will impact the parent tensordict too.
+        Returns a new tensordict where the values share the storage of the
+        original tensors (even when the index is a torch.Tensor).
```
vmoens, line 2025: Actually reading this makes me think it is not true for `torch.Tensor` indices. I think now you need to call `get_subtensordict` explicitely to get a tensor that shares the same data pointer.

apbard: It's a "doctring issue" or code one? I haven't changed __getitem__. Just followed linter warning for the docstring

vmoens: it's a docstring issue, more a note to myself than a comment on this PR

tensordict/tensordict.py
apbard: didn't know about this discussion. I Placed here as a mean to check for compatible type, "fail fast" if the value type is not correct and to provide a meaningful error message to the user.
Also, as mentioned in chat, I would give another shot at that discussion and think about refactoring the way preprocessing/input validation is handled. 
I personally think that, for the sake of clarity and ease of maintenance, that it would be better to place "processing"/input validation calls only where needed and constraints checked where they are "declared" instead of bury them in low level calls and resort to flags to determine whether they are applied or not.

tensordict/tensordict.py
apbard: Opened #240

tensordict/tensordict.py
apbard: done

test/test_tensorclass_nofuture.py
apbard: switched from object to types. It's ugly...I think there should be a more clean solution.

tensordict/tensordict.py
apbard: > I think I see your point. Aren't we at risk of missing calls to `_convert_to_tensor` by doing this? 

if we miss calls and that goes undetected it means our tests are lacking and, thus, I would consider it is a false problem. 

> Why isn't it the case already that preprocessing of input tensors is achieved only where needed? 
The other risk is that we could be calling preprocessing multiple times [...] allows us to code the key type check only once.

My sentence was probably not that clear. The main point it is about clarity and "ownership" of the checks. E.g., Why are delegating the "value" validation but not that of the keys?
  
I think the risk of multiple validation could apply here as well if you add a check not knowing/forgetting that its checked down in the chain or setting the "do_check" flags to True.

> Regarding flags, they can be changed / removed, they're private and we can do what we want with them. If there's a better alternative we can consider it.

I would decouple the input validation from the private set function into two calls:
- validation/preprocess 
- actual set

this would remove the need for the flags (*), make the check/validation explicit and "visible" at high level. If you adopt the convention that private methods do not check the inputs (**) you should have not "multiple calls".

* unless you would like to expose them to the user.
**that is you need to check only user inputs not data you produce

zeenolife: > we had a long discussion at the time with @zeenolife https://github.com/pytorch/rl/pull/438 IIRC and we decided that it was better to leave set do the convertion rather than having __setitem__ do it.

@vmoens AFAIR, the idea was exactly not to risk missing `_convert_to_tensor` method, as you've already said above.

vmoens: > this would remove the need for the flags (*), make the check/validation explicit and "visible" at high level. If you adopt the convention that private methods do not check the inputs (**) you should have not "multiple calls".

Sure, that sounds reasonable. I guess something like
```python
def set(self, key, value, inplace=False):
    key = self._validate_key(key)
```
Then if we want to skip the validation we just call the private `_set` method, is that what you had in mind?

apbard: yep, exactly. In practice it could still require some ifs or multiple validation steps but this is the idea

tensordict/prototype/tensorclass.py
```diff
@@ -481,7 +483,26 @@ def _setitem(self, item: NestedKey, value: Any) -> None:
...
+        )
+
+    # if it is one of accepted "broadcast" types
+    if not is_tensorclass(value):
+        # attempt broadcast on all tensordata and nested tensorclasses
+        self._tensordict[item] = value
+        for key, val in self._non_tensordict.items():
+            if is_tensorclass(val):
+                _setitem(self._non_tensordict[key], item, value)
+        return
```
tcbegley, line 502: I believe as a result of this change, [this code](https://github.com/pytorch-labs/tensordict/pull/226/files#diff-8327839160f3184ce4a4e04c9e19ec4afbf392c5c3705fe8a3c534162d32e6eeR515-R517) below is now unreachable. Perhaps we can restructure the control flow here?

tensordict/prototype/tensorclass.py
```diff
@@ -481,7 +483,26 @@ def _setitem(self, item: NestedKey, value: Any) -> None:
...
+
+    # else it is tensorclass
+    if not isinstance(value, self.__class__):
```
tcbegley, line 505: Perhaps this can be moved inside `else` block under above `if` statement and merged with `else` block below. I.e.

```python
if not is_tensorclass(value):
    ...
else:
    if not isinstance(value, self.__class__):
        # check keys
    # contents of else block below
```

Could even consider swapping these blocks so the initial statement is `if is_tensorclass(value)`. I find minimising negations makes things easier to read back.

tcbegley (commented):
LGTM.

Just to check I understand, the changes in `tensordict.py` are largely just refactoring / cleaning up the logic? Afaict the only change in functionality is [here](https://github.com/pytorch-labs/tensordict/pull/226/files#diff-8f1c52b037105554a3287e9a3a531069c1e9d64c64c019b43e0b8ed5f1a3be78R2611-R2613), is that right or did I miss something?

apbard:
> LGTM.
> 
> Just to check I understand, the changes in `tensordict.py` are largely just refactoring / cleaning up the logic? Afaict the only change in functionality is [here](https://github.com/pytorch-labs/tensordict/pull/226/files#diff-8f1c52b037105554a3287e9a3a531069c1e9d64c64c019b43e0b8ed5f1a3be78R2611-R2613), is that right or did I miss something?

after your improvements on set* methods, yes. this is the only functional change

tensordict/prototype/tensorclass.py
apbard: right, done

tensordict/prototype/tensorclass.py
apbard: I agree on negation, I chose to keep it in this case to avoid a long nested if that will also decrease readability. 
But I can swap if you prefer so.

apbard: didn't notice we can get rid off of the return. definitely worth swapping then

tcbegley (approved):
LGTM! Just needs a `pre-commit run --all-files` I think.

PR #230, opened 2023-02-22
[Feature] Distributed capabilities

vmoens:
## Description

Kicks off the work of distributing tensordict content across devices.
Allows for sync and async coordination.

See tests and docstrings for examples

cc @albertbou92


vmoens:
FYI convergage decreases because we don't test coverage across procs.
That's something we should do

tcbegley (approved):
LGTM, cool stuff! Small inline comment.

PR #221, opened 2023-02-17
[Doc] Port torchrl's tutorials where they belong

vmoens:
## Description

Port and modifies the TensorDict tutorials from torchrl to this repo.

@tcbegley worked on splitting them in smaller standalone pieces, quicker to consume and easier to read.

PR #231, opened 2023-02-22
[Docs] Add simplified reshape tutorial

tcbegley:
## Description

This PR adds a simplified and self-contained tutorial for reshaping tensordicts.

There is currently some redundancy between this and `tensordict_basics.py`. I think ultimately the latter should be deleted / completely split up before #221 is merged.

### DIFF
--- a/tutorials/sphinx_tuto/tensordict_shapes.py
+++ b/tutorials/sphinx_tuto/tensordict_shapes.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+"""
+Manipulating the shape of a TensorDict
+======================================
+**Author**: `Vincent Moens <https://github.com/vmoens>`_
+
+In this tutorial you will learn how to manipulate the shape of a ``TensorDict`` and its
+contents.
+"""
+
+##############################################################################
+# When we create a ``TensorDict`` we specify a ``batch_size``, which must agree with
+# the leading dimensions of all entries in the ``TensorDict``. Since we have a guarantee
+# that all entries share those dimensions in common, ``TensorDict`` is able to expose
+# a number of methods with which we can manipulate the shape of the ``TensorDict`` and
+# its contents.
+
+# sphinx_gallery_start_ignore
+import warnings
+
+warnings.filterwarnings("ignore")
+# sphinx_gallery_end_ignore
+import torch
+from tensordict.tensordict import TensorDict
+
+##############################################################################
+# Reshaping a ``TensorDict``
+# --------------------------
+#
+# ``TensorDict.reshape`` works just like ``torch.Tensor.reshape``. It applies to all of
+# the contents of the ``TensorDict`` along the batch dimensions - note the shape of
+# ``b`` in the example below. It also updates the ``batch_size`` attribute.
+
+a = torch.rand(3, 4)
+b = torch.rand(3, 4, 5)
+tensordict = TensorDict({"a": a, "b": b}, batch_size=[3, 4])
+
+reshaped_tensordict = tensordict.reshape(-1)
+assert reshaped_tensordict.batch_size == torch.Size([12])
+assert reshaped_tensordict["a"].shape == torch.Size([12])
+assert reshaped_tensordict["b"].shape == torch.Size([12, 5])
+
+##############################################################################
+# Splitting a ``TensorDict``
+# --------------------------
+#
+# ``TensorDict.split`` is similar to ``torch.Tensor.split``. It splits the
+# ``TensorDict`` into chunks. Each chunk is a ``TensorDict`` with the same structure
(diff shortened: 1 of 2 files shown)


tutorials/sphinx_tuto/tensordict_shapes.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+"""
+Manipulating the shape of a TensorDict
+======================================
+**Author**: `Vincent Moens <https://github.com/vmoens>`_
```
vmoens, line 5: Put your name there mate!

tutorials/sphinx_tuto/tensordict_shapes.py
```diff
@@ -0,0 +1,117 @@
...
+warnings.filterwarnings("ignore")
+# sphinx_gallery_end_ignore
+import torch
+from tensordict.tensordict import TensorDict
+
+##############################################################################
+# Reshaping a ``TensorDict``
+# --------------------------
+#
+# ``TensorDict.reshape`` works just like ``torch.Tensor.reshape``. It applies to all of
```
vmoens, line 30: We can use `:func:ˋ for functions, `:class:` for classes and `:mod:` for modules. If the path is complete this will generate a link to the api reference in the rendered tutorial.

tutorials/sphinx_tuto/tensordict_shapes.py
```diff
@@ -0,0 +1,117 @@
...
+# in cases where you have a large ``TensorDict`` with many entries, and you don't need
+# to stack all of them.
+
+cloned_tensordict = tensordict.clone()
+# no stacking happens on the next line
+stacked_tensordict = torch.stack([tensordict, cloned_tensordict], dim=0)
+print(stacked_tensordict)
+
+##############################################################################
+# If we index a ``LazyStackedTensorDict`` we recover the original ``TensorDict``.
```
vmoens, line 77: Index along the stack dimension

tutorials/sphinx_tuto/tensordict_shapes.py
```diff
@@ -0,0 +1,117 @@
...
+
+assert stacked_tensordict[0] is tensordict
+assert stacked_tensordict[1] is cloned_tensordict
+
+##############################################################################
+# Accessing a key in the ``LazyStackedTensorDict`` results in those values being
+# stacked. If the key corresponds to a nested ``TensorDict`` then we will recover
+# another ``LaxyStackedTensorDict``.
```
vmoens, line 85: Let's add a note to tell that accessing the value multiple times stacks it multiple times, which is inefficient. One should call contiguous if there is not specific advantage in using a lazy stack.

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

tutorials/sphinx_tuto/tensordict_shapes.py
```diff
@@ -0,0 +1,117 @@
...
+
+assert stacked_tensordict["a"].shape == torch.Size([2, 3, 4])
+
+###############################################################################
+# If we want to have a contiguous ``TensorDict``, we can call ``.to_tensordict()``
```
vmoens, line 90: Here it is haha
Maybe let's decorate this comment with a .. note::

tutorials/sphinx_tuto/tensordict_shapes.py
```diff
@@ -0,0 +1,117 @@
...
+# Concatenation is not done lazily, instead calling ``torch.cat`` on a list of
+# ``TensorDict`` instances simply returns a ``TensorDict`` whose entries are the
+# concatenated entries of the elements of the list.
+
+concatenated_tensordict = torch.cat([tensordict, cloned_tensordict], dim=0)
+assert isinstance(concatenated_tensordict, TensorDict)
+assert concatenated_tensordict.batch_size == torch.Size([6, 4])
+assert concatenated_tensordict["b"].shape == torch.Size([6, 4, 5])
+
+##############################################################################
```
vmoens, line 110: Missing:
Unsqueeze, squeeze, permute, gather, indexing, setting items (if they won't be covered elsewhere)
For those where a dim is passed we should mention what behaviour is expected when using negative dim (eg unsqueeze(-1) adds one singleton after the last dim of the batch size, not the last of each tensor) and what will happen for nested tensordicts with more dims in that case (ie if a nested TD has 1 more dim than the root, unsqueeze(-1) should do an unsqueeze(-2) on it).

Finally we can also say that `apply` can be used for all the functions that are not covered.

tutorials/sphinx_tuto/tensordict_shapes.py
tcbegley: Oops, I copied a different tutorial as a starting point and missed that.

vmoens (approved):
LGTM!

PR #234, opened 2023-02-23
[BugFix] Fix sorting of keys in send and recv

vmoens:
## Description

Fixes cases where the key order mismatch between a src and dest tensordict.

The `sorted_keys` attribute is intentionally not adapted to `tensorclass` as this would require to duplicate a lot of code + we will probably refactor tensorclass soon.


tcbegley (approved):
LGTM

PR #235, opened 2023-02-24
[Refactor] Rename set_default -> setdefault

tcbegley:
## Description

This PR renames `set_default` to `setdefault` for parity with Python's `dict`.

I've actually left `set_default` in place for now with a deprecation warning as TorchRL uses `setdefault` and so we would break their CI with a straight switch. Once this lands, we can update TorchRL, then drop the alias here and hopefully not cause any disruption.


vmoens (approved):
LGTM

PR #238, opened 2023-02-25
[Feature] Full type annotations

XuehaiPan:
## Description

Add full type annotations and fix some bugs, e.g.:

1. Add missing return statements
2. Remove extra return statements in property setters.
3. Remove extra `metaclass=abc.ABCMeta` when inheriting from `collections.abc.Mapping`
4. Change base class from `Mapping` to `MutableMapping`
5. Remove extra `@classmethod` decoration for `Class.__new__` and use `super().__new__(...)`.

Some enhancement and style fixes:

6. Change relative imports to absolute imports.
7. Use postponed type evaluation when `from __future__ import annotations` enabled.
8. Change `Union[XXX, YYY]` to `XXX | YYY` when `from __future__ import annotations` enabled.
9. CHange constants to `UPPER_CASE` and type alias to `CamelCase`.

## Motivation and Context

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds core functionality)

## Checklist

- [X] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)
- [X] My change requires a change to the documentation.
- [X] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [X] I have updated the documentation accordingly.

### DIFF
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -12,7 +12,7 @@ import warnings
 from copy import copy, deepcopy
 from pathlib import Path
 from tempfile import _TemporaryFileWrapper
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable
 
 import numpy as np
 import torch
@@ -58,7 +58,7 @@ def implements_for_memmap(torch_function) -> Callable:
     return decorator
 
 
-def to_numpy(tensor: Union[torch.Tensor, np.ndarray]) -> np.ndarray:
+def to_numpy(tensor: torch.Tensor | np.ndarray) -> np.ndarray:
     if isinstance(tensor, torch.Tensor):
         return tensor.detach().cpu().numpy()
     else:
@@ -138,8 +138,8 @@ class MemmapTensor:
         device: DEVICE_TYPING = None,
         dtype: torch.dtype = None,
         transfer_ownership: bool = False,
-        prefix: Optional[str] = None,
-        filename: Optional[str] = None,
+        prefix: str | None = None,
+        filename: str | None = None,
         mode: str = "r+",
     ):
         self.idx = None
@@ -159,7 +159,8 @@ class MemmapTensor:
         if prefix is not None:
             warnings.warn(
                 "prefix has been deprecated. If you want to control the location of "
-                "the MemmapTensor on disk, consider using filename instead."
+                "the MemmapTensor on disk, consider using filename instead.",
+                stacklevel=2,
             )
             if filename is not None:
                 raise ValueError("filename and prefix should not both be specified")
@@ -208,10 +209,10 @@ class MemmapTensor:
     @classmethod
     def from_tensor(
         cls,
-        tensor: Union[torch.Tensor, MemmapTensor, np.ndarray],
+        tensor: torch.Tensor | MemmapTensor | np.ndarray,
         transfer_ownership=False,
         prefix=None,
-        filename: Optional[str] = None,
+        filename: str | None = None,
         mode: str = "r+",
     ) -> MemmapTensor:
         if isinstance(tensor, MemmapTensor):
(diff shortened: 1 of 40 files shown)


vmoens (commented):
Hey thanks for this
We'd need to check if these things don't break our internal CI before merging -- I'm afraid it will.
We calibrated the versions of pre-commit and related packages to the ones of our internal infra.

vmoens:
Great work by the way!

vmoens:
> Hey thanks for this
> We'd need to check if these things don't break our internal CI before merging -- I'm afraid it will.
> We calibrated the versions of pre-commit and related packages to the ones of our internal infra.

@XuehaiPan do you think we could start by doing the changes without upgrading the linters and then see if we can upgrade? Otherwise if the upgrade is not compatible with Meta's infra we won't be able to merge anything

XuehaiPan:
> @XuehaiPan do you think we could start by doing the changes without upgrading the linters and then see if we can upgrade?

Yes, we can. I will revert the linter changes.

tensordict/utils.py
```diff
@@ -655,7 +651,9 @@ def _get_item(tensor: torch.Tensor, index):
         return tensor[index]
 
 
-def _set_item(tensor: torch.Tensor, value, index):
+def _set_item(
+    tensor: torch.Tensor, value: torch.Tensor, index: IndexType
+) -> torch.Tensor:
```
XuehaiPan, line 656: The function signature is kind of wared here.

```python
setitem(tensor, value, index)
setitem(tensor, index, value)
```

vmoens: It's a private method, we can change it freely if we want :)

tensordict/memmap.py
```diff
@@ -12,53 +12,42 @@
...
-    def decorator(func):
+    def decorator(func: Callable) -> Callable:
         MEMMAP_HANDLED_FN[torch_function] = func
         return func
 
     return decorator
 
 
-def to_numpy(tensor: Union[torch.Tensor, np.ndarray]) -> np.ndarray:
+def to_numpy(tensor: torch.Tensor | np.ndarray) -> np.ndarray:
```
vmoens, line 50: Can you elaborate on why we're moving away from union?

tensordict/memmap.py
```diff
@@ -637,7 +619,7 @@ def __setitem__(self, idx: INDEX_TYPING, value: torch.Tensor):
                 )
             self.memmap_array[idx] = to_numpy(value)
 
-    def __setstate__(self, state: Dict[str, Any]) -> None:
+    def __setstate__(self, state: dict[str, Any]) -> None:
```
vmoens, line 622: Why dict instead of Dict?

vmoens (approved):
Il happy with these changes. Truly amazing work.
Can you comment on "why" some of these style changes?
Like dict vs Dict, Union vs | etc
I'd be interested to know why one is better than the other

tensordict/memmap.py
XuehaiPan: They are semanticly identical. The `|` operator on classes syntax was introduced in Python 3.10 and can be enabled by `from __future__ import annotations` in Python 3.7. This is more clear and does not need unnecessary imports for `Union` and `Optional`.

tensordict/memmap.py
XuehaiPan: Since Python 3.9, all Python builtin container types support `__class_getitem__`. For example, `list[int]`, `dict[str, Tensor]`, `collections.OrderedDict[str, Tensor | None]`, `collections.abc.Sequence[int]`. This behavior can be enabled by `from __future__ import annotations` in Python 3.7 since the annotations are postend evaluated. The old `typing.List`, `typing.Dict`, etc., are deprecated and may be removed after 5 years since Python 3.9. That is 2024 when Python 3.13 is released.

XuehaiPan: See https://docs.python.org/3/library/typing.html#typing.Dict

> Deprecated since version 3.9: [builtins.dict](https://docs.python.org/3/library/stdtypes.html#dict) now supports subscripting ([]). See [PEP 585](https://peps.python.org/pep-0585/) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias).

PR #239, opened 2023-02-26
[BugFix] Allow for nested output keys in Probabilitic modules

vmoens:
## Description

Addresses https://github.com/pytorch/rl/issues/927

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -39,10 +39,25 @@ __all__ = [
 def _check_all_str(sequence_of_str: Sequence[str]) -> None:
     if isinstance(sequence_of_str, str):
         raise RuntimeError(
-            f"Expected a sequence of strings but got a string: {sequence_of_str}"
+            f"Expected a sequence of strings or tuples of strings, but got a string: {sequence_of_str}"
         )
-    if any(not isinstance(key, str) for key in sequence_of_str):
-        raise TypeError(f"Expected a sequence of strings but got: {sequence_of_str}")
+    for key in sequence_of_str:
+        if isinstance(key, str):
+            continue
+        elif isinstance(key, tuple):
+            if not len(key):
+                raise RuntimeError("Empty tuples are not permitted.")
+            for sub_key in key:
+                if isinstance(sub_key, str):
+                    continue
+                else:
+                    RuntimeError(
+                        f"Expected a sequence of strings or tuples of strings, but got a string: {sequence_of_str}"
+                    )
+        else:
+            RuntimeError(
+                f"Expected a sequence of strings or tuples of strings, but got a string: {sequence_of_str}"
+            )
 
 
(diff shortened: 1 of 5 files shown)


tensordict/nn/common.py
```diff
@@ -39,10 +39,25 @@ class FunctionalModuleWithBuffers:
...
+            continue
+        elif isinstance(key, tuple):
+            if not len(key):
+                raise RuntimeError("Empty tuples are not permitted.")
+            for sub_key in key:
+                if isinstance(sub_key, str):
+                    continue
+                else:
+                    RuntimeError(
+                        f"Expected a sequence of strings or tuples of strings, but got a string: {sequence_of_str}"
```
tcbegley, line 55: I think we can lose `a string` from this error message. It makes sense on line 42, but here the `sub_key` could be any type

vmoens: You're right
I addresses the comments, can you have a look? It's ready on my side

tensordict/nn/common.py
```diff
@@ -39,10 +39,25 @@ class FunctionalModuleWithBuffers:
...
+                if isinstance(sub_key, str):
+                    continue
+                else:
+                    RuntimeError(
+                        f"Expected a sequence of strings or tuples of strings, but got: {sequence_of_str}"
+                    )
+        else:
+            RuntimeError(
+                f"Expected a sequence of strings or tuples of strings, but got: {sequence_of_str}"
+            )
```
tcbegley, line 60: We also have the `_check_all_nested` function defined immediately below this which, after these changes, is doing essentially the same thing. 

I took a look, and the only place that `_check_all_str` is used is inside the probabilistic module, so perhaps these two functions can be merged?

tcbegley (commented):
Seems that using nested `in_keys` is not possible currently, because the keys are passed directly to `get_dist` as kwargs for the distribution.

```python
>>> prob_module = ProbabilisticTensorDictModule(
...     in_keys=[("param", "loc"), ("param", "scale")],
...     out_keys=["action"],
...     distribution_class=Normal,
...     return_log_prob=True,
... )
```

I get `TypeError: keywords must be strings`. 

~Perhaps we could consider changing the logic in `get_dist` to something like~

```diff
  dist_kwargs = {
-     dist_key: tensordict[td_key]
+     dist_key[-1] if isinstance(dist_key, tuple) else dist_key: tensordict[td_key]
      for dist_key, td_key in self.in_keys.items()
  }
```

~Then the assumption would be that the final key in the nested sequence is the kwarg, but would allow the used to store them somewhere other than the top level of the tensordict.~

EDIT - actually perhaps a better solution might be to modify how we convert `in_keys` from a list of `str` to a `dict`?

tensordict/nn/common.py
vmoens: We need to keep it for the distribution kwargs but we can use the other in all other cases, let me do that

vmoens:
@tcbegley Ready for review

PR #241, opened 2023-02-27
[Docs, Tutorials] Memory management tutorial

tcbegley:
## Description

This PR adds a tutorial to cover both `TensorDict.to` / `TensorDict.cuda` and `TensorDict.memmap_`.

Because the docs are built on a CI runner without GPUs available (as far as I'm aware), most of the code snippets here are code-blocks inside the restructured text.


tutorials/sphinx_tuto/tensordict_memory.py
```diff
@@ -0,0 +1,145 @@
...
+import warnings
+
+import torch
+from tensordict import TensorDict
+
+warnings.filterwarnings("ignore")
+# sphinx_gallery_end_ignore
+
+tensordict = TensorDict({"a": torch.rand(10), "b": {"c": torch.rand(10)}}, [10])
+tensordict.memmap_()
```
vmoens, line 123: A crucial point here is the `memmap_like` method. In general, it should be preferred (as it does not write the anything on the entire array) and allows to build things that are populated a small fraction at a time.
Can we add this?

vmoens (approved):
LGTM!

PR #242, opened 2023-02-27
[BugFix] Return self from `clear_device`

tcbegley:
## Description

While working on the tutorials I noticed that `clear_device` doesn't return anything, unlike the closely related `to` method.

This small change returns `self`, allowing `clear_device` to be chained with other methods.

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -390,8 +390,9 @@ class TensorDictBase(MutableMapping):
     def device(self, value: DeviceType) -> None:
         raise NotImplementedError
 
-    def clear_device(self) -> None:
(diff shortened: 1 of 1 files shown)


tensordict/tensordict.py
```diff
@@ -390,8 +390,9 @@ def device(self) -> torch.device | None:
     def device(self, value: DeviceType) -> None:
         raise NotImplementedError
 
-    def clear_device(self) -> None:
+    def clear_device(self) -> TensorDictBase:
```
vmoens, line 393:
```suggestion
    def clear_device(self) -> TensorDictBase:
        """Clears the device of the tensordict.
        
        Returns: self
        
        """
```

PR #243, opened 2023-02-27
[Docs, Tutorials] Add indexing tutorial

tcbegley:
## Description

This PR adds an indexing / masking tutorial. It is the last major bit of content to be moved out of the basic tutorial before we can deprecate that.


vmoens (approved):
LGTM

PR #246, opened 2023-02-27
[BugFix] Memmap-like for lazy stacked tds

vmoens:
## Description

Allow for memmap-like with heterogeneous tds.


tcbegley (approved):
LGTM!

PR #247, opened 2023-02-28
[Feature] Nesting tensorclass instances in tensordicts

vmoens:
## Description

- Allows for tensorclass instances to be used as values for a TensorDict
- Adds a series of tests for this use case
- Removes the tensorclass instances from `tensorclass._non_tensordict`

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -18,14 +18,10 @@ from pathlib import Path
 from textwrap import indent
 from typing import Any, Callable, Sequence, TypeVar, Union
 
+import tensordict
+
 import torch
-from tensordict.tensordict import (
-    _ACCEPTED_CLASSES,
-    get_repr,
-    is_tensordict,
-    TensorDict,
-    TensorDictBase,
-)
+from tensordict.tensordict import get_repr, is_tensordict, TensorDict, TensorDictBase
 from tensordict.utils import DeviceType, NestedKey
 from torch import Tensor
 
@@ -181,6 +177,7 @@ def tensorclass(cls: T) -> T:
     cls.__doc__ = f"{cls.__name__}{inspect.signature(cls)}"
 
     CLASSES_DICT[cls.__name__] = cls
+    tensordict.tensordict._ACCEPTED_CLASSES += [cls]
     return cls
 
 
@@ -237,24 +234,26 @@ def _init_wrapper(init: Callable) -> Callable:
 
 
 def _from_tensordict_wrapper(expected_keys):
-    def wrapper(cls, tensordict, non_tensordict=None):
+    def wrapper(cls, td, non_tensordict=None):
         """Tensor class wrapper to instantiate a new tensor class object
 
         Args:
-            tensordict (TensorDict): Dictionary of tensor types
+            td (TensorDict): Dictionary of tensor types
             non_tensordict (dict): Dictionary with non-tensor and nested tensor class objects
 
         """
+        if not isinstance(td, TensorDictBase):
+            raise RuntimeError(f"Expected a TensorDictBase instance but got {type(td)}")
         # Validating keys of tensordict
-        for key in tensordict.keys():
+        for key in td.keys():
             if key not in expected_keys:
                 raise ValueError(
-                    f"Keys from the tensordict ({set(tensordict.keys())}) must "
+                    f"Keys from the tensordict ({set(td.keys())}) must "
                     f"correspond to the class attributes ({expected_keys})."
                 )
 
         # Validating non-tensor keys and for key clash
-        tensor_keys = set(tensordict.keys())
+        tensor_keys = set(td.keys())
         if non_tensordict is not None:
             for key in non_tensordict.keys():
                 if key not in expected_keys:
@@ -270,7 +269,8 @@ def _from_tensordict_wrapper(expected_keys):
         # empty tensordict and writing values to it. we can skip this because we already
         # have a tensordict to use as the underlying tensordict
         tc = cls.__new__(cls)
-        tc.__dict__["_tensordict"] = tensordict
+        assert isinstance(td, TensorDictBase), td
+        tc.__dict__["_tensordict"] = td
 
         tc.__dict__["_non_tensordict"] = (
             non_tensordict if non_tensordict is not None else {}
@@ -322,6 +322,7 @@ def _getattribute_wrapper(getattribute: Callable) -> Callable:
                 "_tensordict" in self.__dict__
                 and item in self.__dict__["_tensordict"].keys()
             ):
+                assert isinstance(self._tensordict, TensorDictBase)
                 out = self._tensordict[item]
                 expected_type = self.__dataclass_fields__[item].type
                 out = _get_typed_output(out, expected_type)
@@ -359,7 +360,7 @@ def _setattr_wrapper(setattr_: Callable, expected_keys: set[str]) -> Callable:
                 f"Cannot set the attribute '{key}', expected attributes are {expected_keys}."
             )
 
-        if isinstance(value, _ACCEPTED_CLASSES):
+        if isinstance(value, tuple(tensordict.tensordict._ACCEPTED_CLASSES)):
             # Avoiding key clash, honoring the user input to assign tensor type data to the key
             if key in self._non_tensordict.keys():
                 del self._non_tensordict[key]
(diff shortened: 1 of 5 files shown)


tensordict/prototype/tensorclass.py
```diff
@@ -270,7 +269,8 @@ def wrapper(cls, tensordict, non_tensordict=None):
         # empty tensordict and writing values to it. we can skip this because we already
         # have a tensordict to use as the underlying tensordict
         tc = cls.__new__(cls)
-        tc.__dict__["_tensordict"] = tensordict
+        assert isinstance(td, TensorDictBase), td
```
tcbegley, line 272: I think it's better to avoid `assert` statements for input validation because they can be disabled with certain interpreter flags. We can raise a `TypeError` or something instead.

tensordict/prototype/tensorclass.py
```diff
@@ -237,24 +234,26 @@ def wrapper(
 
 
 def _from_tensordict_wrapper(expected_keys):
-    def wrapper(cls, tensordict, non_tensordict=None):
+    def wrapper(cls, td, non_tensordict=None):
         """Tensor class wrapper to instantiate a new tensor class object
 
         Args:
-            tensordict (TensorDict): Dictionary of tensor types
+            td (TensorDict): Dictionary of tensor types
```
tcbegley, line 241: Instead of renaming the variables throughout, which I think has a minor negative impact on the quality of docstrings and function signatures, what if we alias the import of `tensordict`?

tensordict/prototype/tensorclass.py
```diff
@@ -322,6 +322,7 @@ def wrapper(self, item: str) -> Any:
                 "_tensordict" in self.__dict__
                 and item in self.__dict__["_tensordict"].keys()
             ):
+                assert isinstance(self._tensordict, TensorDictBase)
```
tcbegley, line 325: ditto

tensordict/prototype/tensorclass.py
```diff
@@ -359,7 +360,7 @@ def wrapper(self, key: str, value: Any) -> None:
                 f"Cannot set the attribute '{key}', expected attributes are {expected_keys}."
             )
 
-        if isinstance(value, _ACCEPTED_CLASSES):
+        if isinstance(value, tuple(tensordict.tensordict._ACCEPTED_CLASSES)):
```
tcbegley, line 363: Perhaps we could maintain `_ACCEPTED_CLASSES` as a tuple rather than casting it every time we call `setattr`?

tensordict/prototype/tensorclass.py
```diff
@@ -427,14 +428,11 @@ def _getitem(self, item: NestedKey) -> Any:
...
-        raise ValueError("Invalid indexing arguments.")
+        raise ValueError(f"Invalid indexing arguments: {item}.")
     tensor_res = self._tensordict[item]
     non_tensor_res = {}
     for key, value in self._non_tensordict.items():
-        if is_tensorclass(value):
-            non_tensor_res[key] = _getitem(value, item)
-        else:
-            non_tensor_res[key] = value
+        non_tensor_res[key] = value
```
tcbegley, line 435: This is just making a shallow copy now right? Could do something like `non_tensor_res = copy(self._non_tensordict)` or `non_tensor_res = dict(self._non_tensordict)` instead

tensordict/prototype/tensorclass.py
```diff
@@ -590,13 +578,7 @@ def _memmap_like(
...
     out_other = {}
     for key, val in self._non_tensordict.items():
-        if is_tensorclass(val):
-            if prefix is not None:
-                out_other[key] = val.memmap_like(prefix=prefix / key)
-            else:
-                out_other[key] = val.memmap_like()
-        else:
-            out_other[key] = val
+        out_other[key] = val
```
tcbegley, line 581: similarly here, we can just do `out_other = dict(self._non_tensordict)` for a shallow copy

tensordict/prototype/tensorclass.py
```diff
@@ -638,8 +620,7 @@ def _state_dict(self) -> dict[str, Any]:
     """Returns a state_dict dictionary that can be used to save and load data from a tensorclass."""
     state_dict = {"_tensordict": self._tensordict.state_dict()}
     state_dict["_non_tensordict"] = {
-        key: value if not is_tensorclass(value) else value.state_dict()
-        for key, value in self._non_tensordict.items()
+        key: value for key, value in self._non_tensordict.items()
     }
```
tcbegley, line 624: ditto

tensordict/prototype/tensorclass.py
```diff
@@ -731,21 +710,12 @@ def _any(self, dim: int | None = None) -> bool:
...
+        return self._tensordict.any()
 
     if dim < 0:
         dim = self.batch_dims + dim
 
-    non_tensor = {
-        key: None if not is_tensorclass(obj) else obj.any(dim=dim)
-        for key, obj in self._non_tensordict.items()
-    }
+    non_tensor = {key: None for key, obj in self._non_tensordict.items()}
```
tcbegley, line 718: Can iterate over keys here, no need for `obj`

tensordict/prototype/tensorclass.py
```diff
@@ -799,19 +769,12 @@ def _all(self, dim: int | None = None) -> bool:
...
+        return self._tensordict.all()
 
     if dim < 0:
         dim = self.batch_dims + dim
 
-    non_tensor = {
-        key: None if not is_tensorclass(obj) else obj.all(dim=dim)
-        for key, obj in self._non_tensordict.items()
-    }
+    non_tensor = {key: None for key, obj in self._non_tensordict.items()}
```
tcbegley, line 777: ditto

tensordict/prototype/tensorclass.py
```diff
@@ -855,21 +818,7 @@ def _get_out(key, nontensor):
...
-        index_expand = index_expand.expand(target_shape)
-        return index_expand
-
-    non_tensor = {
-        key: obj
-        if not is_tensorclass(obj)
-        else obj.gather(dim=dim, index=_expand_index(obj), out=_get_out(key, True))
-        for key, obj in self._non_tensordict.items()
-    }
+    non_tensor = {key: obj for key, obj in self._non_tensordict.items()}
```
tcbegley, line 821: can just make a shallow copy

tensordict/tensordict.py
```diff
@@ -111,11 +111,13 @@ class KeyedJaggedTensor:
...
 
 def is_tensordict(datatype: type | Any) -> bool:
+    from tensordict.prototype import is_tensorclass
+
     return (
         issubclass(datatype, TensorDictBase)
         if isinstance(datatype, type)
         else isinstance(datatype, TensorDictBase)
-    )
+    ) or is_tensorclass(datatype)
```
tcbegley, line 120: Seems a bit surprising to me that `is_tensordict(some_tensorclass)` would return `True`. What's the motivation for this?

tcbegley (approved):
LGTM. Left a few inline comments, but nothing major.

tensordict/prototype/tensorclass.py
vmoens: yeah it was for debugging thanks for spotting this

tensordict/prototype/tensorclass.py
vmoens: I wanted it to be mutable to make it easy to have all modules see the changes no matter what we do with it.
Can we have the same result with a tuple?

tensordict/tensordict.py
vmoens: with this PR tensorclass becomes interchangeable with tensordict in many cases. We could rename the check though, I agree it's confusing.

vmoens: To elaborate: every time we check if something is a tensordict, it is to propagate a method to its fields. The same is now done if it's a tensorclass.

vmoens: @tcbegley I added a docstring. I'm cheating a bit, `is_tensordict` stands for "if from the tensordict library" 😅

tensordict/tensordict.py
```diff
@@ -111,11 +111,30 @@ class KeyedJaggedTensor:
 
 
 def is_tensordict(datatype: type | Any) -> bool:
+    """Checks if a data object or a type is a tensor container from the tensordict lib.
+
+    Returns:
+        ``True`` if the input is a TensorDictBase subclass, a tensorclass or an istance of these.
```
tcbegley, line 117:
```suggestion
        ``True`` if the input is a TensorDictBase subclass, a tensorclass or an instance of these.
```

tensordict/prototype/tensorclass.py
tcbegley: I believe so, because all modules are actually accessing the same variable, rather than different references to the same object. So reassigning that variable is seen by all.

I will validate that though.

tcbegley:
I agree `is_tensordict_or_tensorclass` is clunky. I don't have a better suggestion. Maybe something like `is_tensor_collection` or `is_tensor_container` but I don't think those are really any less confusing.

tensordict/prototype/tensorclass.py
tcbegley: I think I may actually be wrong on this point. Let's leave it as it is. Casting to tuple for small number of elements like that is pretty cheap anyway.

vmoens:
What about `from_tensordict_lib`?
`is_tensor_collection` could also work.

tcbegley:
Happy with either of them. Maybe a slight preference for the latter, otherwise `from_tensordict_lib(TensorDictModule)` or `from_tensordict_lib(MemmapTensor)` etc. could be confusing.

PR #250, opened 2023-02-28
[Refactor] Reduce boilerplate in tensorclass methods

tcbegley:
## Description

This PR reduces boilerplate in tensorclass methods. Many of them copy the metadata or replace entries with `None`. Here we create two helper functions that can be called from inside the methods to streamline the code.

### DIFF
--- a/tensordict/prototype/tensorclass.py
+++ b/tensordict/prototype/tensorclass.py
@@ -16,7 +16,7 @@ import warnings
 from copy import copy
 from dataclasses import dataclass
 from textwrap import indent
-from typing import Any, Callable, Protocol, Sequence, TypeVar, Union
+from typing import Any, Callable, Sequence, TypeVar, Union
 
 import tensordict as tensordict_lib
 
@@ -33,12 +33,6 @@ from torch import Tensor
 
 T = TypeVar("T", bound=TensorDictBase)
 PY37 = sys.version_info < (3, 8)
-
-
-class TensorClassProtocol(Protocol):
-    _tensordict: TensorDictBase
-    _non_tensordict: dict[str, Any]
-
 
 # We keep a dict of str -> class to call the class based on the string
 CLASSES_DICT: dict[str, type] = {}
@@ -129,17 +123,6 @@ def tensorclass(cls: T) -> T:
 
 
     """
-    tc_handled_functions: dict[Callable, Callable] = {}
-
-    def implements_for_tdc(torch_function: Callable) -> Callable:
-        """Register a torch function override for _TensorClass."""
-
-        @functools.wraps(torch_function)
-        def decorator(func: Callable) -> Callable:
-            tc_handled_functions[torch_function] = func
-            return func
-
-        return decorator
 
     def __torch_function__(
         cls,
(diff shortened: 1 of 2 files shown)


vmoens (commented):
Is there a way of having even less code than that?
For instance: if we remove `_ones_like` we can do
```
def __getattr__(self, attr):
    if attr not in self.__dict__:
        attr = getattr(self._tensordict, attr)
        if callable(attr):
            attr = wrap(attr)
    return attr

def wrap(fun):
    def new_fun(*args, **kwargs):
        out = fun(*args, **kwargs)
        if isinstance(out, TensorDictBase):
            return MyClass._from_tensordict(out, out._non_tensordict)
        return out
```

For split and unbind we'd need to check if the output is a tuple of tensordicts but you get the gist.
For `__all__` and `__any__` we would need to erase the meta data though.

Is it feasible? I think it would greatly reduce the amount of code not?

tensordict/prototype/tensorclass.py
```diff
@@ -812,18 +680,11 @@ def _gather(self, dim: int, index: torch.Tensor, out: TensorDictBase | None = No
...
-            return out._tensordict
-
-    non_tensor = copy(self._non_tensordict)
-    return self._from_tensordict(
-        self._tensordict.gather(dim=dim, index=index, out=_get_out(None, False)),
-        non_tensor,
+    return _from_tensordict_with_copy(
+        self,
+        self._tensordict.gather(
+            dim=dim, index=index, out=out._tensordict if out is not None else None
```
tcbegley, line 686: I wanted to drop `_gather` but I can't figure out how to generically intercept `out` and replace it with `out._tensordict`.

tensordict/prototype/tensorclass.py
```diff
@@ -397,26 +442,22 @@ def _getattr(self, attr: str) -> Any:
...
-            else:
-                non_tensor_dict[key] = value
         if isinstance(res, TensorDictBase):
-            new = self._from_tensordict(res, non_tensor_dict)
-            return new
-        else:
-            return res
+            if attr.endswith("_"):
+                # in-place operation, return the current object
+                return self
```
tcbegley, line 451: I think this is fine? If we ever added a method with a trailing underscore that wasn't an in-place operation it could cause some problems, but that would already be a big break from convention

tensordict/prototype/tensorclass.py
```diff
@@ -123,14 +148,34 @@ def __torch_function__(
...
+                tc = input_[0]
+                input_ = input_.__class__(tc._tensordict for tc in input_)
+            else:
+                tc = input_
+                input_ = input_._tensordict
 
-        return td_handled_functions[func](*args, **kwargs)
+            out = kwargs.get("out")
+            if out is not None:
+                kwargs["out"] = out._tensordict
```
tcbegley, line 171: I think this is too brittle, as `out` could be passed as a positional argument in which case this breaks. I wonder if we should instead replace _all_ values in `args` and `kwargs` that happen to be tensorclasses with `value._tensordict`?

tensordict/prototype/tensorclass.py
```diff
@@ -104,14 +129,14 @@ def tensorclass(cls: T) -> T:
...
     def implements_for_tdc(torch_function: Callable) -> Callable:
         """Register a torch function override for _TensorClass."""
 
         @functools.wraps(torch_function)
         def decorator(func: Callable) -> Callable:
-            td_handled_functions[torch_function] = func
+            tc_handled_functions[torch_function] = func
             return func
 
         return decorator
```
tcbegley, line 142: We're actually not using this at all anymore, and could perhaps remove it?

tensordict/prototype/tensorclass.py
```diff
@@ -15,16 +15,16 @@
 import warnings
 from copy import copy
 from dataclasses import dataclass
-from pathlib import Path
 from textwrap import indent
-from typing import Any, Callable, Sequence, TypeVar, Union
+from typing import Any, Callable, Protocol, Sequence, TypeVar, Union
```
tcbegley, line 19: `Protocol` was only added in Python 3.8 so I'll rethink this

tensordict/prototype/tensorclass.py
```diff
@@ -123,14 +148,34 @@ def __torch_function__(
...
             kwargs = {}
-        if func not in td_handled_functions or not all(
-            issubclass(t, (Tensor, cls)) for t in types
-        ):
-            return NotImplemented
+        if func in _TD_PASS_THROUGH:
+            if len(args) > 0:
+                input_ = args[0]
+            else:
+                input_ = kwargs["input"] if "input" in kwargs else kwargs["tensors"]
```
vmoens, line 161:
```suggestion
                input_ = kwargs.get("input", kwargs["tensors"])
```

tensordict/prototype/tensorclass.py
vmoens: Yep I guess we could just scan kwargs and replace the tensorclasses by the tensordict they contain

tensordict/prototype/tensorclass.py
```diff
@@ -123,14 +148,34 @@ def __torch_function__(
         args: tuple[Any, ...] = (),
         kwargs: dict[str, Any] | None = None,
     ) -> Callable:
+        if (
```
vmoens, line 151: What happens to the metadata in all these cases? Is it copied from the original tensordict?
If this is too brittle we can only take care of the `tensorclass.method` and not the overridable torch functions.
We should also consider if this implementation does not cause a (small) overhead when scanning the input and output types in search of the strategy, and whether it won't break on some torch function we have not implemented yet.

I'm trying to understand what other torch function we may want to implement in the future.
To get a complete list of overridable functions run `torch.overrides.get_overridable_functions()`
There are some that we may want to implement like `index_fill`, most of these are algebraic ops though.

tensordict/prototype/tensorclass.py
tcbegley: I've done this

tensordict/prototype/tensorclass.py
tcbegley: Oh I actually had the same idea! This is changed in the latest commit

tensordict/prototype/tensorclass.py
vmoens: Ok then if you feel like this is something we can work with I'm happy to merge it

tcbegley: I think this amount of overhead is acceptable. The alternative would mean reintroducing a lot of the overrides that I've just deleted. There's a bit of a trade-off between having small amount of generic code with some overhead vs. a large amount of specific code.

The only question then is whether you're happy with how we handle the metadata. I'm not sure what the alternatives are tbh, and afaict this PR is at least not changing behaviour.

tensordict/prototype/tensorclass.py
vmoens: Yep I can already see a couple
`lock`
`unlock`
`pin_memory` IIRC
We'd need to edit these methods to match the `_` trailing convention

vmoens (approved):
LGTM

PR #251, opened 2023-03-01
[BugFix] Uncaught wrong instance check in update_at_

vmoens:
## Description

`Update_at_` was not tested and a wrong instance check was found in it.


tcbegley (approved):
LGTM

PR #252, opened 2023-03-01
[BugFix,Feature] Rename with nested keys

vmoens:
## Description

Allow `rename_key` to be called with nested keys


tcbegley (approved):
LGTM

PR #254, opened 2023-03-01
[Docs] Remove unneeded __len__ definition in tutorial

tcbegley:
## Description

As of #150 this is not needed, so removing from tutorial to avoid confusion.


vmoens (approved):
LGTM

PR #255, opened 2023-03-01
[Tests] Add tensorclass benchmarks

tcbegley:
## Description

This PR adds benchmarks for torch functions in tensorclasses that were used during development of #250


vmoens (approved):
LGTM

PR #256, opened 2023-03-03
[Feature] Add dispatch source and dest arguments

vmoens:
## Description

Adds a `source` and `dest` keyword arguments to `dispatch` (renamed from `dispatch_kwargs`).
One can also call the module with unnamed arguments.

Renames the `test_tensordictmodule.py` in `test_nn.py` which is clearer.

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -161,6 +161,37 @@ class dispatch:
         >>> assert (b == 1).all()
 
 
+    Since the input keys is a sorted sequence of strings,
+    :func:`dispatch` can also be used with unnamed arguments where the order
+    must match the order of the input keys.
+
+    .. note::
+
+        This can lead to unexpected behaviours when the first unnamed argument
+        is itself a tensordict instance with keys matching those specified
+        by the source.
+
+    Examples:
+        >>> class MyModuleNest(nn.Module):
+        ...     in_keys = [("a", "c"), "d"]
+        ...     out_keys = ["b"]
+        ...
+        ...     @dispatch
+        ...     def forward(self, tensordict):
+        ...         tensordict['b'] = tensordict['a', 'c'] + tensordict["d"]
+        ...         return tensordict
+        ...
+        >>> module = MyModuleNest()
+        >>> b, = module(torch.zeros(1, 2), d=torch.ones(1, 2))  # works
+        >>> assert (b == 1).all()
+        >>> b, = module(torch.zeros(1, 2), torch.ones(1, 2))  # works
+        >>> assert (b == 1).all()
+        >>> try:
+        ...     b, = module(torch.zeros(1, 2), a_c=torch.ones(1, 2))  # fails
+        ... except:
+        ...     print("oopsy!")
+        ...
+
     """
 
     DEFAULT_SEPARATOR = "_"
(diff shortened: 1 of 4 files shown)


test/test_nn.py
```diff
@@ -479,12 +479,49 @@ def test_dispatch_kwargs_nested(self):
         out = tdm(a_c=torch.zeros(1, 1))
         assert (out == td["b", "d"]).all()
 
+    def test_dispatch_kwargs_nested_confusing(self):
```
tcbegley, line 482: minor: should we rename the tests `dispatch_kwargs` -> `dispatch`

tensordict/nn/common.py
```diff
@@ -148,16 +230,34 @@ def __call__(self, func: Callable) -> Callable:
...
-            _self, tensordict: TensorDictBase | None = None, *args: Any, **kwargs: Any
-        ) -> Any:
+        def wrapper(_self, *args: Any, **kwargs: Any) -> Any:
+            source = self.source
+            if isinstance(source, str):
+                source = getattr(_self, source)
+            tensordict = None
+            if len(args):
+                tensordict = args[0]
+                if not isinstance(tensordict, TensorDictBase) or not all(
```
vmoens, line 240: Is this clunky? Or expensive? Is it worth it?

vmoens:
@tcbegley I added a bunch of stuff

vmoens:
Seems like tests are failing :(

vmoens:
@tcbegley Should be ready now.
It's a bit more restrictive but also clearer.

tensordict/nn/common.py
```diff
@@ -166,8 +269,8 @@ def wrapper(
                                 f"but is expected to execute that function."
                             )
                 tensordict = make_tensordict(tensordict_values)
-                out = func(_self, tensordict, *args, **kwargs)
-                out = tuple(out[key] for key in _self.out_keys)
+                out = func(_self, tensordict, **kwargs)
```
tcbegley, line 272: Why no `*args` here? Is it not possible that `source` has fewer entries than `args` and that we don't exhaust it?

tcbegley (approved):
Looks good except for one final comment. I think we want to restore `*args` in the forward pass?

tensordict/nn/common.py
vmoens: I made the change, but it's a bit confusing IMO.
Imagine this scenario:
```python
        class MyModuleNest(nn.Module):
            in_keys = [("a", "c"), "d"]
            out_keys = ["b"]

            @dispatch(separator="_")
            def forward(self, tensordict, other):
                tensordict["b"] = tensordict["a", "c"] + tensordict["d"] + other
                return tensordict

        module = MyModuleNest()
        other = 1
        (b,) = module(torch.zeros(1, 2), torch.ones(1, 2), other)
        assert (b == 2).all()
```
Now say you want to use `d` as a kwarg:
```python
        class MyModuleNest(nn.Module):
            in_keys = [("a", "c"), "d"]
            out_keys = ["b"]

            @dispatch(separator="_")
            def forward(self, tensordict, other):
                tensordict["b"] = tensordict["a", "c"] + tensordict["d"] + other
                return tensordict

        module = MyModuleNest()
        other = 1
        (b,) = module(torch.zeros(1, 2), other, d=torch.ones(1, 2), )
        assert (b == 2).all()
```
That breaks but with not apparent reason (it's not super clear from the function signature).

tcbegley: That's not great, but I can create the same problem with a `nn.Module` no?

```python
class MyModule(nn.Module):
    def forward(self, a, d, other):
        pass

module = MyModule()
other = 1
module(torch.zeros(10), other, d=torch.ones(10))
```

Isn't this just a case of the user providing arguments to the function incorrectly? If I want to use `other` as a positional argument, then I necessarily need to supply both `a` and `d` as positional. 

In any case, in your example don't we get an error from [this line](https://github.com/pytorch-labs/tensordict/pull/256/files#diff-731e98da172c0a815c68a2e01d1e3ba85a59a9b0bec820aff620537fd76b303aR259)?

PR #259, opened 2023-03-03
[Refactor] Refactor set methods

tcbegley:
## Description

This PR refactors `set` and its variants (`set_`, `set_at_` etc.), the primary goals being:

- Make it clearer which methods have responsibility to perform input validation, particularly with a view to ensuring that we're not incurring unnecessary overhead by duplicating checks
- Create a new private method `_set` which can be used to bypass runtime checks, in favour of passing around a flag `_run_checks`.

The basic idea is to have methods for validation of keys and values, perform the validation, and delegate the actual setting logic to `_set`. Something like

```python
def set(self, key, value, inplace=False):
    key = self._validate_key(key)
```

There are some difficulties here though. The first is with nested keys. Often validation of the value needs to be done in the context of the "leaf tensordict", specifically when checking the batch_size or casting the value to the device, it doesn't suffice to check against the device and batch_size of the root tensordict. As a result I've ended up doing something like this

```python
def set(self, key, value, inplace=False):
    key = self._validate_key(key)
    if isinstance(key, tuple):
        td, subkey = _get_leaf_tensordict(self, key)
        td.set(subkey, value)  # value validation happens in context of leaf
        return self
```

This pattern does mean that we perform checks in the correct context, and there is no duplication, but it does mean that we either end up with code duplication in set for handling nested keys, or we accept that `_set` is only able to handle string keys, and hence is not a drop in replacement for `set` but without runtime checks.

Very happy to have a discussion about how best to proceed here.

cc @apbard

### DIFF
--- a/tensordict/prototype/fx.py
+++ b/tensordict/prototype/fx.py
@@ -40,7 +40,7 @@ class TDGraphModule(nn.Module):
 
         for out_key, output in zip(self.out_keys, outputs):
             if out_key != "_":
-                tensordict_out.set(out_key, output, _run_checks=False)
+                tensordict_out.set(out_key, output)
 
         return tensordict_out
 
--- a/tensordict/prototype/fx.py
+++ b/tensordict/prototype/fx.py
@@ -40,7 +40,7 @@ class TDGraphModule(nn.Module):
 
         for out_key, output in zip(self.out_keys, outputs):
             if out_key != "_":
-                tensordict_out.set(out_key, output)
+                tensordict_out._set(out_key, output)
 
         return tensordict_out
 
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -369,7 +369,7 @@ class TensorDictBase(MutableMapping):
                 if len(tensordict.batch_size) < len(new_batch_size):
                     # document as edge case
                     tensordict.batch_size = new_batch_size
-                    self.set(key, tensordict)
+                    self._set(key, tensordict)
         self._check_new_batch_size(new_batch_size)
         self._change_batch_size(new_batch_size)
 
@@ -1124,7 +1124,6 @@ class TensorDictBase(MutableMapping):
         input_dict_or_td: dict[str, CompatibleType] | TensorDictBase,
         clone: bool = False,
         inplace: bool = False,
-        **kwargs: Any,
     ) -> TensorDictBase:
         """Updates the TensorDict with values from either a dictionary or another TensorDict.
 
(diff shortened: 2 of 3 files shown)


tensordict/prototype/fx.py
```diff
@@ -40,7 +40,7 @@ def forward(
 
         for out_key, output in zip(self.out_keys, outputs):
             if out_key != "_":
-                tensordict_out.set(out_key, output, _run_checks=False)
+                tensordict_out.set(out_key, output)
```
tcbegley, line 43: This change is a good example of where we would like to be able to call `_set` with nested keys support, as the runtime checks in the forward pass are very bad.

tensordict/tensordict.py
```diff
@@ -1112,7 +1110,11 @@ def apply(
             else:
                 item_trsf = fn(item, *_others)
             if item_trsf is not None:
-                out.set(key, item_trsf, inplace=inplace, _run_checks=False, **kwargs)
+                if isinstance(self, SubTensorDict):
+                    out.set(key, item_trsf, inplace=inplace)
+                else:
+                    out._set(key, item_trsf, inplace=inplace)
```
tcbegley, line 1116: This is to replicate the logic above, if `self` is a `SubTensorDict` we want to process the input, hence we call `set` rather than `_set`.

vmoens:
> This pattern does mean that we perform checks in the correct context, and there is no duplication, but it does mean that we either end up with code duplication in set for handling nested keys, or we accept that `_set` is only able to handle string keys, and hence is not a drop in replacement for `set` but without runtime checks.

having `_set` handling only strings seems fine to me at first glance.

tensordict/prototype/fx.py
vmoens: Oh I see.
What if we duplicate the logic:
```python
def set(self, key, value, inplace=False):
    key = self._validate_key(key)
    if isinstance(key, tuple):
        td, subkey = _get_leaf_tensordict(self, key)
        td.set(subkey, value)  # value validation happens in context of leaf
        return self
def _set(self, key, value, inplace=False):
    if isinstance(key, tuple):
        td, subkey = _get_leaf_tensordict(self, key)
        td._set(subkey, value)  # value validation happens in context of leaf
        return self
```

tcbegley: I think that might be the best approach. It's not ideal, but I don't see a better alternative.

vmoens (commented):
Superb
I would just add the possibility of using _set with nested keys to skip the checks (even if it requires us to duplicate the logic)

vmoens (approved):
LGTM!

PR #261, opened 2023-03-06
[Quality] More informative error message when an argument is None in TensorDictModule

vmoens:
## Description

`TensorDictModule` allows some arguments to be missing in the input tensordict, in which case a `None` value is used.
The error message that occurs if that value is actually necessary is misleading, always being something like "NoneType has no attribute shape" or similar.

We now capture errors, check if any input is None and if the error contains None, and turn the error message to something more indicative of what is going on.

Similarly, we capture errors like
```python
td = TensorDict({"a": torch.zeros(())}, [])
td["a", "b"]
```
to let users know what is going on with the wrongful indexing.

Suggestions are welcome!

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -3179,7 +3179,7 @@ class TensorDict(TensorDictBase):
                         if "has no attribute" in str(err):
                             raise ValueError(
                                 f"Expected a TensorDictBase instance but got {type(first_lev)} instead"
-                                f" for key '{key[0]}' in tensordict:\n{self}."
+                                f" for key '{key[0]}' and subkeys {key[1:]} in tensordict:\n{self}."
                             )
                 return self.get(key[0])
             return self._tensordict[key]
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -3173,7 +3173,14 @@ class TensorDict(TensorDictBase):
                     first_lev = self.get(key[0])
                     if len(key) == 2 and isinstance(first_lev, KeyedJaggedTensor):
                         return first_lev[key[1]]
-                    return first_lev.get(key[1:])
+                    try:
+                        return first_lev.get(key[1:])
+                    except AttributeError as err:
+                        if "has no attribute" in str(err):
+                            raise ValueError(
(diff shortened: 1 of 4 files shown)


tcbegley (approved):
LGTM

vmoens:
Tests are not passing let me have a look

test/test_tensordict.py
```diff
@@ -1443,7 +1443,7 @@ def test_tensordict_set(self, td_name, device):
         val2 = np.zeros(shape=(4, 3, 2, 1, 10))
         td.set_("key1", val2)
         assert (td.get("key1") == 0).all()
-        with pytest.raises((KeyError, AttributeError)):
+        with pytest.raises((KeyError, AttributeError, ValueError)):
```
tcbegley, line 1446: This test feels pretty loose at this point. It's possible e.g. something else goes wrong and a different `ValueError` gets raised. Should we limit the possibilities and maybe add a `match` argument?

tensordict/tensordict.py
```diff
@@ -3005,7 +3005,14 @@ def _set(self, key: str, value, inplace: bool = False) -> TensorDictBase:
         else:
             td, subkey = self, key
         if inplace:
-            td.get(subkey).copy_(value)
+            try:
+                td.get(subkey).copy_(value)
+            except KeyError as err:
+                raise err
+            except Exception as err:
```
tcbegley, line 3012: Can we be more specific than this? Or is the problem that we don't know what error the module will raise if passed a `None` value where something concrete is expected?

tcbegley (commented):
LGTM, though I wonder if we can be a bit more specific both in the exceptions that we catch and also that we test for?

tensordict/tensordict.py
vmoens: we still raise the error, we just say which key is culprit (to help debugging)
Why would we need to be more specific?

tcbegley: Yeah ok, I think this is actually ok. I was a bit wary because `except Exception` will catch everything, but as you say because we reraise it perhaps there's no danger here.

test/test_tensordict.py
vmoens: Solved

PR #265, opened 2023-03-06
[DOC] update docstring for set method

apbard:


tensordict/tensordict.py
```diff
@@ -3022,8 +3022,7 @@ def set(
     ) -> TensorDictBase:
         """Sets a value in the TensorDict.
 
-        If inplace=True (default is False), and if the key already exists, set will
-        call set_ (in-place setting).
+        See TensorDictBase.set
```
vmoens, line 3025: I'd rather have no docstring at all and keep the one from tensordictbase, which will be used where needed in the doc.
This can be put as a comment rather than a docstring.

tensordict/tensordict.py
```diff
@@ -3043,11 +3042,16 @@ def set(
...
         value = self._validate_value(value)
+        # not calling set_ to avoid re-validate key
         return self._set(key, value, inplace=inplace)
 
     def set_(
         self, key: str, value: dict[str, CompatibleType] | CompatibleType
     ) -> TensorDictBase:
+        """Sets a value in the TensorDict in-place. Assumes keys exists.
+
+        See TensorDictBase.set_
```
vmoens, line 3053: same as above

vmoens (approved):
LGTM

PR #266, opened 2023-03-06
[Refactor] Inplace methods

tcbegley:
## Description

This PR appends `_` to various methods that are inplace in order to match the PyTorch convention.

There are some that are arguably inplace but which I have not changed in order to match corresponding methods in PyTorch, namely:
- `recv`
- `irecv`
- `load_state_dict`


vmoens (approved):
LGTM! Clever way to approach this!

PR #267, opened 2023-03-06
[BugFix] Default value for lazy stack

vmoens:
## Description

Solves a bug where lazy stacks fails to retrieve default value when requested

### DIFF
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -1189,6 +1189,13 @@ class TestTensorDicts(TestTensorDictsBase):
         assert (td_view.get("a") == 1).all()
         assert (td.get("a") == 1).all()
 
+    def test_default_nested(self, td_name, device):
+        torch.manual_seed(1)
+        td = getattr(self, td_name)(device)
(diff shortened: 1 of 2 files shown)


test/test_tensordict.py
```diff
@@ -1189,6 +1189,13 @@ def test_view(self, td_name, device):
         assert (td_view.get("a") == 1).all()
         assert (td.get("a") == 1).all()
 
+    def test_default_nested(self, td_name, device):
+        torch.manual_seed(1)
+        td = getattr(self, td_name)(device)
+        default_val = torch.randn(())
+        timbers = td.get(("shiver", "my", "timbers"), default_val)
```
tcbegley, line 1196: I like the keys you come up with for tests 😛

tcbegley (approved):
LGTM!

test/test_tensordict.py
vmoens: just to check that my code is properly reviewed
I can check that box now!

PR #268, opened 2023-03-06
[Build] missing torchsnapshot dependency

apbard:


vmoens (approved):
~~LGTM~~ see my comment below

vmoens (approved):
LGTM thanks!

test/test_tensorclass_nofuture.py
```diff
@@ -11,7 +11,15 @@
...
 import torch
-import torchsnapshot
+
+try:
+    import torchsnapshot
+
+    _has_torchsnapshot = True
+except ImportError as err:
+    _has_torchsnapshot = False
+    TORCHSNAPSHOT_ERR = str(err)
```
vmoens, line 21: this should be created out of the try/except to be visible
TORCHSNAPSHOT_ERR=None will suffice

apbard: done in the try...except block like we did for TORCHREC

vmoens (approved):

PR #271, opened 2023-03-08
[Benchmarks] Perform multiple rounds in benchmarks

tcbegley:
## Description

Currently most of our benchmarks are run once with multiple repeats. This means that the stats such as min, max, stddev of the runtime which get reported are not super useful.

This PR splits up the iterations into multiple rounds, resulting in more meaningful output.


vmoens (approved):
LGTM

PR #283, opened 2023-03-17
[BugFix] Fix reshape with non-expanded sizes

vmoens:
## Description

`TensorDict.reshape` calls `TensorDict.view` when the shape is not expanded.

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -2060,7 +2060,7 @@ class TensorDictBase(MutableMapping):
 
         """
         if len(shape) == 0 and size is not None:
-            return self.view(*size)
+            return self.reshape(*size)
         elif len(shape) == 1 and isinstance(shape[0], (list, tuple, torch.Size)):
             return self.reshape(*shape[0])
         elif not isinstance(shape, torch.Size):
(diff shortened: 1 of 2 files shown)


test/test_tensordict.py
```diff
@@ -1178,6 +1178,21 @@ def test_pad(self, td_name, device):
         with pytest.raises(RuntimeError):
             pad(td, [0])
 
+    def test_reshape(self, td_name, device):
+        td = getattr(self, td_name)(device)
+        td_reshape = td.reshape(td.shape)
+        assert isinstance(td_reshape, TensorDict)
+        assert td_reshape.shape.numel() == td.shape.numel()
```
tcbegley, line 1185: Why compare `numel` when in this case we expect `td_reshape.shape == td.shape` no?

test/test_tensordict.py
```diff
@@ -1178,6 +1178,21 @@ def test_pad(self, td_name, device):
...
+        td_reshape = td.reshape(*td.shape)
+        assert isinstance(td_reshape, TensorDict)
+        assert td_reshape.shape.numel() == td.shape.numel()
+        td_reshape = td.reshape(-1)
+        assert isinstance(td_reshape, TensorDict)
+        assert td_reshape.shape.numel() == td.shape.numel()
```
tcbegley, line 1191: We could make this test more specific by doing something like

```
assert td_reshape.shape == torch.Size([td.shape.numel()])
```

tensordict/tensordict.py
```diff
@@ -2062,7 +2062,7 @@ def reshape(
         if len(shape) == 0 and size is not None:
             return self.view(*size)
```
tcbegley, line 2063: Should this also become `reshape`?

tcbegley (approved):
LGTM, a few comments inline.

PR #288, opened 2023-03-20
[BugFix] Fix memmap ownership to make it process-wise and allow for indexed memmap persistance

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTION](https://github.com/pytorch/rl/blob/main/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -36,7 +36,7 @@ EllipsisType = type(Ellipsis)
 
 MEMMAP_HANDLED_FN: dict[Callable, Callable] = {}
 HAS_OWNERSHIP = {}
-HAD_OWNERSHIP = {}
+# HAD_OWNERSHIP = {}
 TRANSFER_OWNERSHIP = {}
 
 
@@ -169,7 +169,7 @@ class MemmapTensor:
             Path(filename).touch(exist_ok=True)
             self.file = open(filename, mode="r+")
 
-        self.filename = self.file.name
+        self.filename = copy(self.file.name)
         self.file.close()  # we close the file for now, but don't delete it
 
         if isinstance(size[0], (torch.Tensor, MemmapTensor, np.ndarray)):
@@ -287,6 +287,7 @@ class MemmapTensor:
         # memmap_copy.transfer_ownership = False
         memmap_copy._shape_indexed = None
         memmap_copy.file = memmap_tensor.file
+
         return memmap_copy
 
     def __iter__(self):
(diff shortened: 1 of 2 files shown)


tensordict/memmap.py
```diff
@@ -298,13 +303,13 @@ def _init_shape(
...
-        self.transfer_ownership = transfer_ownership
         self.np_shape = tuple(self._shape)
         self._dtype = dtype
         self._ndim = len(shape)
         self._numel = prod(shape)
-        self._has_ownership = True
-        self._had_ownership = True
+        TRANSFER_OWNERSHIP[id(self.file)] = transfer_ownership
+        HAS_OWNERSHIP[id(self.file)] = True
+        HAD_OWNERSHIP[id(self.file)] = True
```
tcbegley, line 312: I'm trying to work out if tying everything to the `id` of the file handle could result in duplication, and whether it's instead better to use a string of the absolute path?

I think we can probably get away with it because we only clean up the file if the user doesn't specify it and we create a temporary file for them. So for example this would be ok, even though the ids of `self.file` would differ

```python
mm1 = MemmapTensor(10, filename="myfile.memmap")
mm2 = MemmapTensor(10, filename="myfile.memmap")
```

tcbegley (approved):
LGTM. I'm a little unsure about using the `id` of the file handles (see inline comment) but I don't believe it will actually cause a problem, because I can't see how the user would end up with two distinct file handles to the same generated temporary file without doing something bizarre.

tensordict/memmap.py
vmoens: Should we use the filename instead?

tcbegley: If it's easy enough to get the absolute filename then I think maybe yes. Just because if we change the behaviour for specified filenames in the future, we could end up with a race condition where one instance of `MemmapTensor` is not aware of another because they have different file handles pointing to the same file on disk.

tensordict/memmap.py
```diff
@@ -537,8 +544,13 @@ def __deepcopy__(self, memo: dict[int, Any] | None = None) -> MemmapTensor:
         return MemmapTensor.from_tensor(self.clone())
 
     def __del__(self) -> None:
-        if "_has_ownership" in self.__dir__() and self._has_ownership:
-            if isinstance(self.file, tempfile._TemporaryFileWrapper):
+        if not hasattr(self, "file"):
+            return
+        # for some reason Memmap keeps 2 refs to the file
+        if HAS_OWNERSHIP.get(self.filename, False) and getrefcount(self.file) <= 2:
```
tcbegley, line 550: I think ref counting the file handle results in the same problem we tried to fix before by switching to filenames, but fixing this is probably a lot harder, so maybe we just leave it for now. As discussed a user would really have to go out of their way to obtain two distinct file handles to the same randomly generated temporary file.

tcbegley (approved):
LGTM

PR #289, opened 2023-03-21
[Feature] h5 compatibility

vmoens:
## Description

Test:
```python
import numpy as np
import h5py
from tensordict import PersistentTensorDict

arr = np.random.randn(1000)

with h5py.File('groups.hdf5', 'w') as f:
    g = f.create_group('Base_Group')
    gg = g.create_group('Sub_Group')

    d = g.create_dataset('default', data=arr)
    dd = gg.create_dataset('default', data=arr)

td = PersistentTensorDict(filename='groups.hdf5', batch_size=[1000])
print(td.get(("Base_Group", "Sub_Group", "default"))[:2])
print(td[:2]["Base_Group", "Sub_Group", "default"])
print(td[:2]["Base_Group", "Sub_Group"]["default"])
print(td["Base_Group"])
print(td)
td = PersistentTensorDict(filename='groups.hdf5', batch_size=[3])
print(td)  # raises an exception
```

Follow up:
- [ ] Test repr
- [ ] Benchmarks
- [ ] kwargs (e.g. `compression="gzip", compression_opts=9`)

### DIFF
--- a/tensordict/persistent.py
+++ b/tensordict/persistent.py
@@ -133,6 +133,7 @@ class PersistentTensorDict(TensorDictBase):
 
         # we use this to allow nested tensordicts to have a different batch-size
         self._nested_tensordicts = {}
+        self._pin_mem = False
 
         # this must be kept last
         self._check_batch_size(self._batch_size)
@@ -225,7 +226,10 @@ class PersistentTensorDict(TensorDictBase):
                 device = torch.device("cpu")
             # we convert to an array first to avoid "Creating a tensor from a list of numpy.ndarrays is extremely slow."
             array = array[:]
-            return torch.as_tensor(array, device=device)
+            out = torch.as_tensor(array, device=device)
+            if self._pin_mem:
+                return out.pin_memory()
+            return out
         elif array is not default:
             out = self._nested_tensordicts.get(key, None)
             if out is None:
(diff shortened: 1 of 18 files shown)


tensordict/tensordict.py
```diff
@@ -4209,12 +4234,48 @@ def __init__(
...
         self._source = source
-        idx = (idx,) if not isinstance(idx, (tuple, list, range)) else tuple(idx)
+        idx = (
+            (idx,)
+            if not isinstance(
+                idx,
+                (
+                    tuple,
+                    list,
+                ),
```
tcbegley, line 4244: Why did `range` get removed here?

tcbegley (commented):
Amazing effort! This is really cool. I left a few comments and questions, but essentially this LGTM!

test/test_h5.py
vmoens, line 1: @giadefa
Does this cover what you were pointing at?

tensordict/persistent.py
```diff
@@ -0,0 +1,840 @@
...
+        state["file"] = None
+        state["filename"] = filename
+        state["group_name"] = group_name
+        return state
+
+    def __setstate__(self, state):
+        state["file"] = h5py.File(state["filename"], mode=state["mode"])
+        if state["group_name"] != "/":
+            state["file"] = state["file"][state["group_name"]]
+        del state["group_name"]
```
tcbegley, line 811: Why delete `group_name` but not `filename`?

tcbegley (approved):
LGTM! Great work

tensordict/persistent.py
vmoens: because we don't have an attribute groupname in the class
but it's not super elegant I admit

tensordict/tensordict.py
vmoens: bc we dont need to convert it to a tuple and it's a bit crazy to do `idx = tuple(range(1_000_000))` no?

tensordict/persistent.py
```diff
@@ -0,0 +1,841 @@
...
+        self.__dict__.update(state)
```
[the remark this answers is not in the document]
giadefa: ok

PR #290, opened 2023-03-22
[Feature] forward getattr to wrapped module

apbard:


tensordict/nn/common.py
```diff
@@ -453,6 +453,12 @@ def __repr__(self) -> str:
 
         return f"{self.__class__.__name__}(\n{fields})"
 
+    def __getattr__(self, name: str) -> Any:
+        try:
+            return super().__getattr__(name)
+        except AttributeError:
+            return getattr(super().__getattr__("module"), name)
```
vmoens, line 460: I think that what will happen here is that a missing attribute will raise an exception for module.
I guess we want both error messages
```suggestion
    def __getattr__(self, name: str) -> Any:
        try:
            return super().__getattr__(name)
        except AttributeError as err1:
            try:
                return getattr(super().__getattr__("module"), name)
            except Exception as err2:
                raise err2 from err1
```
if that makes sense (?)

vmoens (approved):
LGTM!

vmoens (approved):

PR #302, opened 2023-03-24
[Test] consolidate test_tensorclass and test_tensorclass_nofuture

apbard:


test/test_tensorclass.py
tcbegley, line 1: We can get rid of the comments `# future: drop quotes` from this file.

PR #308, opened 2023-03-28
[Refactor] Upgrade vmap imports

vmoens:
## Description

vmap is now loaded from torch and not functorch


test/test_functorch.py
```diff
@@ -22,8 +22,8 @@
 try:
     from functorch import (
         make_functional_with_buffers as functorch_make_functional_with_buffers,
-        vmap,
     )
+    from torch import vmap
```
tcbegley, line 26: When testing on PyTorch<2.0 this will result in an `ImportError` and hence the tests getting skipped. So we'll lose test coverage of older versions where it would have been possible to import `vmap` from `functorch` instead.

tcbegley (changes requested):
Happy with the updates to the examples, but we're losing test coverage on older versions of PyTorch here. It'll be ugly but I think we need to have more complicated import logic that will attempt to import `vmap` from `torch` and fall back on `functorch` if it can't find it.

PR #324, opened 2023-04-06
[CI] run benchmarks on PR

apbard:


.github/workflows/benchmarks_pr.yml
```diff
@@ -0,0 +1,123 @@
+name: Benchmarks
```
vmoens, line 1: Maybe Continuous Benchmarks to match the other workflow?

apbard: I thought they would conflict

vmoens: no you will just see the branch appear (main or the one from the PR)

.github/workflows/benchmarks_pr.yml
```diff
@@ -0,0 +1,59 @@
+name: Continuous Benchmarks
```
vmoens, line 1:
```suggestion
name: Continuous Benchmark
```

vmoens:
They are duplicated now haha

I guess they should be in the same file, is it possible?

apbard:
> They are duplicated now haha
> 
> I guess they should be in the same file, is it possible?

let's start with getting this to work... :)

PR #333, opened 2023-04-13
Introduce InteractoinType Enum

Goldspear:
## Description

Introduced `InteractionType` Enum for better method checking/selecting, with backward compatibility with existing string arg configurations.  

## Motivation and Context

This is to address https://github.com/pytorch/rl/issues/1016

- [x] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

## Checklist

## Test
* Added unit tests
* `pytest test/`

### DIFF
--- a/tensordict/nn/probabilistic.py
+++ b/tensordict/nn/probabilistic.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import re
+from enum import auto, Enum
 from textwrap import indent
 from typing import Any, Callable, Sequence
 from warnings import warn
@@ -22,7 +23,22 @@ from torch.autograd.grad_mode import _DecoratorContextManager
 __all__ = ["ProbabilisticTensorDictModule", "ProbabilisticTensorDictSequential"]
 
 
-_INTERACTION_TYPE = None
+class InteractionType(Enum):
+    MODE = auto()
+    MEDIAN = auto()
+    MEAN = auto()
+    RANDOM = auto()
+
+    @classmethod
+    def from_str(cls, type_str: str) -> InteractionType:
+        """Return the interaction_type with name matched to the provided string (case insensitive)."""
+        for member_type in cls:
+            if member_type.name == type_str.upper():
+                return member_type
+        raise ValueError(f"The provided interaction type {type_str} is unsupported!")
+
+
+_INTERACTION_TYPE: InteractionType | None = None
 
 
 def _insert_interaction_mode_deprecation_warning(
@@ -35,7 +51,7 @@ def _insert_interaction_mode_deprecation_warning(
     )
 
 
-def interaction_type() -> str | None:
+def interaction_type() -> InteractionType | None:
     """Returns the current sampling type."""
     return _INTERACTION_TYPE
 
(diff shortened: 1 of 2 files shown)


tensordict/nn/probabilistic.py
```diff
@@ -56,7 +73,7 @@ class set_interaction_mode(_DecoratorContextManager):
     def __init__(self, mode: str = "mode") -> None:
         _insert_interaction_mode_deprecation_warning("set_")
         super().__init__()
-        self.mode = mode
+        self.mode = InteractionType.from_str(mode)
```
vmoens, line 76: Can I do
```
set_interaction_mode(InteractionType.MODE)
```
instead?
Shouldn't this be the "proper" way to call it?
IMO we should do
```suggestion
        if isinstance(mode, str):
            warning(...)
            self.mode = InteractionType.from_str(mode)
       else:
             self.mode = mode
```

Goldspear: Yeah, I've thought about this, but since we're deprecating this fcn, any new usage should preferably be using the new alternative `set_interaction_type(InteractionType.MODE)` and all the existing usage only passes `str` arg, so didn't think changing the old interface to add this support was necessary. Unless I've missed any use cases where old fcn + new interface is somehow preferred? 🤔

Goldspear:
@vmoens for the two CI failures, not sure whether/how is the error `Error: Workflow failed! Resource not accessible by integration` related to my changes?

vmoens:
> @vmoens for the two CI failures, not sure whether/how is the error `Error: Workflow failed! Resource not accessible by integration` related to my changes?

The GHA CI is not working properly, but there seems to be a more grounded problem with this on CircleCI
https://app.circleci.com/pipelines/github/pytorch-labs/tensordict/1906/workflows/54da6585-d703-4eff-88a8-9cd67d3fdb89/jobs/43965
Can you have a look?

tensordict/nn/probabilistic.py
vmoens: > Yeah, I've thought about this, but since we're deprecating this fcn, any new usage should preferably be using the new alternative `set_interaction_type(InteractionType.MODE)` and all the existing usage only passes `str` arg, so didn't think changing the old interface to add this support was necessary. Unless I've missed any use cases where old fcn + new interface is somehow preferred? 🤔

Oh got it, I missed that
So IIUC with `set_interaction_type` you can only pass `Enum` and with `set_interaction_mode` you can pass strings (but you'll get a warning), right?
Does the warning tells you that when changing your code to `set_interaction_mode` you should also use `Enum`?

Goldspear: > So IIUC with set_interaction_type you can only pass Enum and with set_interaction_mode you can pass strings (but you'll get a warning), right?

Yes, that's right!

> Does the warning tells you that when changing your code to set_interaction_mode you should also use Enum?

The new `set_interaction_type` and `default_interaction_type` are type-hinted to `InteractionType` enum only, so linting should pick it up if one tries to feed a str to them. But I will add a note to the warning message to highlight the type change as well 👍

vmoens (approved):
LGTM!

PR #349, opened 2023-04-19
[Feature] add selection/filter of in and out keys to TDSequenceModules

apbard:
This PR add `in_keys` and `out_keys` as arguments to TensorDictSequential to specify what are the expected input and outputs.

this will:
- Improve the behaviour of wrapped modules when used with tensors as it will output only the specified tensors.

  Before:
  ```
  Seq(
    Mod(lambda :x, [“x”], [“intermediate”]),
    Mod(lambda :x, [“intermediate”], [“intermediate2”]),
    Mod(lambda :x, [“intermediate2”], [“z”]),
  )
  
  _, _, z = Seq(obs)
  ```
  Now:
  ```
  Seq(
    Mod(lambda :x, [“x”], [“intermediate”]),
    Mod(lambda :x, [“intermediate”], [“intermediate2”]),
    Mod(lambda :x, [“intermediate2”], [“z”]),
    out_keys = "z"
  )
  
  z = Seq(obs)
  ```

- potentially optimize the sequence removing redundant or no longer used modules:
  ```
  Seq(
    Mod(lambda :x, [“x”], [“y”]),
    Mod(lambda :x, [“y”], [“z”]),
    Mod(lambda :x, [“y”], [“k”]),
    out_keys = "k"
  )
  ```
  can be optimized into
  ```
  Seq(
    Mod(lambda :x, [“x”], [“y”]),
    Mod(lambda :x, [“y”], [“k”]),
    out_keys = "k"
  )
  ```
- this enables the tensordict passed through the network to be further optimized by avoiding storing not needed temporary intermediate keys. This could save a lot of memory if several tensordictmodules are stacked. This will complement and extend the usage of "_" or "" when:
    - using third-party modules
    - when the visibility of the intermediate key is needed only up to a certain point in the network
  
This PR should not alter default behaviour except with some specific combinations of subselection and dispatch use cases.

### DIFF
--- a/tensordict/nn/sequence.py
+++ b/tensordict/nn/sequence.py
@@ -30,6 +30,55 @@ from torch import nn
 __all__ = ["TensorDictSequential"]
 
 
+def _get_subsequence(module_list, in_keys, out_keys):
+    """ Selects the minimal sequence of modules that take at least in_keys as input and out_keys as output"""
+    in_keys = deepcopy(in_keys)
+    out_keys = deepcopy(out_keys)
+    id_to_keep = set(range(len(module_list)))
+    for i, module in enumerate(module_list):
+        # check if the modules uses some of the in_keys needed so far
+        if (type(module) is TensorDictSequential):  
+            # no isinstance because we don't want to mess up subclasses
+            try:
+                module = module_list[i] = module.select_subsequence(in_keys=in_keys)
+                in_keys.extend(module.out_keys)
+            except ValueError as e:
+                # then the module can be removed
+                id_to_keep.remove(i)
+                continue
+        elif set(in_keys) & set(module.in_keys):
+            in_keys.extend(module.out_keys)
+        else:
+            # then the module can be removed
+            id_to_keep.remove(i)
+
+    for i, module in reversed(list(enumerate(module_list))):
+        if i in id_to_keep:
+            # if the model is producing some of the out_keys needed so far
+            if (type(module) is TensorDictSequential):  
+                # no isinstance because we don't want to mess up subclasses
+                try:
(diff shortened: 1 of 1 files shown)


vmoens:
I think those changes are not necessary for the following reasons:
- the first usage is already covered by dispatch, and this will work with any TensorDictModuleBase subclass (not only Sequential):
```
model = Seq(...)
model = dispatch(model, dest=["theKeyIWant"])
# you can check the keys here
model.dest
val = model(tensor)
```

- I'm a bit puzzled by the second example: why would someone give a model with unused modules?
What does it add compared to the current select_subsequence?

- in grad_mode(True) you won't save any memory since the graph is registered. In other cases we can use "_" and filter it out if needed. I also don't see why we would save memory since you can't ditch tensors in advance:

Seq(
  Mod(lambda x:x, [“x”], [“y”]),
  Mod(lambda x:x, [“y”], [“z”]),
  Mod(lambda x:x, [“z”], [“a”]),
  Mod(lambda x, y: x+y, [“x”, “a”], [“k”]),
)
```

in this case, even if we say we only want “k” we must store every intermediate value except z until the end. Deleting them on the fly can only be done if sequential is clever enough to know if an entry is accessed down the line. Again I think this can be already be done by using the same “_” key for values that can be overwritten, in a less automated way I agree.

If your sequence is very granular and you save every intermediate value then yes the men could increase at inference time but I don't think it is the proper way to use the class.

vmoens:
Cc'ing people for comments / thoughts
@shagunsodhani @tcbegley @matteobettini @albertbou95

shagunsodhani:
> Cc'ing people for comments / thoughts @shagunsodhani @tcbegley @matteobettini @albertbou92 @BY571

Agreed with your points @vmoens

PR #356, opened 2023-04-25
[Feature] Named dims

vmoens:
## Description

Introduces optional named dimensions

Helps with features such as https://github.com/pytorch/rl/pull/1080

Covers
- [x] indexing
- [x] cat
- [x] stack
- [x] squeeze
- [x] unsqueeze
- [x] clone
- [x] permute
- [x] expand
- [x] sub-tds
- [x] flatten
- [x] apply
- [x] gather
- [x] masked_fill
- [x] select
- [x] memmap_like
- [x] detach
- [x] unbind
- [x] split
- [x] any / all with dim
- [x] set and set_

Reshape and view loose the names as expected

TODO
- [ ] transpose => not implemented yet
- [ ] scatter => not implemented yet
- [ ] where => not implemented yet

### DIFF
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -92,6 +92,19 @@ To reshape the batch dimensions one can do
 
 The class supports many other operations, including squeeze, unsqueeze, view, permute, unbind, stack, cat and many more. If an operation is not present, the TensorDict.apply method will usually provide the solution that was needed.
 
+Named dimensions
+----------------
+
+TensorDict and related classes also support dimension names.
+The names can be given at construction time or refined later. The semantic is
+similar to the torch.Tensor dimension name feature:
+
+>>> tensordict = TensorDict({}, batch_size=[3, 4], names=["a", None])
+>>> tensordict.refine_names(..., "b")
+>>> tensordict.names = ["z", "y"]
+>>> tensordict.rename("m", "n")
+>>> tensordict.rename(m="h")
+
 Nested TensorDicts
 ------------------
 


vmoens:
> ## ⚠
> ⚠
> Warning
>  Result of CPU Benchmark Tests
> ### Total Benchmarks: 46. Improved: 
> 14
> . Worsened: 
> 2
> .

Seriously?

tensordict/tensordict.py
```diff
@@ -360,6 +361,132 @@ def batch_size(self) -> torch.Size:
...
+        greedily; it is expanded in-place to fill names to the same length as
+        self.dim() using names from the corresponding indices of self.names.
+
+        Returns: the tensordict with dimensions named accordingly.
+
+        """
+        # replace ellipsis if any
+        names_copy = copy(names)
+        if any(name is Ellipsis for name in names):
+            ellipsis_name = [NO_DEFAULT for _ in self.ndim - len(names)]
```
tcbegley, line 412: This doesn't look right. At the least it's missing `range`, but also I think there's an off by one error (need to correct for the fact that we're removing `Ellipsis` which is counted in `len(names)`)

tensordict/tensordict.py
```diff
@@ -360,6 +361,132 @@ def batch_size(self) -> torch.Size:
...
+        self._rename_subtds(self.names)
+        return self
+
+    def rename(self, *names, **rename_map):
+        clone = self.clone(recurse=False)
+        if len(names) == 1 and names[0] is None:
+            clone.names = None
+        if rename_map and names:
+            raise ValueError(
+                "Passed both a name map and a name list. " "Only one is accepted."
```
tcbegley, line 443:
```suggestion
                "Passed both a name map and a name list. Only one is accepted."
```

tcbegley (commented):
This looks like a good start.

I think we should add a test for `refine_names` as there's no coverage there atm.

Also should names be propagated to tensors in the tensor dict? I'm seeing for example

```python
import torch

td = TensorDict({"a": torch.rand(10)}, [10], names=["my_dim"])
assert td.names == ["my_dim"]
assert td["a"].names == (None,)
```

tensordict/tensordict.py
```diff
@@ -360,6 +361,132 @@ def batch_size(self) -> torch.Size:
...
+        # replace ellipsis if any
+        names_copy = copy(names)
+        if any(name is Ellipsis for name in names):
+            ellipsis_name = [NO_DEFAULT for _ in self.ndim - len(names)]
+            names = []
+            for name in names_copy:
+                if name is Ellipsis:
+                    names += ellipsis_name
+                else:
+                    names.append(name)
```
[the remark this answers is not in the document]
vmoens: Can you have more than one ellipsis?
[..., 1, ...] will not work
[1, ..., ...] is dubious
[..., ..., 1] also
[..., 1, ..., 1] 
others?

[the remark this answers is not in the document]
tcbegley: I don't think there would be a legitimate usage, so I guess maybe fine to leave it on the user to provide reasonable input rather than to do additional runtime checks that might slow things down 👍

tensordict/tensordict.py
vmoens: Edited!

PR #376, opened 2023-05-07
[Performance] Faster indexing

vmoens:


vmoens:
Using Nd boolean masks presented as lists now fails but I think it is because this is supported by tensordict and not by tensors anymore. 
I think this feature can be dropped for the next release.

PR #381, opened 2023-05-16
[Refactor] Refactor to()

vmoens:
@tcbegley I wanted to refactor to() to use apply but it breaks. Can you have a look? Not 100% sure those changes make sense anyway...


vmoens:
> @tcbegley I wanted to refactor to() to use apply but it breaks. Can you have a look? Not 100% sure those changes make sense anyway...

My mistake, the failing tests aren't really failing (except for lint)

tcbegley (approved):
Cool, in which case this LGTM!

Issue #382, opened 2023-05-17
[BUG] Error in sequential indexing of TensorDict with MemmapTensor


teopir:
## Describe the bug

Sampling two dimensions in a consecutive way produces an error when using `MemmapTensor`. If a standard `Tensor` is used, no error is raised.

## To Reproduce

Steps to reproduce the behavior.

```python
import tensordict, numpy, sys, torch
from tensordict import TensorDict
torch.manual_seed(4584371022706121143)

A, B = 10, 2
SAMPLE_SIZE = 1
tensordict = TensorDict({"a": torch.rand(A, B, 2), "b": torch.rand(A, B, 1)}, [A, B])
print("b: ", tensordict["b"])
tensordict.memmap_()
print(tensordict)

# sample first dimension
idx = torch.randint(0, A, (SAMPLE_SIZE,))
tensordict1 = tensordict[idx]
print(tensordict1)

# sample second dimension
step_idx = torch.randint(0, B, (SAMPLE_SIZE,))
tensordict2 = tensordict1[range(SAMPLE_SIZE), step_idx]
print(tensordict2)

tensordict2['b'] * 0.9
```

Output and stack traces

```bash
b:  tensor([[[0.4750],
         [0.1718]],

        [[0.0098],
         [0.7934]],

        [[0.2763],
         [0.5519]],

        [[0.2940],
         [0.8162]],

        [[0.3873],
         [0.9035]],

        [[0.6649],
         [0.1111]],

        [[0.4237],
         [0.0228]],

        [[0.2010],
         [0.4994]],

        [[0.4820],
         [0.4220]],

        [[0.0147],
         [0.4110]]])
TensorDict(
    fields={
        a: MemmapTensor(shape=torch.Size([10, 2, 2]), device=cpu, dtype=torch.float32, is_shared=False),
        b: MemmapTensor(shape=torch.Size([10, 2, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([10, 2]),
    device=None,
    is_shared=False)
TensorDict(
    fields={
        a: MemmapTensor(shape=torch.Size([1, 2, 2]), device=cpu, dtype=torch.float32, is_shared=False),
        b: MemmapTensor(shape=torch.Size([1, 2, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([1, 2]),
    device=None,
    is_shared=False)
TensorDict(
    fields={
[...]

PR #383, opened 2023-05-17
[Bugfix] Indexing memmaps with torch.Tensor

tcbegley:
## Description

This PR fixes a bug that arises when indexing `MemmapTensor` with a `torch.Tensor`. I found this investigating the bug reported in #382 which these changes fix.

The basic problem is that indexing a `numpy.memmap` with a `torch.Tensor` behaves differently to indexing with a `numpy.ndarray` or a list. For example

```python
>>> import numpy as np
>>> import torch
>>> from tensordict import MemmapTensor

>>> t = torch.rand(3, 2, 1)
>>> mm = MemmapTensor.from_tensor(t)
>>> idx_list, idx_np, idx_tch = [0], np.array([0]), torch.tensor([0])

>>> print(t[idx_list].shape, t[idx_np].shape, t[idx_tch].shape)
torch.Size([1, 2, 1]) torch.Size([1, 2, 1]) torch.Size([1, 2, 1])

>>> print(mm[idx_list].shape, mm[idx_np].shape, mm[idx_tch].shape)
torch.Size([1, 2, 1]) torch.Size([1, 2, 1]) torch.Size([1, 2, 1])

>>> mm_np = mm._memmap_array  # underlying memmap behaves differently
>>> print(mm_np[idx_list].shape, mm_np[idx_np].shape, mm_np[idx_tch].shape)
(1, 2, 1) (1, 2, 1) (2, 1)
```

Notice that when indexing the `numpy.memmap` in the final example, we lose the singleton first dimension, resulting in a different shape.

This mismatch in the shape that `MemmapTensor` expects and the actual shape of the underlying memmap leads to indexing errors such as that reported in the issue.

The solution I've adopted here is to simply convert any tensors used to index a `MemmapTensor` to `np.ndarray` before indexing. This results in indexing behaviour that is consistent with indexing a `torch.Tensor` with another `torch.Tensor`

### DIFF
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -362,14 +362,27 @@ class MemmapTensor:
 
     def _get_item(self, idx: IndexType, memmap_array: np.ndarray) -> np.ndarray:
         if isinstance(idx, torch.Tensor):
-            idx = idx.cpu()
+            # indexing a numpy.memmap with a torch.Tensor doesn't behave as expected, we
+            # convert to numpy.ndarray for behaviour that is consistent with indexing
+            # a torch.Tensor with a torch.Tensor
+            idx = idx.cpu().numpy()
         elif isinstance(idx, tuple) and any(
             isinstance(sub_index, torch.Tensor) for sub_index in idx
         ):
             idx = tuple(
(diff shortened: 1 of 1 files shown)


vmoens (approved):
LGTM thanks for this tom!

PR #384, opened 2023-05-17
[Bugfix] fix getitem

apbard:
getitem is not working correctly paired with select_out_keys as the input is not a tuple but a list:
```python
mod = TensorDictModule(
    lambda x, y: (x + 2, y + 2, x), in_keys=["a", "b"], out_keys=["c", "d", "e"]
)
print(mod.out_keys)
mod.select_out_keys(["c", "d", "e"])
print(mod.out_keys)
mod.reset_out_keys()
print(mod.out_keys)
```
Results in:
```
['c', 'd', 'e']
[['c', 'd', 'e']]
['c', 'd', 'e']
```
then when dispatched [this](https://github.com/pytorch-labs/tensordict/blob/main/tensordict/nn/common.py#L271) raises

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -517,6 +517,16 @@ class TensorDictModuleBase(nn.Module):
                 is_shared=False)
 
         """
+        if len(out_keys) == 1:
+            if out_keys[0] not in self.out_keys:
+                err_msg = f"Can't select non existent key: {out_keys[0]}. "
+                if (
+                    out_keys[0]
+                    and isinstance(out_keys[0], (tuple, list))
+                    and out_keys[0][0] in self.out_keys
+                ):
+                    err_msg += f"Are you passing the keys in a list? Try unpacking as: `{', '.join(out_keys[0])}`"
+                raise ValueError(err_msg)
         self.register_forward_hook(_OutKeysSelect(out_keys))
         for hook in self._forward_hooks.values():
             hook._init(self)
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -2025,105 +2025,147 @@ class TestSkipExisting:
         assert (td["out"] == 1).all()
 
 
+@pytest.mark.parametrize("out_d_key", [("d", "e"), ["d"], ["d", "e"]])
+@pytest.mark.parametrize("unpack", [True, False])
 class TestSelectOutKeys:
-    def test_tdmodule(self):
+    def test_tdmodule(self, out_d_key, unpack):
         mod = TensorDictModule(
-            lambda x, y: (x + 2, y + 2), in_keys=["a", "b"], out_keys=["c", "d"]
+            lambda x, y: (x + 2, y + 2, x), in_keys=["a", "b"], out_keys=["c", "d", "e"]
         )
+        assert mod.out_keys == ["c", "d", "e"]
         td = mod(TensorDict({"a": torch.zeros(()), "b": torch.ones(())}, []))
-        assert all(key in td.keys() for key in ["a", "b", "c", "d"])
-        mod2 = mod.select_out_keys("d")
-        assert mod2 is mod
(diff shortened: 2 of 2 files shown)


vmoens:
Do you have a reprod example?

tensordict/tensordict.py
```diff
@@ -3010,10 +3010,11 @@ def __getitem__(self, idx: IndexType) -> TensorDictBase:
             >>> print(td.get("a"))  # values have not changed
 
         """
-        if isinstance(idx, tuple) and len(idx) == 1:
+        if isinstance(idx, (list, tuple)) and len(idx) == 1:
```
vmoens, line 3013: i don't think tuples and lists should be semantically equivalent
```python
>>> import torch
x = torc>>> x = torch.randn(3, 4)
>>> x[0]
tensor([ 0.3932,  0.2631,  0.4270, -0.1543])
>>> x[[0]]
tensor([[ 0.3932,  0.2631,  0.4270, -0.1543]])
>>> x[(0,)]
tensor([ 0.3932,  0.2631,  0.4270, -0.1543])
```

tensordict/tensordict.py
```diff
@@ -3010,10 +3010,11 @@ def __getitem__(self, idx: IndexType) -> TensorDictBase:
...
             idx = idx[0]
         if isinstance(idx, str) or (
-            isinstance(idx, tuple) and all(isinstance(sub_idx, str) for sub_idx in idx)
+            isinstance(idx, (list, tuple))
```
vmoens, line 3016: ditto

tensordict/nn/common.py
```diff
@@ -517,6 +517,8 @@ def select_out_keys(self, *out_keys):
                 is_shared=False)
 
         """
+        if len(out_keys) == 1:
```
vmoens, line 520: I don't think we need to support this: the function signature and docstrings are pretty clear that a sequence of keys is expected, not a list/tuple of keys. It becomes confusing IMO since we support nested keys. 
Eg
```
TensorDict({
 "a": {"b": foo},
 "b": bar}, [])
```
What is `["a", "b"]` in this context?
To me it should only be the nested entry (`"foo"`)

tensordict/nn/common.py
```diff
@@ -517,6 +517,16 @@ def select_out_keys(self, *out_keys):
...
+            if out_keys[0] not in self.out_keys:
+                err_msg = f"Can't select non existent key: {out_keys[0]}. "
+                try:
+                    if out_keys[0][0] in self.out_keys:
+                        err_msg += f"Are you passing the keys in a list? Try unpacking as: `{', '.join(out_keys[0])}`"
+                except Exception:
+                    pass
```
vmoens, line 527: what does this do?

apbard: if you pass a list with just one element it checks if that element is a key. otherwise it prints a meaningful message suggesting a possible solution.
that is do not call mod.select_out_keys(["a","b","c"]) but mod.select_out_keys("a","b","c")

vmoens: Sure! I was wondering why the nested try/except. Can we get rid of the inner one? Why should it fail?
I'm just afraid that if something else is wrong (eg self.out_keys is void or smth strange like that) we display the wrong error message because every exception is skipped.

apbard: I have done this just to be concise.
do you think something like `if isinstance(out_keys[0], (tuple,list)) and out_keys[0] and out_keys[0][0] in self.out_keys:` would be better?

vmoens: this is not superb I must admit
we maybe we can put the raise directly after exept and duplicate it after without the `from`

apbard: we are not interested in other exceptions. it's just to provide an hint on the solution. switched to the if version

vmoens (approved):

PR #389, opened 2023-05-19
[Refactor] Faster creation on device

vmoens:
We call `to(..., non_blocking=True)` in TD creation to make things faster

Before
```python
%%timeit
td = TensorDict({s: torch.zeros(100, 100) for s in string.ascii_lowercase}, [], device="cuda:0")
```
1.16 ms ± 279 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

After
783 µs ± 167 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

Conversely, before:
```python
%%timeit
td = TensorDict({s: torch.zeros(100, 100, device="cuda:0") for s in string.ascii_lowercase}, [], device="cpu")
```
1.25 ms ± 170 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

after
613 µs ± 27 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

For the record:
I tested pin_memory but it does not help

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1545,9 +1545,6 @@ class TensorDictBase(MutableMapping):
                     f" numeric scalars and tensors. Got {type(value)}"
                 )
 
-        if self.device is not None:
-            value = value.to(self.device)
-
         if check_shape and _shape(value)[: self.batch_dims] != self.batch_size:
             # if TensorDict, let's try to map it to the desired shape
             if is_tensor_collection(value):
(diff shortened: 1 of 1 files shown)


tcbegley (commented):
Cool! I'm just wondering if there is any situation in which moving the value onto the device would in fact be blocking? Is it possible we create unexpected behaviour for the user here?

I also am confused why the improvements you measured aren't being reflected in the benchmarks? Looks like none were reported as having improved. Maybe none of the benchmarks are capturing this change?

vmoens:
This is due to the problem I mentioned yesterday that no benchmark is actually running on gpu :( 
I tried a couple of things [there](#388) without success.
If you want to have a look ;)

vmoens:
In general I think it's safe
The only thing that this does it to let us load all the data asynchronously on device, but to access it you'll have to wait for it to be there...

tcbegley (approved):
Cool, LGTM then!

PR #395, opened 2023-05-25
[Feature] Registering a tensordict as a module buffer

vmoens:
If a tensordict is registered as a module buffer, it must accept a bunch of calls like `double`, `float` etc.

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -2226,22 +2226,23 @@ class TensorDictBase(MutableMapping):
         return tuple(self[(*[slice(None) for _ in range(dim)], idx)] for idx in indices)
 
     def clone(self, recurse: bool = True) -> TensorDictBase:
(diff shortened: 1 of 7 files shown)


tensordict/tensordict.py
```diff
@@ -3248,6 +3248,46 @@ def unlock_(self) -> TensorDictBase:
 
     unlock = _renamed_inplace_method(unlock_)
 
+    def is_floating_point(self):
+        for item in self.values(True, True):
```
apbard, line 3252: why don't we use kwargs for improved readability?

tcbegley (approved):
LGTM

apbard (approved):
LGTM, just added minor

tensordict/tensordict.py
vmoens: fixed

PR #436, opened 2023-06-20
[Feature] Prevent lock/unlock conflicts for stacked and nested tensordicts

vmoens:
Introduces a lock tracking mechanism across tensordicts in a graph to tie them together such that once a root tensordict is locked, no sub-tensordict (either from a stack or a nested tensordict) can be unlocked and modified.
The use case we want to prevent is this
```python
td = TensorDict({("a", "b", "c", "d"): 1.0}, [])
a = td["a"]
b = td["a", "b"]
c = td["a", "b", "c"]
td.lock_()
a.unlock_()
a["b"] = 1.0
a.lock_()
```
In this case, one should not be allowed to modify "a" as it is locked by "td".

This means that some edge cases now need to be considered:

```python
td = TensorDict({("a", "b", "c", "d"): 1.0}, [])
other_td = TensorDict({"a": td["a"]}, [])
td.lock_()
other_td.lock_()
td.unlock_() # raises an exception
```
This new exception comes from the fact that "a" and the other sub-tesnsordicts are now included in two separated graphs so we can't unlock them.
To unlock this, there are 2 solutions:
```python
del other_td
td.unlock_() # works
```
or
```python
super_td = TensorDict({"td": td, "other_td": other_td}, [])
super_td.unlock_() # works
```

We include some new tests to check that everything works as expected for tensordict and lazy stacks.

### DIFF
--- a/tensordict/persistent.py
+++ b/tensordict/persistent.py
@@ -135,6 +135,8 @@ class PersistentTensorDict(TensorDictBase):
         device=None,
         **kwargs,
     ):
+        self._locked_tensordicts = []
+        self._lock_id = set()
         if not _has_h5:
             raise ModuleNotFoundError("Could not load h5py.") from H5_ERR
         super().__init__()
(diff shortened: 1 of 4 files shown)


apbard (approved):
LGTM

PR #445, opened 2023-06-21
[Performance] Refactor get

vmoens:


tensordict/tensordict.py
```diff
@@ -7008,31 +7000,22 @@ def _change_batch_size(self, new_size: torch.Size) -> None:
...
+        if tensor is default:
+            return tensor
+        print("tensor", tensor, key, default, self._source)
+        return self._transform_value(tensor)
+
+    def _get_tuple(self, key, default):
+        tensor = self._source._get_tuple(key, default)
+        if tensor is default:
+            return tensor
+        print("tensor", tensor, key, default, self._source)
```
skandermoalla, line 7014: @vmoens These prints may have been forgotten (?)

vmoens: Yep sorry about that
I have to fix a couple of things more, seems like some edge cases where not accounted for by the tears within tensordict

PR #473, opened 2023-07-03
[Refactor] unravel_key fixes

vmoens:
cc @matteobettini

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -13,7 +13,7 @@ from typing import Any, Callable, Dict, Iterable, List, Sequence, Tuple, Union
 
 import torch
 from cloudpickle import dumps as cloudpickle_dumps, loads as cloudpickle_loads
-from tensordict._tensordict import unravel_keys
+from tensordict._tensordict import unravel_key_list, unravel_keys
 
 from tensordict.nn.functional_modules import make_functional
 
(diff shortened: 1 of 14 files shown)


vmoens:
interestingly it does not work yet but the benchmarks are better!

tensordict/persistent.py
```diff
@@ -228,10 +228,7 @@ def close(self):
 
     def _process_key(self, key):
         key = unravel_keys(key)
```
matteobettini, line 230: one more general thought: shouldn't `unravel_keys` be called `unravel_key`?
cause it always acts on a single key.
especially cause now we have `unravel_key_list` for multiple keys

matteobettini (approved):
LGTM

tensordict/nn/common.py
```diff
@@ -1087,6 +1090,10 @@ def forward(
                 else:
                     raise err
             if isinstance(tensors, (dict, TensorDictBase)):
+                if isinstance(tensors, dict):
+                    keys = unravel_key_list(list(tensors.keys()))
```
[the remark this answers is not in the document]
vmoens: we use multiple dispatch to make it work with tuples and lists (cf c++ function)

But it's also a bit dangerous, as one may quickly end up mixing a tuple of tuples as being a single key or a tuple of tuple as being a list of keys...

tensordict/persistent.py
vmoens: yep I agree :) let me do that

vmoens:
Some benchmarks are faster
Almost all stacked benchmarks are slower but that's the price to pay for the bugfix
I'm more worried about things like `test_getitem[slice_int]` which should not be impacted. I suspect that the `try / except` we add in `__getitem__` is slow when the exception is reached. The alternative is to check the content of the index and then call the unravel function but that means going through the index twice...

PR #474, opened 2023-07-04
[Feature] unravel_key_list

vmoens:
cc @matteobettini
I temporarily keep unravel_keys AND unravel_key, but the first is unchanged. That way we don't need to adopt anything and unravel_key and unravel_key_list are dormant.
That leaves us the possibility of adapting torchrl and adopt the two new functions, then once that is done we can merge #473 and deprecate/change unravel_keys

### DIFF
--- a/tensordict/__init__.py
+++ b/tensordict/__init__.py
@@ -26,6 +26,7 @@ try:
 except ImportError:
     __version__ = None
 
+from tensordict._tensordict import unravel_key, unravel_key_list
 
 __all__ = [
     "LazyStackedTensorDict",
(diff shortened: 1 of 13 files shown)


matteobettini (approved):
LGTM

matteobettini:
I imagine that torchrl will also need a lot of changes when #473 is merged. How do you usually do this? you prepare a corresponding pr in torchrl and merge at the same time?

vmoens:
> I imagine that torchrl will also need a lot of changes when #473 is merged. How do you usually do this? you prepare a corresponding pr in torchrl and merge at the same time?

the goal is that tests always pass on both end, hence this PR
I think that we can make the transition smoothly on both ends

PR #476, opened 2023-07-05
[Feature] Add reset_parameters support to TensorDictModuleBase

smorad:
## Description

This allows us to call `reset_parameters()` on a `TensorDictModuleBase` like one would with an `nn.Module`. cc @vmoens or @matteobettini

## Motivation and Context

I am trying to implement ensembles in TorchRL (https://github.com/pytorch/rl/issues/1344) and being able to call `reset_parameters` would make things easier. An added bonus is that now you can reinitialize the parameters for any `TensorDictModule` which seems like it could be handy.

- [ oops ] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

## Checklist

- [ ] I have updated the documentation accordingly. (How do I do this?)

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -771,6 +771,28 @@ class TensorDictModuleBase(nn.Module):
                 del self._forward_hooks[i]
         return self
 
+    def reset_parameters(self):
+        """Recursively reset the parameters of the module and its children.
+        
+        Examples:
+        >>> from tensordict.nn import TensorDictModule
+        >>> from torch import nn
+        >>> net = nn.Sequential(nn.Linear(2,3), nn.ReLU())
+        >>> old_param = net[0].weight.clone()
+        >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
+        >>> module.reset_parameters()
+        >>> (old_param == net[0].weight).all()
+        tensor(False)
+        """
+        [self._reset_parameters(m) for m in self.children()]
+
+    def _reset_parameters(self, module: Union[nn.Module, TensorDictModuleBase]):
+        if isinstance(module, Union[TensorDictModuleBase, nn.Module]):
+            if hasattr(module, "reset_parameters"):
+                module.reset_parameters()
+            else:
+                [self._reset_parameters(m) for m in module.children()]
+
 
 class TensorDictModule(TensorDictModuleBase):
     """A TensorDictModule, is a python wrapper around a :obj:`nn.Module` that reads and writes to a TensorDict.
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -792,7 +792,8 @@ class TensorDictModuleBase(nn.Module):
             if hasattr(module, "reset_parameters"):
                 module.reset_parameters()
             else:
-                [self._reset_parameters(m) for m in module.children()]
+                for m in module.children():
+                    self._reset_parameters(m)
 
 
 class TensorDictModule(TensorDictModuleBase):
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -788,7 +788,7 @@ class TensorDictModuleBase(nn.Module):
             self._reset_parameters(m)
 
     def _reset_parameters(self, module: Union[nn.Module, TensorDictModuleBase]):
-        if isinstance(module, Union[TensorDictModuleBase, nn.Module]):
+        if isinstance(module, nn.Module):
             if hasattr(module, "reset_parameters"):
                 module.reset_parameters()
             else:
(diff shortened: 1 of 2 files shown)


tensordict/nn/common.py
```diff
@@ -771,6 +771,28 @@ def reset_out_keys(self):
...
+        >>> from tensordict.nn import TensorDictModule
+        >>> from torch import nn
+        >>> net = nn.Sequential(nn.Linear(2,3), nn.ReLU())
+        >>> old_param = net[0].weight.clone()
+        >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
+        >>> module.reset_parameters()
+        >>> (old_param == net[0].weight).all()
+        tensor(False)
+        """
+        [self._reset_parameters(m) for m in self.children()]
```
vmoens, line 787:
```suggestion
        for m in self.children():
            self._reset_parameters(m)
```

tensordict/nn/common.py
```diff
@@ -771,6 +771,28 @@ def reset_out_keys(self):
...
+
+    def _reset_parameters(self, module: Union[nn.Module, TensorDictModuleBase]):
+        if isinstance(module, Union[TensorDictModuleBase, nn.Module]):
+            if hasattr(module, "reset_parameters"):
+                module.reset_parameters()
+            else:
+                [self._reset_parameters(m) for m in module.children()]
```
vmoens, line 794:
```suggestion
                for m in module.children():
                    self._reset_parameters(m)
```

tensordict/nn/common.py
```diff
@@ -771,6 +771,28 @@ def reset_out_keys(self):
...
+        >>> old_param = net[0].weight.clone()
+        >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
+        >>> module.reset_parameters()
+        >>> (old_param == net[0].weight).all()
+        tensor(False)
+        """
+        [self._reset_parameters(m) for m in self.children()]
+
+    def _reset_parameters(self, module: Union[nn.Module, TensorDictModuleBase]):
+        if isinstance(module, Union[TensorDictModuleBase, nn.Module]):
```
vmoens, line 790: TensorDictModuleBase is an nn.Module

```suggestion
        if isinstance(module, nn.Module):
```

vmoens (commented):
Great stuff!
Thanks for this
Just a couple of comments.

Also: since we often use this functionally, it would be neat to have a way to reset tensordict params:
```python
module = TensorDictModule(...)
params = TensorDict.from_module(module)
module.reset_parameters(params) # does the modification of params
```

Then we could vmap through that (provided that vmap supports random number generation)

```python
params = TensorDict.from_module(module).expand(2).clone()
vmap(module.reset_parameters)(params) # does the modification of params for the 2 networks
```
This is cleaner than creating the network twice and then concatenating the params IMO

(no need to include this in this PR 🙃 )

smorad:
That does sound nicer. Is there a good way to make `reset_parameters` functional? E.g. `nn.Linear.reset_params` will do something like `nn.init.xavier_uniform(self.weight)`. We'd need to do something like
```python
old_param = getattr(module, 'weight')
setattr(module, 'weight') = lambda: td['weight']
module.reset_parameters()
setattr(module, 'weight') = old_param
``` 
which seems kinda dangerous. I suspect doing this will break pytorch somewhere.

vmoens (approved):

vmoens:
> That does sound nicer. Is there a good way to make `reset_parameters` functional? E.g. `nn.Linear.reset_parameters` will do something like `nn.init.xavier_uniform(self.weight)`. We'd need to do something like
> 
> ```python
> old_param = getattr(module, 'weight')
> setattr(module, 'weight') = lambda: td['weight']
> module.reset_parameters()
> setattr(module, 'weight') = old_param
> ```
> 
> which seems kinda dangerous. I suspect doing this will break pytorch somewhere.

That's defo for a separate PR but here's how I would approach it:
```python
from tensordict.nn import make_functional, TensorDictModuleBase, TensorDictModule, repopulate_module
from tensordict.nn.functional_modules import extract_weights_and_buffers
from torch import nn
import torch

mod = nn.Sequential(nn.Linear(3, 4))

params = make_functional(mod)

def reset_parameters(mod, params):
    repopulate_module(mod, params.apply(lambda x: nn.Parameter(x.detach().clone().requires_grad_())))
    mod.apply(lambda mod: mod.reset_parameters() if hasattr(mod, "reset_parameters") else None)
    params = extract_weights_and_buffers(mod)
    return params

params2 = reset_parameters(mod, params)
assert not (params == params2).any()
```

PR #478, opened 2023-07-05
Functional reset_parameters

smorad:
## Description

This adds support for `TensorDictModuleBase.reset_parameters` when the parameters exist as in a tensordict, rather than being attached to the module. It also simplifies the existing `reset_parameters` implementation. cc @matteobettini @vmoens 

## Motivation and Context

Related to https://github.com/pytorch-labs/tensordict/pull/476 and https://github.com/pytorch/rl/issues/1344

- [ :( ] I have raised an issue to propose this change ([required](https://github.com/pytorch/rl/issues) for new features and bug fixes)

## Types of changes

## Checklist

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -809,12 +809,10 @@ class TensorDictModuleBase(nn.Module):
             )
             repopulate_module(self, sanitized_parameters)
 
-        for child in list(self.children()):
-            child.apply(
-                lambda m: m.reset_parameters()
-                if hasattr(m, "reset_parameters")
-                else None
-            )
+        def _reset(module):
+            if hasattr(module, 'reset_parameters'):
+                module.reset_parameters()
+        self.apply(reset_parameters)
 
 
 class TensorDictModule(TensorDictModuleBase):
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -800,7 +800,7 @@ class TensorDictModuleBase(nn.Module):
             >>> (old_param == net[0].weight).any()
             tensor(False)
 
-            This method also supports functional parameter sampling:
+        This method also supports functional parameter sampling:
 
             >>> from tensordict import TensorDict
             >>> from tensordict.nn import TensorDictModule
@@ -816,6 +816,10 @@ class TensorDictModuleBase(nn.Module):
         if parameters is None:
             self._reset_parameters(self)
             return
+        elif parameters.ndim:
+            raise NotImplementedError(
+                "reset_parameters_recursive does not support batched TensorDicts, ensure batch_size is []"
+            )
 
         sanitized_parameters = parameters.apply(
             lambda x: x.detach().requires_grad_(), inplace=False
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -777,7 +777,7 @@ class TensorDictModuleBase(nn.Module):
                 del self._forward_hooks[i]
         return self
 
-    def reset_parameters(
+    def reset_parameters_recursive(
         self, parameters: Optional[TensorDictBase] = None
     ) -> Optional[TensorDictBase]:
         """Recursively reset the parameters of the module and its children.
@@ -814,7 +814,7 @@ class TensorDictModuleBase(nn.Module):
             False
         """
         if parameters is None:
-            self._reset_parameters()
+            self._reset_parameters(self)
             return
 
         sanitized_parameters = parameters.apply(
@@ -834,7 +834,7 @@ class TensorDictModuleBase(nn.Module):
                 return_old_tensordict=True,
             )
 
-        self._reset_parameters()
+        self._reset_parameters(self)
 
         if is_stateless:
             new_parameters = extract_weights_and_buffers(self)
@@ -845,13 +845,13 @@ class TensorDictModuleBase(nn.Module):
 
         return new_parameters
 
-    def _reset_parameters(self):
-        for child in self.children():
-            child.apply(
-                lambda m: m.reset_parameters()
-                if hasattr(m, "reset_parameters")
-                else None
-            )
+    def _reset_parameters(self, module: nn.Module) -> None:
+        for child in module.children():
+            if isinstance(child, nn.Module):
+                self._reset_parameters(child)
+
+            if hasattr(child, "reset_parameters"):
+                child.reset_parameters()
 
 
 class TensorDictModule(TensorDictModuleBase):
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -773,7 +773,7 @@ class TensorDictModuleBase(nn.Module):
 
     def reset_parameters(self):
         """Recursively reset the parameters of the module and its children.
-
+        
         Examples:
         >>> from tensordict.nn import TensorDictModule
         >>> from torch import nn
@@ -784,16 +784,14 @@ class TensorDictModuleBase(nn.Module):
         >>> (old_param == net[0].weight).all()
         tensor(False)
         """
-        for m in self.children():
-            self._reset_parameters(m)
+        [self._reset_parameters(m) for m in self.children()]
 
     def _reset_parameters(self, module: Union[nn.Module, TensorDictModuleBase]):
-        if isinstance(module, nn.Module):
+        if isinstance(module, Union[TensorDictModuleBase, nn.Module]):
             if hasattr(module, "reset_parameters"):
                 module.reset_parameters()
             else:
-                for m in module.children():
-                    self._reset_parameters(m)
+                [self._reset_parameters(m) for m in module.children()]
 
 
 class TensorDictModule(TensorDictModuleBase):
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -792,7 +792,8 @@ class TensorDictModuleBase(nn.Module):
             if hasattr(module, "reset_parameters"):
                 module.reset_parameters()
             else:
-                [self._reset_parameters(m) for m in module.children()]
+                for m in module.children():
+                    self._reset_parameters(m)
 
 
 class TensorDictModule(TensorDictModuleBase):
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -775,7 +775,7 @@ class TensorDictModuleBase(nn.Module):
         """Recursively reset the parameters of the module and its children.
 
         Args:
-            parameters: If set to None, the module will reset using self.parameters().
+            parameters (TensorDict of parameters, optional): If set to None, the module will reset using self.parameters().
                 Otherwise, we will reset the parameters in the tensordict in-place. This is
                 useful for functional modules where the parameters are not stored in the module itself.
 
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -789,16 +789,19 @@ class TensorDictModuleBase(nn.Module):
         >>> (old_param == net[0].weight).any()
         tensor(False)
 
-        >>> from tensordict import TensorDict
-        >>> from tensordict.nn import TensorDictModule
-        >>> from torch import nn
-        >>> net = nn.Sequential(nn.Linear(2,3), nn.ReLU())
-        >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
-        >>> params = TensorDict.from_module(module)
-        >>> old_params = params.clone(recurse=True)
-        >>> module.reset_parameters(params)
-        >>> (old_params == params).any()
-        False
+        This method also supports functional parameter sampling:
+        
+        Examples:
+            >>> from tensordict import TensorDict
+            >>> from tensordict.nn import TensorDictModule
+            >>> from torch import nn
+            >>> net = nn.Sequential(nn.Linear(2,3), nn.ReLU())
+            >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
+            >>> params = TensorDict.from_module(module)
+            >>> old_params = params.clone(recurse=True)
+            >>> module.reset_parameters(params)
+            >>> (old_params == params).any()
+            False
         """
         if parameters is not None:
             sanitized_parameters = parameters.apply(
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -783,18 +783,17 @@ class TensorDictModuleBase(nn.Module):
             A tensordict of the new parameters, only if parameters was not None.
 
         Examples:
-        >>> from tensordict.nn import TensorDictModule
-        >>> from torch import nn
-        >>> net = nn.Sequential(nn.Linear(2,3), nn.ReLU())
-        >>> old_param = net[0].weight.clone()
-        >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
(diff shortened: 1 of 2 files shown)


tensordict/nn/common.py
```diff
@@ -771,29 +771,47 @@ def reset_out_keys(self):
...
-                module.reset_parameters()
-            else:
-                for m in module.children():
-                    self._reset_parameters(m)
+        for child in list(self.children()):
+            child.apply(
+                lambda m: m.reset_parameters()
+                if hasattr(m, "reset_parameters")
+                else None
+            )
```
vmoens, line 814: Just noticed, why not

```suggestion
        def _reset(module):
            if hasattr(module, 'reset_parameters'):
                module.reset_parameters()
        self.apply(reset_parameters)
```

vmoens (commented):
Looks promising!
One catch: if the module has parameters, we don't want to loose them.
The function should have something like
```python
if params is not None:
    is_func = is_functional(self)
    if not is_func:
        make_functional(self, keep_params=True)
    is_stateless = self._is_stateless
    if is_stateless:
        # we don't need to care about the params, but we will have to get them out before returning
        repopulate_module(self, params)
    else:
         old_params = _swap_state(self, params, is_stateless=False, return_old_tensordict=True)
```

Before returning, we check if we should reset the old params
```
if is_stateless:
    # remove params
    out = extract_weights_and_buffers(self)
else:
    out = _swap_state(self, old_params, is_stateless=False, return_old_tensordict=True)
return out
```

smorad:
So I don't think we can do the `def _reset` thing because `mod.apply(fn)` will apply `fn` to all children **and itself**, so we get infinite recursion. We need to explicitly call `reset_parameters()` only on the children.

vmoens:
> So I don't think we can do the `def _reset` thing because `mod.apply(fn)` will apply `fn` to all children **and itself**, so we get infinite recursion. We need to explicitly call `reset_parameters()` only on the children.

True, but not all nn.Module have a `reset_parameters` method right? With `children`, we will break anytime a module does not have this
```python
>>> hasattr(torch.nn.Module, "reset_parameters")
False
```
I believe only our method is recursive. If so, we can modify `_reset` such that it checks the module type, and only calls modules that are not `TensorDictModuleBase` instances / modules that have parameters of their own (eg Linear vs Sequential). This logic will break once a `TensorDictModuleBase` has parameters of its own though. 

One way out of this could be to use a different name than `reset_parameters` at the `TensorDictModuleBase` level and leave `reset_parameter` only for the leaves (Linear, Conv etc). That apply does its job. Like `reset_module_params` or similar.

smorad:
Yeah if we named it something different it would fix the issue, but IMO that’s less nice from a user perspective because now you need to call different functions based on whether it’s an nn module or a tensordictmodule.

smorad:
> True, but not all nn.Module have a `reset_parameters` method right? With `children`, we will break anytime a module does not have this

I'm not sure I completely understand what you're saying, we call `apply` on the child, which is recursive. So the immediate child (or even the nth grandchild) does not need to implement `reset_parameters` as long as the leaf does. I explicitly test for this (the modulelist+sequential are two layers of no `reset_parameters`):

```python
    @pytest.mark.parametrize(
        "net",
        [
            nn.ModuleList([nn.Sequential(nn.Linear(1, 1), nn.ReLU())]),
            nn.Linear(2, 1),
            nn.Sequential(nn.Tanh(), nn.Linear(1, 1), nn.Linear(2, 1)),
        ],
    )
```

vmoens:
> Yeah if we named it something different it would fix the issue, but IMO that’s less nice from a user perspective because now you need to call different functions based on whether it’s an nn module or a tensordictmodule.

but all nn.Module not having a reset_parameters, and all reset_parameters being AFAICT non-recursive, i'm not sure that it makes sense to build a method that does a different job (recursively calling reset_parameters if there is) with the same name

vmoens:
> I'm not sure I completely understand what you're saying, we call `apply` on the child, [which is recursive](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.apply) and will attempt call to `reset_parameters` on all descendants. It does not "break" on a child not implementing `reset_parameters`. 

Here's an example where we I believe we have an infinite loop because the names clash:
```python
class MyClass(TensorDictModuleBase):
    in_keys=["a"]
    out_keys=["b"]
    def forward(self, tensordict):
        a = tensordict.get("a")
        tensordict.set("b", a+1)
        return tensordict

class MySuperClass(TensorDictModuleBase):
    in_keys=["a"]
    out_keys=["b"]
    def __init__(self):
         super().__init__()
         self.module = MyClass()
    def forward(self, tensordict):
        return self.module(tensordict)
```
It's a somewhat valid example since we don't prohibit nested `TensorDictModuleBase`

smorad:
This is equivalent to something like this, right?

```python
module = TensorDictModule(net, in_keys=["in"], out_keys=["out"])
nested_module = TensorDictModule(module, in_keys=["in"], out_keys=["out"])
tripled_nested = TensorDictModule(nested_module, in_keys=["in"], out_keys=["out"])
```

I've added a unit test showing that this is fine. I've inspected the parameters within the test to verify
```python
(Pdb) !list(new_params.values(include_nested=True, leaves_only=True))
[tensor([[-0.0198]], grad_fn=<CloneBackward0>), tensor([0.7929], grad_fn=<CloneBackward0>)]
(Pdb) !list(old_params.values(include_nested=True, leaves_only=True))
[tensor([[0.5672]], grad_fn=<CloneBackward0>), tensor([0.6473], grad_fn=<CloneBackward0>)]
```

smorad:
We're talking about the current implementation, right? Not the theoretical `def _reset` version? I apologize for not understanding. Maybe if you have time we could jump on a quick call?

```python
In [1]: from torch import nn

In [2]: from tensordict.nn.common import TensorDictModuleBase
In [5]: class MyClass(TensorDictModuleBase):
   ...:     in_keys=["a"]
   ...:     out_keys=["b"]
   ...:     def __init__(self):
   ...:         super().__init__()
   ...:         self.p = nn.Linear(1,1)
   ...:     def forward(self, tensordict):
   ...:         a = tensordict.get("a")
   ...:         tensordict.set("b", a+1)
   ...:         return tensordict
   ...: 
   ...: class MySuperClass(TensorDictModuleBase):
   ...:     in_keys=["a"]
   ...:     out_keys=["b"]
   ...:     def __init__(self):
   ...:          super().__init__()
   ...:          self.module = MyClass()
   ...:     def forward(self, tensordict):
   ...:         return self.module(tensordict)
   ...: 
In [6]: c = MySuperClass()
In [10]: c.module.p.weight
Out[10]: 
Parameter containing:
tensor([[-0.4766]], requires_grad=True)
In [11]: c.reset_parameters()
In [12]: c.module.p.weight
Out[12]: 
Parameter containing:
tensor([[-0.2866]], requires_grad=True)
```

You have a valid point about `reset_parameters` operating recursively that might be confusing. I can rename it to `recurse_reset_parameters` or something.

vmoens:
> > It's a somewhat valid example since we don't prohibit nested TensorDictModuleBase
> 
> This is equivalent to something like this, right?
> 
> ```python
> net = nn.ModuleList([nn.Sequential(nn.Linear(1, 1), nn.ReLU())])
> module = TensorDictModule(net, in_keys=["in"], out_keys=["out"])
> nested_module = TensorDictModule(module, in_keys=["in"], out_keys=["out"])
> tripled_nested = TensorDictModule(nested_module, in_keys=["in"], out_keys=["out"])
> ```
> 
> I've added a unit test showing that this is fine. I've inspected the parameters within the test to verify
> 
> ```python
> (Pdb) !list(new_params.values(include_nested=True, leaves_only=True))
> [tensor([[-0.0198]], grad_fn=<CloneBackward0>), tensor([0.7929], grad_fn=<CloneBackward0>)]
> (Pdb) !list(old_params.values(include_nested=True, leaves_only=True))
> [tensor([[0.5672]], grad_fn=<CloneBackward0>), tensor([0.6473], grad_fn=<CloneBackward0>)]
> ```

I modified the private `_reset_parameters` in this way:
```python
    def _reset_parameters(self):
        def _reset(m):
            if hasattr(m, "reset_parameters"):
                print('calling reset on', id(m))
                m.reset_parameters()
        for child in self.children():
            child.apply(_reset)
```

Then I called the script you had above:
```python
from tensordict.nn import TensorDictModule
from torch import nn
import torch

net = nn.ModuleList([nn.Sequential(nn.Linear(1, 1), nn.ReLU())])
module = TensorDictModule(net, in_keys=["in"], out_keys=["out"])
nested_module = TensorDictModule(module, in_keys=["in"], out_keys=["out"])
tripled_nested = TensorDictModule(nested_module, in_keys=["in"], out_keys=["out"])
print(id(net[0][0]), id(module), id(nested_module), id(tripled_nested))

tripled_nested.reset_parameters()
```
It prints this
```
4894986448 4894989904 4894987648 4894987552
calling reset on 4894986448
calling reset on 4894989904
calling reset on 4894986448
calling reset on 4894987648
calling reset on 4894986448
calling reset on 4894989904
calling reset on 4894986448
```
Hence we're calling reset_parameters multiple times.
The reason is that triple nested calls apply on nested. During this call to apply (which will traverse the leaf modules), another call to apply is initiated, which traversed the leaf modules a second time.

Happy to jump on a call to debulk it!

smorad:
Ah ok, got it! I'll have a fix up soon.

smorad:
Bump :3

tensordict/nn/common.py
```diff
@@ -771,29 +777,81 @@ def reset_out_keys(self):
...
+            >>> old_params = params.clone(recurse=True)
+            >>> module.reset_parameters(params)
+            >>> (old_params == params).any()
+            False
         """
-        for m in self.children():
-            self._reset_parameters(m)
+        if parameters is None:
+            self._reset_parameters(self)
+            return
```
vmoens, line 818: Can we also test this? (for usage with vmap)

```suggestion
            return
        elif parameters.ndim:
            for param in parameters.unbind(0):
                self.reset_parameters_recursive(self, param)
            return
```

vmoens (commented):
Cool! Thanks so much! Almost there!
I just added one little thing regarding batches of params (e.g. model ensembling), can you check it works and add it to the doc + tests?

smorad:
> I just added one little thing regarding batches of params (e.g. model ensembling), can you check it works and add it to the doc + tests?

I'm not sure this will work!
```python
(Pdb) torch.vmap(seq.reset_parameters_recursive)(new_params)
*** RuntimeError: You are attempting to call Tensor.requires_grad_() (or perhaps using torch.autograd.functional.* APIs) inside of a function being transformed by a functorch transform. This is unsupported, please attempt to use the functorch transforms (e.g. grad, vjp, jacrev, jacfwd, hessian) or call requires_grad_() outside of a function being transformed instead.
```
Fortunately, in my later diffs I show ensembling isn't too hard using the current implementation:
```python
params_td = make_functional(module).expand(num_copies).to_tensordict()
module.reset_parameters(params_td)
```

vmoens:
> > I just added one little thing regarding batches of params (e.g. model ensembling), can you check it works and add it to the doc + tests?
> 
> I'm not sure this will work!
> 
> ```python
> (Pdb) torch.vmap(seq.reset_parameters_recursive)(new_params)
> *** RuntimeError: You are attempting to call Tensor.requires_grad_() (or perhaps using torch.autograd.functional.* APIs) inside of a function being transformed by a functorch transform. This is unsupported, please attempt to use the functorch transforms (e.g. grad, vjp, jacrev, jacfwd, hessian) or call requires_grad_() outside of a function being transformed instead.
> ```
> 

Yeah your solution is exactly what I was pointing. The dispatching I was proposing is just to make sure that if any model is super specific regarding the shape we don't break, e.g.
```
class MyModule(nn.Module):
    def reset_parameters(self):
        self.params.copy_(self.randn(3, 4, 5))
```
I believe your solution will break in that case (I might have overlooked something)

smorad:
The ensemble tests passed (or at least did before the last round of changes). For most cases, it should be ok. `nn.init.kaiming_uniform_()` can be called with however many leading dims. But `nn.init.orthogonal_` would cause an issue because it will flatten all dims >2. It will not crash, but the weights won't be orthogonal.

If `vmap` does not support `requires_grad` I see a few ways forward:
1. Keep as-is, where the user cannot `vmap` `reset_parameters_recursive`. Rather than `params_td = make_functional(module).expand(num_copies).to_tensordict()` ensembles can be initialized using a for loop
```python
for _ in range(num_copies):
  param_list.append(module.reset_parameters_recursive(params.clone())
...
ensemble_params = torch.stack(param_list)
```
2. Use vmap, assume the user has called `params.detach().requires_grad()`, also write a very big warning that the user must do so.

vmoens:
The solution I was suggesting is to check if the params have a batch size. if they do, we simply unbind them and iterate over the list, calling reset on each copy, then stack them back together.
`vmap` currently breaks with random number generation and here we're defo doing that, so it's not a viable solution.
Thanks again for this feature, it's extra useful!

smorad:
I've run into another issue -- I'm unable to recursively call `unbind` and `stack`
```python
            for param in parameters.unbind(0):
                batched_params.append(self.reset_parameters_recursive(param))
            return torch.stack(batched_params, dim=-1)
```
I'm away from my laptop so I don't have the full error message, but torch is angry that I'm doing this in-place. With the now reverted commit, I modified `reset_parameters_recursive` to be out of place. This worked, however, then the user needs to call `repopulate_module` which does not support extra leading ensemble dimensions in the parameter tensordict.

I think logically, the latest commit is fine as-is. I don't think `reset_parameters_recursive` should handle the ensemble case. All ensembling can be done much more easily in a `TensorDictModule` external to `reset_parameters_recursive`, where we don't run into these various issues. Furthermore, I think keeping all ensembling logic in an `Ensemble` class is a cleaner logical separation.

tensordict/nn/common.py
```diff
@@ -771,29 +777,81 @@ def reset_out_keys(self):
...
+            >>> from tensordict.nn import TensorDictModule
+            >>> from torch import nn
+            >>> net = nn.Sequential(nn.Linear(2,3), nn.ReLU())
+            >>> old_param = net[0].weight.clone()
+            >>> module = TensorDictModule(net, in_keys=['bork'], out_keys=['dork'])
+            >>> module.reset_parameters()
+            >>> (old_param == net[0].weight).any()
+            tensor(False)
+
+            This method also supports functional parameter sampling:
```
vmoens, line 803:
```suggestion
        This method also supports functional parameter sampling:
```

vmoens (approved):
LGTM, see minor comment.
I agree with your point. Perhaps we should check that the TD does not have a non-empty batch size then (eg to avoid wrong init with orthogonal etc)?

tensordict/nn/common.py
```diff
@@ -816,6 +816,10 @@ def reset_parameters_recursive(
         if parameters is None:
             self._reset_parameters(self)
             return
+        elif parameters.ndim:
+            raise NotImplementedError(
+                "reset_parameters_recursive does not support batched TensorDicts, ensure batch_size is []"
```
vmoens, line 821:
```suggestion
                "reset_parameters_recursive does not support batched TensorDicts, ensure `batch_size` is empty and the parameters shape match their original shape."
```

vmoens (approved):

PR #479, opened 2023-07-06
[BugFix] Nested keys to probabilistic modules

matteobettini:
FIxes https://github.com/pytorch/rl/issues/1279


vmoens:
I think this isn't synced with main

matteobettini:
Oh noo :(

tensordict/nn/probabilistic.py
```diff
@@ -258,25 +261,28 @@ class ProbabilisticTensorDictModule(TensorDictModuleBase):
...
+        log_prob_key: Optional[NestedKey] = "sample_log_prob",
         cache_dist: bool = False,
         n_empirical_estimate: int = 1000,
     ) -> None:
         super().__init__()
-        if isinstance(in_keys, str):
+        if isinstance(in_keys, str) or (
+            isinstance(in_keys, tuple) and set(map(type, in_keys)) == {str}
+        ):
             in_keys = [in_keys]
```
vmoens, line 282: What are we trying to cover here? Why isn't a tuple check enough? Do we expect tuples with other things than strings?

tensordict/nn/probabilistic.py
```diff
@@ -258,25 +261,28 @@ class ProbabilisticTensorDictModule(TensorDictModuleBase):
 
     """
 
-    SAMPLE_LOG_PROB_KEY = "sample_log_prob"
```
vmoens, line 261: maybe create a property that links to the new attribute for bc compatibility (+ deprec warning)

vmoens (approved):
LGTM 
see some minor comments

tensordict/nn/probabilistic.py
matteobettini: since we allow iterables the tuple could be the sequence of keys or a nested key itself. here we check it is the second case

tensordict/nn/probabilistic.py
matteobettini: i have tried to do this, is it what you were looking for?

tensordict/nn/probabilistic.py
vmoens: how can we differentiate 
```
{"a"}
```
from
```
{("a",)}
```
from
```
(("a",))
```
from
```
(["a",])
```
?

I think we should allow either: str/tuple => consider as single key // list => consider as list of keys

tensordict/nn/probabilistic.py
```diff
@@ -275,9 +275,9 @@ def __init__(
         n_empirical_estimate: int = 1000,
     ) -> None:
         super().__init__()
-        if isinstance(in_keys, str):
+        if isinstance(in_keys, str) or (isinstance(in_keys,tuple) and set(map(type, in_keys)) == {str}):
```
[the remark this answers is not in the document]
matteobettini: The problem is:

This check was meant to identify if a single key has been passed as arg.
But we allow iterables of nestedkeys (which includes tuples).
Therefore there could be the following cases:
- in_keys="ciao" # Should return true
- in_keys=("ciao",) # Should return true
- in_keys=("ciao","oone") # Should return true
- in_keys=(("ciao","oone"),"other") # Should return false
- in_keys=(("ciao",),"other") # Should return false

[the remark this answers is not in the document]
matteobettini: i am open to suggestions though, feel free to change it.
We could also only allow list of nestedkeys to cut it short

tensordict/nn/probabilistic.py
matteobettini: ok yeah let's do this so we can "cut the tail to the cat".

Should I update the docstring to list of nested key?

vmoens (approved):
LGTM!

PR #482, opened 2023-07-08
[Performance] Better names handling in LazyStackTD

vmoens:


vmoens:
Merging as it solves a bug formerly hidden in RL

PR #483, opened 2023-07-10
Removing myself from notified users in benchmarks

tcbegley:
I think unfortunately it no longer makes sense to ping me with these regressions. Perhaps there is someone else to whom I can pass the torch (pun not intended but enjoyed nonetheless 😄).


vmoens (approved):
Looks terrible to me...

PR #487, opened 2023-07-11
[Performance] Faster dispatch

vmoens:


tensordict/tensordict.py
```diff
@@ -36,8260 +36,3 @@
     lock_blocked,
     NestedKey,
 )
-from torch import distributed as dist, Tensor
```
shagunsodhani: Where is all this stuff going ?

PR #488, opened 2023-07-14
[BugFix] Fix indexing of lazy stacks

vmoens:
Removes support for list of bools as this is not supported by torch


matteobettini (approved):
LGTM

PR #489, opened 2023-07-17
[CI] Fix doc CI

matteobettini:


matteobettini:
@vmoens we may also want the stuff from the rl pr regarding filing and exposing the artifacts

vmoens (approved):
LGTM

vmoens:
> @vmoens we may also want the stuff from the rl pr regarding filing and exposing the artifacts

As in: moving files to s3? Sure we can do that in a follow up

matteobettini:
It was more about the possibility of building the doc and downloading the artifacts from the ci but yes we can do later

matteobettini:
Actually i think the CI of this PR was broken, we might wanna remove versions for mystpaser and docutils too

PR #499, opened 2023-07-26
[Feature] Cat lazy tds

matteobettini:


tensordict/tensordict.py
```diff
@@ -4668,24 +4680,36 @@ def assert_allclose_td(
...
-        if _is_tensor_collection(input1.__class__):
-            assert_allclose_td(input1, input2, rtol=rtol, atol=atol)
-            continue
+        shape1 = actual.get_item_shape(key)
+        shape2 = expected.get_item_shape(key)
+        if -1 in shape1 or -1 in shape2:
+            if not (-1 in shape1 and -1 in shape2):
+                raise KeyError(
+                    f"{key} corresponds to an heterogeneous entry only in one of the tensordicts provided"
+                )
```
vmoens, line 4689: Isn't this a bit loose?
What if you have [-1, 2] and [2, -1]?

tensordict/tensordict.py
```diff
@@ -4668,24 +4680,36 @@ def assert_allclose_td(
...
     keys = sorted(actual.keys(), key=str)
     for key in keys:
-        input1 = actual.get(key)
-        input2 = expected.get(key)
-        if _is_tensor_collection(input1.__class__):
-            assert_allclose_td(input1, input2, rtol=rtol, atol=atol)
-            continue
+        shape1 = actual.get_item_shape(key)
+        shape2 = expected.get_item_shape(key)
+        if -1 in shape1 or -1 in shape2:
```
vmoens, line 4685: Can't we just loop over sub-tds if it's a lazy stack?

tensordict/tensordict.py
```diff
@@ -4902,6 +4926,78 @@ def _cat(
...
+    batch_size = list(list_of_tensordicts[0].batch_size)
+    if dim < 0:
+        dim = len(batch_size) + dim
+    if dim >= len(batch_size):
+        raise RuntimeError(
+            f"dim must be in the range 0 <= dim < len(batch_size), got dim"
+            f"={dim} and batch_size={batch_size}"
+        )
+    stack_dim = list_of_tensordicts[0].stack_dim
+    if any([(td.stack_dim != stack_dim) for td in list_of_tensordicts]):
```
vmoens, line 4947:
```suggestion
    if any((td.stack_dim != stack_dim) for td in list_of_tensordicts):
```
image

tensordict/tensordict.py
```diff
@@ -4902,6 +4926,78 @@ def _cat(
...
+        raise RuntimeError("cat lazy stacked tds must have same stack dim")
+
+    batch_size[dim] = sum([td.batch_size[dim] for td in list_of_tensordicts])
```
vmoens, line 4950:
```suggestion
    batch_size[dim] = sum(td.batch_size[dim] for td in list_of_tensordicts)
```

tensordict/tensordict.py
```diff
@@ -4902,6 +4926,78 @@ def _cat(
...
+    batch_size = torch.Size(batch_size)
+
+    new_dim = dim
+    if dim > stack_dim:
+        new_dim = dim - 1
+
+    # check that all tensordict match
+    _check_keys(list_of_tensordicts, strict=True)
```
vmoens, line 4958: I wonder if it's not better to avoid this check and just raise an exception during the cat if anything breaks?

tensordict/tensordict.py
```diff
@@ -4902,6 +4926,78 @@ def _cat(
...
+            raise RuntimeError(
+                "out.stack_dim and elements stack_sim did not match, "
+                f"got out.stack_dim={out.stack_dim} and stack_dim={stack_dim}"
+            )
+
+        if dim == stack_dim:
+            td_list = []
+            for lazy_td in list_of_tensordicts:
+                td_list += lazy_td.tensordicts
+            out.tensordicts = td_list
```
vmoens, line 4990: If we have cached values this will break (maybe not now but in the future)
What about `out.append`?

vmoens (approved):
LGTM some minor comments

tensordict/tensordict.py
matteobettini: oh i didn't know we could do that. absolutely, thanks!

tensordict/tensordict.py
matteobettini: we can. also because this check will be run in the normal td cat

tensordict/tensordict.py
matteobettini: we first need to clear out.tensordicts right?
cause when we are here we know that out will have `len(tensordicts) == sum(len(td.tensordicts) for td in list_of_tensordicts) ` 

could you scketch out what you have in mind?

vmoens: you're right but the thing is that this is risky as we change the content of the list without checking what we're doing.
Also thinking about it: this isn't really using `out`, since we discard the data. We should update each tensor in-place, not allocate new tensors

matteobettini: okay so something like
```
 out.tensordicts[i].update( torch.cat(
                    [lazy_td.tensordicts[i] for lazy_td in list_of_tensordicts],
                    new_dim,
                ))
```
?

matteobettini:
i changed the out=smth branch to user `update(, inplace=True)`

I am experiencing some bugs with update in the tests and I dunno if this is just my tests or update() has some problems with lazystacks

test/test_tensordict.py
```diff
@@ -4271,9 +4271,54 @@ def nested_lazy_het_td(batch_size):
...
+        index = (slice(None),) * cat_dim + (slice(0, td_lazy.shape[cat_dim]),)
+        assert assert_allclose_td(res[index], td_lazy)
+        index = (slice(None),) * cat_dim + (slice(td_lazy.shape[cat_dim], None),)
+        assert assert_allclose_td(res[index], td_lazy_2)
+
+        if cat_dim != len(batch_size):  # cat dim is not stack dim
+            batch_size = list(batch_size)
+            batch_size[cat_dim] *= 2
+            td_lazy_dest = self.nested_lazy_het_td(batch_size)["lazy"]
+            res = torch.cat([td_lazy, td_lazy_2], dim=cat_dim, out=td_lazy_dest)
```
vmoens, line 4315: when testing with `out` we should test that all the leaves of `out` before stack have the same data_ptr() of the leaves after stack:
```python
def decompose(td):
    if isinstance(td, LazyStackedTensorDict):
        for td in td.tensordicts:
            yield from decompose(td)
    else:
        for v in self.values():
            if is_tensor_collection(v):
                yield from decompose(v)
            else:
                yield v

data_ptr_set_before = {val.data_ptr() for val in decompose(out)}
torch.stack(..., out=out)
data_ptr_set_after = {val.data_ptr() for val in decompose(out)}
assert data_ptr_set_after == data_ptr_set_before
```

tensordict/tensordict.py
```diff
@@ -4910,6 +4929,81 @@ def _cat(
...
+            raise RuntimeError(
+                "out.batch_size and cat batch size must match, "
+                f"got out.batch_size={out.batch_size} and batch_size"
+                f"={batch_size}"
+            )
+        if out.stack_dim != stack_dim:
+            raise RuntimeError(
+                "out.stack_dim and elements stack_sim did not match, "
+                f"got out.stack_dim={out.stack_dim} and stack_dim={stack_dim}"
+            )
```
vmoens, line 4985: Is this technically impossible or too much trouble?

matteobettini: i am not sure. 
I have difficulty picturing what it would look like, what do you think?

test/test_tensordict.py
matteobettini: well this was a great suggestion as it led to finidng bugs in the update functions which did not recursively pass args

matteobettini: i added it to all the tests with out=smth so that we can sleep at night

tensordict/tensordict.py
vmoens: let me think about that

matteobettini: I am not understanding this part
```
dim = 0
for i, sub_dest in enumerate(td1.tensordicts):
    index = (slice(None),) * td1.stack_dim + (i,)
    tds_to_cat = [_td[index] for _td in tds]
    torch.cat(tds_to_cat, dim, out=sub_dest)
```

here at the end you are trying to cat lazy stacks onto a tensordict (non-lazy), how is this supposed to work?

vmoens: I'm indexing `td1` along its stack-dim, so I get a regular td if it's a stack of regular tds. That gives me `tds_to_cat`
Since I enumerate all the tds in td1 to get `sub_dest` I get the same dim reduction in both

matteobettini: `tds_to cat` is a list of lazystackedtds while `sub_dest` is a regular td. 
Is that cat operation supposed to handle that?

matteobettini: in other words, you are asking the lazy cat to work with a non lazy output, seems impossible to me

tensordict/tensordict.py
```diff
@@ -4910,6 +4929,74 @@ def _cat(
...
+                out.append(
+                    torch.cat(
+                        [lazy_td.tensordicts[i] for lazy_td in list_of_tensordicts],
+                        new_dim,
+                    )
+                )
+        return LazyStackedTensorDict(*out, stack_dim=stack_dim)
+    else:
+        if isinstance(out, TensorDict):
+            return _cat(list_of_tensordicts, dim=dim, out=out)
```
vmoens, line 4976: what if it's a SubTensorDict or another subtype? I guess we'd like to check if it's not a LazyStack no?

vmoens (approved):
Minor comment then good to go

PR #501, opened 2023-07-26
[Feature] PyTree compatibility

vmoens:
## Description

Makes tensordict compatible with pytree.
Solves the conflict with vmap between custom _add_batch_dims and torch version. 
We ignore the tensordict node in pytree within vmap and rely on TensorDict._add_batch_dims instead.

cc @NicolasHug @ezyang @zou3519

### DIFF
--- a/tensordict/_pytree.py
+++ b/tensordict/_pytree.py
@@ -11,7 +11,34 @@ from tensordict import (
     TensorDict,
 )
 
-from torch.utils._pytree import _register_pytree_node, _str_to_dict, Context
+from torch.utils._pytree import _register_pytree_node, Context
+
+
+def _str_to_dict(str_spec: str) -> Tuple[List[str], str]:
+    assert str_spec[1] == "("
+    assert str_spec[-1] == ")"
+    context_and_child_strings = str_spec[2:-1]
+
+    child_strings = []
+    context_strings = []
+    nested_parentheses = 0
+    start_index = 0
+    for i, char in enumerate(context_and_child_strings):
+        if char == ":":
+            if nested_parentheses == 0:
+                context_strings.append(context_and_child_strings[start_index:i])
+                start_index = i + 1
+        elif char == "(":
+            nested_parentheses += 1
+        elif char == ")":
+            nested_parentheses -= 1
+
+        if nested_parentheses == 0 and char == ",":
+            child_strings.append(context_and_child_strings[start_index:i])
+            start_index = i + 1
+
+    child_strings.append(context_and_child_strings[start_index:])
+    return context_strings, ",".join(child_strings)
 
 
 def _maybe_str_to_tensordict(str_spec: str) -> Optional[Tuple[Any, Context, str]]:
--- a/tensordict/_pytree.py
+++ b/tensordict/_pytree.py
@@ -48,13 +48,6 @@ def _str_to_dict(str_spec: str) -> Tuple[List[str], str]:
     return context_strings, ",".join(child_strings)
 
 
-def _maybe_str_to_tensordict(str_spec: str) -> Optional[Tuple[Any, Context, str]]:
-    if not str_spec.startswith("D"):
-        return None
-    context_strings, child_strings = _str_to_dict(str_spec)
-    return TensorDict, context_strings, child_strings
-
-
 def _str_to_tensordictdict(str_spec: str) -> Tuple[List[str], str]:
     context_and_child_strings = str_spec[2:-1]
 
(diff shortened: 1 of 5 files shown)


tensordict/_pytree.py
```diff
@@ -0,0 +1,106 @@
...
+
+    child_strings.append(context_and_child_strings[start_index:])
+    return context_strings, ",".join(child_strings)
+
+
+def _maybe_str_to_tensordict(str_spec: str) -> Optional[Tuple[Any, Context, str]]:
+    if not str_spec.startswith("D"):
+        return None
+    context_strings, child_strings = _str_to_dict(str_spec)
+    return TensorDict, context_strings, child_strings
```
ezyang, line 48: cc @zou3519

We're going to delete these APIs soon, you don't have to implement it.

ezyang:
Looks reasonable! 

@zou3519 for the non-monkeypatched vmap support, maybe we need some way for non-Tensor objects to say "hey, I need special batch dim handling"?

tensordict/_pytree.py
vmoens: Indeed I saw that the tests were breaking, I guess it's because of that

vmoens:
> Looks reasonable!
> 
> @zou3519 for the non-monkeypatched vmap support, maybe we need some way for non-Tensor objects to say "hey, I need special batch dim handling"?

There are 2 things we monkey patch in tensordict:
- first we need our "own" add_batch_dims since tensordict has its own shape: unlike regular dicts, we need to "hide" part of the shape when we vmap over a td.
Example:
```python
fun = lambda x, td: td.set("y", x+1)
td = TensorDict({}, torch.Size([3]))
x = torch.randn(4)
vmap(fun, (None, 0))(x, td)
```
We "hide" the `torch.Size([3])` within the TD to make it broadcastable with shape `torch.Size([4])` of the tensor `x` then recompute the batch size of the output accordingly.
The result is this:
```
TensorDict(
    fields={
        y: Tensor(shape=torch.Size([3, 4]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([3]),
    device=None,
    is_shared=False)
```

2. The monkey patching from this PR follows the other one: we don't want to decompose the tensordict because:
  a. if we keep the batch size as metadata it will be hard to manipulate it when we want to reconstruct the tensordict with a new batch size
  b. it is inefficient (building nested tensordicts has some non-negligeable overhead due to the shape/metadata checks we do over the leading dimensions of the tensors -- internally we can skip those checks within tensordict because we know what we're feeding to our tensordict).

Having a registration mechanism for vmap as you suggest would be awesome but that would require to look at both of the things I mentioned here: how to have a custom add_batch_dims / remove_batch_dims for a specific class and how to skip pytree's flatten ops if there is a custom dim operation that will follow.

matteobettini (approved):
LGTM

zou3519:
I'm not sure how I feel about extending custom object support in vmap. The current contract is that vmap does not support custom objects, unless you have a pytree.

One workaround I can brainstorm that doesn't involve monkeypatching is to store TensorDict's batch size as a Meta tensor. For example, if we construct a TensorDict({}, batch_size=[3, 4]), then we store a Meta tensor of size (3, 4) somewhere in the TensorDict. Whenever TensorDict needs to know its batch size, it can query the shape of the Meta tensor.

When one vmaps over that, assuming TensorDict is a pytree, then we should end up with the correct semantics (under one vmap, that Meta tensor's shape would be [4], which is what we want).

Now, this doesn't quite solve the performance problem. We've been moving towards "use torch.compile to eliminate overhead for performance" though, so in a long-term state we'd just use torch.compile to make things faster.

zou3519:
Do all the Tensors in the TensorDict have the same batched-ness? E.g. if batch_size is (3, 4), do all the Tensors have shape (3, 4, *)?

vmoens:
> Do all the Tensors in the TensorDict have the same batched-ness? E.g. if batch_size is (3, 4), do all the Tensors have shape (3, 4, *)?

Yes but that does not mean that each level has the same batch size:

You could have 

```python
        td = TensorDict(
            {
                "a": TensorDict(
                    {
                        "b": TensorDict({"c": torch.ones(2, 3, 4)}, batch_size=[2, 3]),
                        "d": torch.ones(2),
                    },
                    batch_size=[2],
                ),
                "e": 1,
            },
            batch_size=[],
        )
```
where each level of the structure has one more batch dimension.
For instance, you could have a dataset with batches of data, each containing a stack of tensordicts stacked together...

PR #502, opened 2023-07-28
[Feature] Hooks and Buffers for TensorDictParams

vmoens:
## Description

Introduces 2 features for TensorDictParams:
- `pre_get_hooks` allow users to register operations to be executed every time a `get` operation is executed (incl. values, items, `__getitem__`, `get`, `_get_str`, `_get_tuple`). For instance, this allows to create an instance of TensorDictParams that gets the `.data` attribute of another one while keeping the content similar after casting to device or dtype.
- Introduces a `Buffer` class, similar to `nn.Parameter` but without gradient by default. If a tensor is not a parameter in TensorDictParam, it will be a `Buffer`. This change allows us to call `TensorDictParams.to(...)` while being sure that every leaf of the structure has the same identity before and after the operation, as well as the buffers in `_buffers`. Not using this class would result in the content of `_buffers` being updated but not the content of the tensordict, thereby breaking the sync between the two.

cc @matteobettini

### DIFF
--- a/tensordict/nn/params.py
+++ b/tensordict/nn/params.py
@@ -41,9 +41,15 @@ def _apply_leaves(data, fn):
                 )
         return data
     elif isinstance(data, LazyStackedTensorDict):
-        for _data in data.tensordicts:
-            _apply_leaves(_data, fn)
-        return data
+        # this is currently not implemented as the registration of params will only work
+        # with plain TensorDict. The solution will be using pytree to get each independent
+        # leaf
+        raise RuntimeError(
+            "Using a LazyStackedTensorDict within a TensorDictParams isn't permitted."
+        )
+        # for _data in data.tensordicts:
+        #     _apply_leaves(_data, fn)
+        # return data
     elif isinstance(data, _CustomOpTensorDict):
         _apply_leaves(data._source, fn)
         return data
@@ -130,12 +136,7 @@ def _get_post_hook(func):
     @wraps(func)
     def new_func(self, *args, **kwargs):
         out = func(self, *args, **kwargs)
-        if not _is_tensor_collection(type(out)):
-            for hook in self._get_post_hook:
-                new_out = hook(self, out)
-                if new_out is not None:
-                    out = new_out
-        return out
+        return self._apply_get_post_hook(out)
 
     return new_func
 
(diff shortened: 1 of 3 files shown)


tensordict/nn/params.py
```diff
@@ -196,15 +264,45 @@ class TensorDictParams(TensorDictBase, nn.Module):
...
+    def _reset_params(self):
+        parameters = self._param_td
+        param_keys = [
+            key
+            for key, value in parameters.items(True, True)
+            if isinstance(value, nn.Parameter)
+        ]
+        buffer_keys = [
+            key
+            for key, value in parameters.items(True, True)
```
matteobettini, line 297: here we are traversing the whole params tree twice.
if this function is called in other parts other than init it might be worth traversing just once

tensordict/nn/params.py
```diff
@@ -69,19 +108,38 @@ def new_func(_self, *args, **kwargs):
...
             return out
 
         return new_func
 
 
+def _get_post_hook(func):
+    @wraps(func)
+    def new_func(self, *args, **kwargs):
+        out = func(self, *args, **kwargs)
+        if not _is_tensor_collection(type(out)):
```
matteobettini, line 133: so hooks are only for leaves? why?

tensordict/nn/params.py
```diff
@@ -567,6 +684,120 @@ def create_nested(self, key):
...
+        self, include_nested: bool = False, leaves_only: bool = False
+    ) -> Iterator[CompatibleType]:
+        for k, v in self._param_td.items(include_nested, leaves_only):
+            if _is_tensor_collection(type(v)):
+                yield k, v
+                continue
+            for hook in self._get_post_hook:
+                new_v = hook(self, v)
+                if new_v is not None:
+                    v = new_v
```
matteobettini, line 710: this code is repeated 3 times. maybe consider a function `_apply_hooks`

matteobettini (approved):
LGTM, some nits

tensordict/nn/params.py
vmoens: it's applied recursively
If I do something like `hook = lambda x: x*0` that won't work on tensordict instances

PR #506, opened 2023-08-02
[Feature] `dense_stack_tds`

matteobettini:


tensordict/tensordict.py
```diff
@@ -8646,6 +8646,72 @@ def make_tensordict(
...
+            If td_list is a LazyStackedTensorDict, it will be retrieved automatically.
+
+    Examples:
+        >>> import torch
+        >>> from tensordict import TensorDict
+        >>> from tensordict import dense_stack_tds
+        >>> from tensordict.tensordict import assert_allclose_td
+        >>> a = TensorDict({"a": torch.zeros(3)},[])
+        >>> b = TensorDict({"a": torch.zeros(4), "b": torch.zeros(2)},[])
+        >>> td_lazy = torch.stack([a,b], dim=0)
```
vmoens, line 8672: the names are a bit confusing

tensordict/tensordict.py
```diff
@@ -8646,6 +8646,72 @@ def make_tensordict(
...
 
 
+def dense_stack_tds(
+    td_list: Sequence[TensorDictBase] | LazyStackedTensorDict,
+    dim: int = None,
+) -> TensorDictBase:
+    """Densely stack a list of :class:`tensordict.TensorDictBase` objects (or a :class:`tensordict.LazyStackedTensorDict`) given that they have the same structure.
+
+    This must be used when some of the :class:`tensordict.TensorDictBase` objects that need to be stacked
+    can have :class:`tensordict.LazyStackedTensorDict` among entries (or nested entries).
```
vmoens, line 8656: why "can have"? why not just "have"?

vmoens: `among their entries`

tensordict/tensordict.py
```diff
@@ -8646,6 +8646,72 @@ def make_tensordict(
...
+    In those cases, calling ``torch.stack(td_list).to_tensordict()`` is infeasible.
+    Thus, this function provides an alternative for densely stacking the list provided.
```
vmoens, line 8658: it's unclear what this function does from this docstring. We say which problem we're trying to solve but not what we do

tensordict/tensordict.py
```diff
@@ -8646,6 +8646,72 @@ def make_tensordict(
...
+        LazyStackedTensorDict(
+            fields={
+                a: Tensor(shape=torch.Size([2, 2, -1]), device=cpu, dtype=torch.float32, is_shared=False)},
+            exclusive_fields={
+                1 ->
+                    b: Tensor(shape=torch.Size([2, 2]), device=cpu, dtype=torch.float32, is_shared=False)},
+            batch_size=torch.Size([2, 2]),
+            device=None,
+            is_shared=False,
+            stack_dim=1)
```
vmoens, line 8695: the appearance of `b` seems to be the only difference but it's still a lazy key
Why do we need 2 levels of nesting? This example isn't super clear to me

test/test_tensordict.py
```diff
@@ -5906,6 +5907,46 @@ def test_empty():
...
+def test_dense_stack_tds(stack_dim, nested_stack_dim):
+    batch_size = (5, 6)
+    a = TensorDict(
+        {"a": torch.zeros(*batch_size, 3)},
+        batch_size,
+    )
+    b = TensorDict(
+        {"a": torch.zeros(*batch_size, 4), "b": torch.zeros(*batch_size, 2)},
+        batch_size,
+    )
```
vmoens, line 5927: names of the tds are a bit confusing

test/test_tensordict.py
```diff
@@ -5906,6 +5907,46 @@ def test_empty():
...
+
+    assert td_lazy.stack_dim == nested_stack_dim
+    td_stack = torch.stack([td_lazy, td_lazy_clone], dim=stack_dim)
+    assert td_stack.stack_dim == stack_dim
+
+    dense_td_stack = dense_stack_tds(td_stack)
+    assert assert_allclose_td(
+        dense_td_stack, dense_stack_tds([td_lazy, td_lazy_clone], dim=stack_dim)
+    )
+    for i in [0, 1]:
```
vmoens, line 5940: range(2)?

test/test_tensordict.py
```diff
@@ -5906,6 +5907,46 @@ def test_empty():
...
 
+@pytest.mark.parametrize(
+    "stack_dim",
+    [0, 1, 2, 3],
+)
+@pytest.mark.parametrize(
+    "nested_stack_dim",
+    [0, 1, 2],
+)
+def test_dense_stack_tds(stack_dim, nested_stack_dim):
```
vmoens, line 5918: I don't see what we're testing specific to this method here. All the assertions would hold with a regular tensordict no?
Why aren't we testing anything specific to the keys we provide?

vmoens (commented):
Thanks a lot for this!
I think we can clarify it a bit, it isn't super transparent atm

tensordict/tensordict.py
matteobettini: because this method can be used even when no lazy stacks are involved

tensordict/tensordict.py
matteobettini: we need 2 levels of nesting because the function is used to stack tds that contain lazy stacks. so it is only useful when there are at least two levels of nesting. The change of the stack dim from 0 to 1 is the core difference here. "b" is just there to highlight that

test/test_tensordict.py
matteobettini: the idea is that this function should behave exactly the same with regualr tds or lazy stacks so the fact that a regualr td passes is ok to me. the main things we are testing is that when stacking tds that contain lazy stacks we can call this function to remove one level of lazyness and densify one stack_dim

matteobettini:
I have updated docstring, example and test.
Hopefully now it is more clear what this function does

vmoens (approved):
Got it sorry I'm a Bit slow

PR #512, opened 2023-08-18
[BugFix] Retrieve all keys for `LazyStackedTensorDict`

matteobettini:
The following was broken
```
    def test_all_keys(self):
        td = TensorDict({"a": torch.zeros(1)}, [])
        td2 = TensorDict({"a": torch.zeros(2)}, [])
        stack = torch.stack([td, td2])
        assert set(stack.keys(True, True)) == {"a"}
```


vmoens (approved):
LGTM

PR #515, opened 2023-08-30
[Feature] Module to add state independent normal scale

albertbou92:
## Description

For continuous action spaces, it is not uncommon to define normal policies with state independent variance. For example, the original PPO paper does so for the MuJoCo trainings. 

This PR proposes o add a module that handles this situation. 

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/tensordict/nn/distributions/continuous.py
+++ b/tensordict/nn/distributions/continuous.py
@@ -153,7 +153,7 @@ class AddStateIndependentNormalScale(torch.nn.Module):
         >>> tensor = torch.randn(4, 2, 3)
         >>> loc, scale, others = module_normal(*module(tensor))
         >>> print(loc.shape, scale.shape)
-        torch.Size([4, 2, 2]) torch.Size([4, 2, 2])
+        torch.Size([4, 2, 4]) torch.Size([4, 2, 4])
         >>> assert (scale > 0).all()
     """
 
--- a/tensordict/nn/distributions/continuous.py
+++ b/tensordict/nn/distributions/continuous.py
@@ -126,7 +126,7 @@ class NormalParamExtractor(nn.Module):
 
 
 class AddStateIndependentNormalScale(torch.nn.Module):
-    """A nn.Module that adds a trainable state-independent scale parameter.
+    """A nn.Module that adds a trainable state-independent scale parameters.
 
     The scale parameters are mapped onto positive values using the specified ``scale_mapping``.
 
--- a/tensordict/nn/distributions/continuous.py
+++ b/tensordict/nn/distributions/continuous.py
@@ -6,7 +6,7 @@
 from __future__ import annotations
 
 from numbers import Number
-from typing import Sequence
+from typing import Sequence, Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


tensordict/nn/distributions/continuous.py
```diff
@@ -120,6 +125,77 @@ def forward(self, *tensors: torch.Tensor) -> tuple[torch.Tensor, ...]:
...
+
+        # Reshape scale to match the number of dimensions in loc
+        input_dims = loc.dim()
+        scale_dims = self.state_independent_scale.dim()
+        scale_shape = self.state_independent_scale.shape
+        scale = self.state_independent_scale.view(
+            (1,) * (input_dims - scale_dims) + scale_shape
+        )
+
+        scale = torch.zeros(loc.size()).to(loc.device) + scale
```
vmoens, line 193:
```suggestion
        scale = torch.zeros_like(loc) + scale
```

tensordict/nn/distributions/continuous.py
```diff
@@ -120,6 +125,77 @@ def forward(self, *tensors: torch.Tensor) -> tuple[torch.Tensor, ...]:
...
+                f"in scale ({self.state_independent_scale.shape})"
+            )
+
+        # Reshape scale to match the number of dimensions in loc
+        input_dims = loc.dim()
+        scale_dims = self.state_independent_scale.dim()
+        scale_shape = self.state_independent_scale.shape
+        scale = self.state_independent_scale.view(
+            (1,) * (input_dims - scale_dims) + scale_shape
+        )
```
vmoens, line 191: that does not seem necessary to me, `torch.zeros_like(loc) + self.state_independent_scale` should work since it's broadcastable

vmoens (commented):
Nice feature!

Would it make sense to integrate that in NormalParamExtractor under a `trainable_bias` or argument in the constructor (named to be chosen)?
It seems like a lot of code is the same as NormalParamExtractor and the functionality would blend in nicely. Happy to hear your thoughts about this.

Tests are missing btw

albertbou92:
> Nice feature!
> 
> Would it make sense to integrate that in NormalParamExtractor under a `trainable_bias` or argument in the constructor (named to be chosen)? It seems like a lot of code is the same as NormalParamExtractor and the functionality would blend in nicely. Happy to hear your thoughts about this.
> 
> Tests are missing btw

vmoens (approved):
Let's add some tests and we're good to go!

test/test_nn.py
```diff
@@ -2897,6 +2901,26 @@ def test_tdparams_clone(self):
             assert (val == td.get(key)).all()
 
 
+class TestAddStateIndependentNormalScale:
+    def test_add_scale_basic(self, num_outputs=4):
+        module = nn.Linear(3, num_outputs)
```
vmoens, line 2906: Should this not be `2*num_outputs` ?

albertbou92: the model predicts only the means, the TestAddStateIndependentNormalScale module adds the scale with the same shape. So instead of 2*num_outputs, the linear layer predicts just num_outputs

vmoens:
Failures seem unrelated to this PR so I'm happy to merge it

PR #517, opened 2023-08-31
[Feature] CompositeDistribution

vmoens:
## Description

Introduces CompositeDistribution, which allows to build bags of distributions.

Addresses https://github.com/pytorch/rl/issues/1473

cc @hersh

### DIFF
--- a/tensordict/nn/distributions/composite.py
+++ b/tensordict/nn/distributions/composite.py
@@ -1,0 +1,133 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from tensordict import TensorDict, TensorDictBase
+from tensordict._tensordict import unravel_keys
+from tensordict.utils import NestedKey
+from torch import distributions as d
+
+
+class CompositeDistribution(d.Distribution):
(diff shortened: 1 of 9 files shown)


vmoens:
Note: we pass a structure `{"sample_name": {"params": stuff}}` to the constructor, which structure is similar to the samples that we'll get. But the value pointed by `"sample_name"` is a dict (or better, a tensordict) within the constructor, and it will be a tensor during sampling. This means that this will never be possible:
```
params = make_params(...)
dist = CompositeDistribution(params, ...)
sample = dist.sample()
params.update(sample) # breaks as tensor and tensordicts conflict
```

tensordict/nn/distributions/composite.py
```diff
@@ -0,0 +1,132 @@
...
+        if shape is None:
+            shape = torch.Size([])
+        return TensorDict(
+            {name: dist.rsample(shape) for name, dist in self.dists.items()},
+            shape + self.batch_shape,
+        )
+
+    def log_prob(self, sample: TensorDictBase):
+        d = {
+            _add_suffix(name, "_log_prob"): dist.log_prob(sample.get(name))
```
matteobettini, line 90: we do not want this keys to be parametric?

matteobettini (approved):
LGTM

PR #518, opened 2023-09-04
[Feature] TensorDict.map

### LINKED ISSUE #504: Is there a way to map a TensorDict?
I am currently using [HuggingFace Datasets](https://huggingface.co/docs/datasets/process) to load, process, and save some data. However, HF Datasets saves the data in the Arrow format and [wastes](https://github.com/huggingface/datasets/issues/6104) a lot of time converting between Arrow and PyTorch tensors.

I am wondering if I can use memory-mapped TensorDicts for this purpose?

How can I do a `map` on batches of TensorDict?

Looking through the tutorials, the nearest example I found was using a Dataloader with `collate_fn` as the map function:
```
DataLoader(
    tensor_dict, batch_size=batch_size, collate_fn=map_fn,
)
```

But this does not allow me to form a pipeline of `map` functions. I also don't know how to save and load the resulting dataset.

vmoens:
## Description

Implements TensorDict.map which dispatches an operation over chunks of a tensordict along a chosen dimension.
Closes #504

### DIFF
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -1485,6 +1485,56 @@ class TensorDictBase(MutableMapping):
         chunksize: int = None,
         pool: mp.Pool = None,
     ):
+        """Maps a function to splits of the tensordict across one dimension.
+
+        This method will apply a function to a tensordict instance by chunking
+        it in tensordicts of equal size and dispatching the operations over the
+        desired number of workers.
+
+        The function signature should be ``Callabe[[TensorDict], Union[TensorDict, Tensor]]``.
+        The output must support the :func:`torch.cat` operation. The function
+        must be serializable.
+
+        Args:
+            fn (callable): function to apply to the tensordict.
+                Signatures similar to ``Callabe[[TensorDict], Union[TensorDict, Tensor]]``
+                are supported.
+            dim (int, optional): the dim along which the tensordict will be chunked.
+            num_workers (int, optional): the number of workers. Exclusive with ``pool``.
+                If none is provided, the number of workers will be set to the
+                number of cpus available.
+            chunksize (int, optional): the number of chunks to split the tensordict
+                into. If none is provided, the number of chunks will equate the number
+                of workers. For very large tensordicts, such large chunks
+                may not fit in memory for the operation to be done and
+                more chunks may be needed to make the operation practically
+                doable.
+            pool (mp.Pool, optional): a multiprocess Pool instance to use
+                to execute the job. If none is provided, a pool will be created
+                within the ``map`` method.
+
+        Examples:
+            >>> import torch
+            >>> from tensordict import TensorDict
+            >>>
+            >>> def process_data(data):
+            ...     data.set("y", data.get("x") + 1)
+            ...     return data
+            >>> if __name__ == "__main__":
+            ...     data = TensorDict({"x": torch.zeros(1, 1_000_000)}, [1, 1_000_000]).memmap_()
+            ...     data = data.map(process_data, dim=1)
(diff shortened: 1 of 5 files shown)


tensordict/tensordict.py
```diff
@@ -1477,6 +1477,84 @@ def apply(
...
+        chunksize: int = None,
+        pool: mp.Pool = None,
+    ):
+        """Maps a function to splits of the tensordict across one dimension.
+
+        This method will apply a function to a tensordict instance by chunking
+        it in tensordicts of equal size and dispatching the operations over the
+        desired number of workers.
+
+        The function signature should be ``Callabe[[TensorDict], Union[TensorDict, Tensor]]``.
```
matteobettini, line 1494: why not writing the type hint then?

tensordict/tensordict.py
```diff
@@ -1477,6 +1477,84 @@ def apply(
...
+        dim_orig = dim
+        if dim < 0:
+            dim = self.ndim + dim
+        if dim < 0 or dim >= self.ndim:
+            raise ValueError(f"Got incompatible dimension {dim_orig}")
+
+        if chunksize is None:
+            chunksize = num_workers
+        self_split = self.chunk(num_workers, dim=dim)
+        chunksize = 1
```
matteobettini, line 1553: I dont't understand here.
you set the chunksize to 1 without having used it?

tensordict/tensordict.py
```diff
@@ -1477,6 +1477,84 @@ def apply(
...
+                return self.map(fn, dim=dim, chunksize=chunksize, pool=pool)
+        num_workers = pool._processes
+        dim_orig = dim
+        if dim < 0:
+            dim = self.ndim + dim
+        if dim < 0 or dim >= self.ndim:
+            raise ValueError(f"Got incompatible dimension {dim_orig}")
+
+        if chunksize is None:
+            chunksize = num_workers
```
matteobettini, line 1551: related to below comment

tensordict/tensordict.py
```diff
@@ -1477,6 +1477,84 @@ def apply(
...
+
+        Args:
+            fn (callable): function to apply to the tensordict.
+                Signatures similar to ``Callabe[[TensorDict], Union[TensorDict, Tensor]]``
+                are supported.
+            dim (int, optional): the dim along which the tensordict will be chunked.
+            num_workers (int, optional): the number of workers. Exclusive with ``pool``.
+                If none is provided, the number of workers will be set to the
+                number of cpus available.
+            chunksize (int, optional): the number of chunks to split the tensordict
```
matteobettini, line 1506: if it is the number of chucks it is not the chunksize right? isn't it the num_chuncks?

matteobettini (commented):
I left some questions.

tensordict/tensordict.py
vmoens: we can do that too! Good point

I think that docstrings >> type hint as people should look at the doc rather than the code to see what a method is made for and how to use it.
(as in: I don't destroy my microwave to see how to use it but I read the manual :) )
But you're right, we can do both

tensordict/tensordict.py
vmoens: how yeah `num_workers` -> `chunksize` on line 1552

tensordict/tensordict.py
matteobettini: yeah we can we all the options. 
the thing i was saying is that when something refers to the size of the chunks it should be called smth that contains "size" and when something refers to the number of chunks it should contain "num"

matteobettini (approved):
LGTM

PR #520, opened 2023-09-04
[Refactor] TensorDict.to() signature

vmoens:
## Description

Refactors the `TensorDict.to` to match more closely `torch.Tensor` and `torch.nn.Module`

This is bc-breaking as previously a single `dest` argument was accepted. Keyword arguments can now be used to map `device`, `dtype`, change the `batch_size`. An example `tensor` / tensordict (`other=...`) can be used to retrieve device and dtype.

### DIFF
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -700,12 +700,12 @@ MemmapTensor of shape {self.shape}."""
 
     def to(
         self,
-        *args, **kwargs,
+        *args,
+        **kwargs,
     ) -> torch.Tensor | MemmapTensor:
         """Maps a MemmapTensor to a given dtype or device.
 
         Args:
-
             device (torch.device, optional): Destination device when tensors
                 will be used.
             dtype (torch.dtype, optional): the desired floating point or complex dtype of
--- a/tensordict/memmap.py
+++ b/tensordict/memmap.py
@@ -700,34 +700,39 @@ MemmapTensor of shape {self.shape}."""
 
     def to(
         self,
-        dest: DeviceType | torch.dtype,
-        non_blocking: bool = False,
+        *args, **kwargs,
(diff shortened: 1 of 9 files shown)


tensordict/memmap.py
```diff
@@ -700,34 +700,39 @@ def __getstate__(self) -> dict[str, Any]:
...
 
         Args:
-            dest (device indicator or torch.dtype): where to cast the
-                MemmapTensor. For devices, this is a lazy operation
-                (as the data is stored on physical memory). For dtypes, the
-                tensor will be retrieved, mapped to the
-                desired dtype and cast to a new MemmapTensor.
-            non_blocking (bool, optional): no-op for MemmapTensors. Default: False.
+            device (torch.device, optional): Destination device when tensors
+                will be used.
```
matteobettini, line 710: This description is a bit unclear

tensordict/tensordict.py
```diff
@@ -2562,22 +2562,63 @@ def __torch_function__(
...
         Args:
-            dest (device, size or TensorDictBase subclass): destination of the
-                tensordict. If it is a torch.Size object, the batch_size
-                will be updated provided that it is compatible with the
-                stored tensors.
+            device (torch.device, optional): the desired device of the parameters
+                and buffers in this module
+            dtype (torch.dtype, optional): the desired floating point or complex dtype of
+                the parameters and buffers in this module
+            tensor (torch.Tensor, optional): Tensor whose dtype and device are the desired
```
matteobettini, line 2602: So these changes are applied only to parameters and buffers? Not to all tensors contained?

matteobettini (approved):
LGTM

tensordict/tensordict.py
vmoens: it's a copy paste from nn.Module, let me correct that

PR #522, opened 2023-09-06
[Refactor] Refactor  to create empty tds where necessary

vmoens:
## Description

- Ensures that empty tensordicts are created for parameter-free modules (eg., nn.Tanh)
- Add a `as_module` arg that allows to get `TensorDIctParams` object that can be used for functional calls as well as storing params in a nn.Module.


matteobettini (approved):
LGTM

PR #523, opened 2023-09-07
[Versioning] M1 compatibility

vmoens:


.github/workflows/build-wheels-m1.yml
```diff
@@ -0,0 +1,45 @@
...
+        # NOTE: Binary build pipelines should only get triggered on release candidate builds
+        # Release candidate tags look like: v1.11.0-rc1
+        - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+
+  workflow_dispatch:
+
+jobs:
+  generate-matrix:
+    strategy:
+      matrix:
+        python_version: ["3.8", "3.9", "3.10", "3.11"]
```
DanilBaibak, line 20: Let's remove this part. The idea is to use only one Python version to avoid wasting resources on every PR.

.github/workflows/build-wheels-m1.yml
```diff
@@ -0,0 +1,45 @@
...
+    uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
+    with:
+      package-type: wheel
+      os: macos-arm64
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: main
+  build:
+    needs: generate-matrix
+    name: pytorch-labs/tensordict
```
DanilBaibak, line 29: Should be "pytorch/tensordic"

.github/workflows/build-wheels-m1.yml
```diff
@@ -0,0 +1,45 @@
...
+    uses: pytorch/test-infra/.github/workflows/build_wheels_macos.yml@main
+    with:
+      repository: pytorch-labs/tensordict
```
DanilBaibak, line 32: Here is also "pytorch/tensordic"

.github/workflows/nightly_build.yml
```diff
@@ -99,6 +99,37 @@ jobs:
           name: tensordict-mac-${{ matrix.python_version[0] }}.whl
           path: dist/*.whl
 
+  build-wheel-m1:
+    if: github.repository_owner == 'pytorch-labs'
```
DanilBaibak, line 103: Should it be "repository_owner" as pytorch?

.github/workflows/build-wheels-m1.yml
vmoens: Got it thanks! So this is a test that wheels can be built, but how do we actually build them for all python versions, then?

PR #525, opened 2023-09-12
[Feature] First class dim compatibility

vmoens:
## Description

WIP to make TensorDict compatible with FCD.

Example usage:
```python
import torch
from functorch import dim

t = torch.randn(3, 4, 5)
td = TensorDict({"t": t}, [3, 4])
d0, d1 = dim.dims(2)

std = td[d0]
print('[d0]', std)

# [d0] TensorDict(
#     fields={
#         t: Tensor(shape=torch.Size([4, 5]), device=cpu, dtype=torch.float32, is_shared=None)},
#     batch_size=torch.Size([4]),
#     device=None,
#     is_shared=False)

std = td[d0, d1]
print('[d0, d1]', std)

# [d0, d1] TensorDict(
#     fields={
#         t: Tensor(shape=torch.Size([5]), device=cpu, dtype=torch.float32, is_shared=None)},
#     batch_size=torch.Size([]),
#     device=None,
#     is_shared=False)

std = td[:, d1]
print('[:, d1]', std)

# [:, d1] TensorDict(
#     fields={
#         t: Tensor(shape=torch.Size([3, 5]), device=cpu, dtype=torch.float32, is_shared=None)},
#     batch_size=torch.Size([3]),
#     device=None,
#     is_shared=False)
```
This can be used with modules to batch operations across sets of parameters seamlessly. We extract the parameters, 
```

from torch import nn

net = nn.Sequential(
    nn.Linear(3, 4),
    nn.Tanh(),
    nn.Linear(4, 4),
)
params = TensorDict.from_module(net)
params = params.expand(10).clone()  # a TensorDict of shape [10]
d0 = dim.dims(1)
params = params[d0]

params.to_module(net)  # replace the params

y = net(torch.randn(11, 3))
print(y)
```

`to_module` is a draft of what we could use.

The idea of this example is to avoid functional calls through tensordict and functorch.dim when working with model ensembles.

cc @ezyang @zou3519 @zdevito
@matteobettini for MARL MLP and ConvNets
@smorad for model ensembles

### DIFF
--- a/tensordict/nn/params.py
+++ b/tensordict/nn/params.py
@@ -13,6 +13,7 @@ from functools import wraps
 from typing import Any, Callable, Iterator, Sequence
 
 import torch
+from functorch import dim as ftdim
 
 from tensordict import TensorDictBase
 from tensordict.nn.utils import Buffer
@@ -72,7 +73,7 @@ def _get_args_dict(func, args, kwargs):
 
 def _maybe_make_param(tensor):
     if (
-        isinstance(tensor, Tensor)
+        isinstance(tensor, (Tensor, ftdim.Tensor))
         and not isinstance(tensor, nn.Parameter)
         and tensor.dtype in (torch.float, torch.double, torch.half)
     ):
@@ -82,7 +83,7 @@ def _maybe_make_param(tensor):
 
 def _maybe_make_param_or_buffer(tensor):
     if (
-        isinstance(tensor, Tensor)
+        isinstance(tensor, (Tensor, ftdim.Tensor))
         and not isinstance(tensor, nn.Parameter)
         and tensor.dtype in (torch.float, torch.double, torch.half)
     ):
@@ -319,7 +320,7 @@ class TensorDictParams(TensorDictBase, nn.Module):
         if kwargs is None:
             kwargs = {}
         if func not in TDPARAM_HANDLED_FUNCTIONS or not all(
-            issubclass(t, (Tensor, TensorDictBase)) for t in types
+            issubclass(t, (Tensor, ftdim.Tensor, TensorDictBase)) for t in types
         ):
             return NotImplemented
         return TDPARAM_HANDLED_FUNCTIONS[func](*args, **kwargs)
--- a/tensordict/tensordict.py
+++ b/tensordict/tensordict.py
@@ -32,14 +32,13 @@ from typing import (
     TypeVar,
     Union,
 )
-from functorch import dim
 
 from warnings import warn
-from functorch import dim as ftdim
 
 import numpy as np
 
 import torch
+from functorch import dim as ftdim
 from tensordict._tensordict import _unravel_key_to_tuple
 from tensordict.memmap import memmap_tensor_as_tensor, MemmapTensor
 from tensordict.utils import (
@@ -72,7 +71,7 @@ from tensordict.utils import (
     NestedKey,
     prod,
 )
-from torch import distributed as dist, multiprocessing as mp, Tensor
+from torch import distributed as dist, multiprocessing as mp, nn, Tensor
 from torch.utils._pytree import tree_map
 
 try:
@@ -412,12 +411,22 @@ class TensorDictBase(MutableMapping):
         return td
 
     def to_module(self, module):
-        children = dict(module.named_children())
+        from tensordict.nn.functional_modules import set_tensor_dict
+
+        __base__setattr__ = nn.Module.__setattr__
+        # we use __dict__ directly to avoid the getattr/setattr overhead whenever we can
+        __dict__ = module.__dict__
+
         for key, value in self.items():
-            if _is_tensor_collection(type(value)):
-                value.to_module(children[key])
+            cls = value.__class__
+            if _is_tensor_collection(cls) or issubclass(cls, dict):
+                value.to_module(__dict__["_modules"][key])
             else:
-                module._parameters[key] = value
+                if module.__class__.__setattr__ is __base__setattr__:
+                    set_tensor_dict(__dict__, module, key, value)
+                else:
+                    # use specialized __setattr__ if needed
+                    setattr(module, key, value)
 
     @property
     def shape(self) -> torch.Size:
@@ -3778,6 +3787,7 @@ _ACCEPTED_CLASSES = [
     Tensor,
     MemmapTensor,
     TensorDictBase,
+    ftdim.Tensor,
 ]
 if _has_torchrec:
     _ACCEPTED_CLASSES += [KeyedJaggedTensor]
@@ -6401,7 +6411,13 @@ class LazyStackedTensorDict(TensorDictBase):
                 continue
             if cursor == self.stack_dim:
                 # we need to check which tds need to be indexed
-                if isinstance(idx, slice) or _is_number(idx):
+                if isinstance(idx, ftdim.Dim):
+                    raise ValueError(
+                        "Cannot index a lazy stacked tensordict along the stack dimension with "
+                        "a first-class dimension index. Consider consolidating the tensordict first "
+                        "using `tensordict.contiguous()`."
+                    )
+                elif isinstance(idx, slice) or _is_number(idx):
                     selected_td_idx = range(len(self.tensordicts))[idx]
                     if not isinstance(selected_td_idx, range):
                         isinteger = True
@@ -6433,6 +6449,7 @@ class LazyStackedTensorDict(TensorDictBase):
                     idx,
                     (
                         int,
+                        ftdim.Dim,
                         slice,
                         list,
                         range,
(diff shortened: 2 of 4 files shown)


ezyang:
Cc @zdevito

ezyang:
Nice! What kind of feedback are you looking for?

vmoens:
For now nothing, it's just an FYI. 
There seem to be some issue regarding compatibility with nn.Parameter and also it's a bit surprising that it doesn't subclass Tensor but I'll make a more comprehensive list of advantages and drawbacks when the PR is ready.

vmoens:
This is ready for review if anyone wants to give it a shot.

One of my goals with this is to do what we did in the example:
- Extract params from a model (or several models) to create a mixture of experts
- replace this stack of params within the module to avoid functional calls.
Basically, I was hoping I could have a regular `nn.Module` with batched params that would act as a mixture of expect without loops, functional calls or direct call to vmap. I love these features but I hoped I could package them in through tensordict in a way that made the end user less reliant on these primitives.

The issues I see currently are:
- It is not very clear how that should interact with nn.Parameter. `nn.Parameter(tensor[dim])` fails because `tensor[dim]` is not a tensor anymore. Maybe we should do `nn.Parameter(tensor)[dim]` but then we replace in the module something that is not a parameter anymore. Some module may fail with that (eg, RNNs but they will fail for other reasons too, like CuDNN compatibility etc).
- For instance, this currently works:
```python
from functorch import dim as ftdim

module = nn.Sequential(nn.Linear(3, 4), nn.Linear(4, 1))
# create batched params
params = TensorDict.from_module(module)
params = params.expand(3).clone().apply(lambda x: x.requires_grad_())
# FCD indexing
d0 = ftdim.dims(1)
# execute grad
x = torch.randn(3)
def func(params):
    params_batched = params[d0]
    params_batched.to_module(module)
    return module(x)._tensor.sum()
g = torch.func.grad(func)(params)
print(g)
print("params", params['0', 'weight'], "\ngrads", g['0', 'weight'])
```
which is cool. But we must set the parameters within the call to `func` which introduces some overhead. Also it isn't very pytorch-y. What I understood from my conversations with @ezyang was that FCD promise was to hide away some dimensions of a tensor while keeping its behaviour as natural as possible. Hence, my (unreasonable) expectation was that I could simply do
```python
module = nn.Sequential(nn.Linear(3, 4), nn.Linear(4, 1))
params = TensorDict.from_module(module)
# create a batch of params
params = params.expand(3).clone().apply(lambda x: nn.Parameter(x))
# reduce the first dim
d0 = dim.dims(1)
params = params[0]
# repopulate module
params.to_module(module)
# run module
loss = loss_fn(module(x), target)
# backprop
loss.backward()
```
But because my tensordict is not full of tensors but `dim.Tensor` and because backward does not work with tensors that are build within functorch, this cannot currently work.

Having to modify that logic makes FCD less attractive -- I might as well use plain vmap + functional calls if I have to repopulate my module at every call, which I believe will be faster to execute. 

Do you guys have any thought on that?

zdevito:
The semantics for adding first-class dimensions to tensordict make sense to me. It also makes sense to be able to install parameters that have first-class dimensions.  I am less clear on how we would accomplish it with how parameters in modules currently exist. First-class dims are implemented as their own objects that are not tensor subclasses in order to run fast enough in eager mode without overhead. I think parameters need to be real tensors in a lot of apis. In particular trying to set `.grad` properties on first-class tensors probably doesn't work. I wonder if there is a way to install the underlying raw tensor that is wrapped by the first-class dim tensor as the parameter, and then overload the property access for that tensor with something that constructs the tensor with first-class dims on the fly. This could be installed in the `to_module` call. I don't thing this is possible today but maybe @ezyang knows more.

ezyang:
So... we could relax the constraint here:

```
# Metaclass to combine _TensorMeta and the instance check override for Parameter.
class _ParameterMeta(torch._C._TensorMeta):
    # Make `isinstance(t, Parameter)` return True for custom tensor instances that have the _is_param flag.
    def __instancecheck__(self, instance):
        return super().__instancecheck__(instance) or (
            isinstance(instance, torch.Tensor) and getattr(instance, '_is_param', False))
```

and say that it has to either be a torch.Tensor, or it can be one of the first-class dim objects. Is that enough? Might be!

ezyang:
Actually, I read the example more carefully (and fixed up some stuff) and I see:

```
import torch
import torch.nn as nn
import functorch.dim as dim

module = nn.Sequential(nn.Linear(3, 4), nn.Linear(4, 1))
params = TensorDict.from_module(module)
# create a batch of params
params = params.expand(3).clone().apply(lambda x: nn.Parameter(x))
# reduce the first dim
d0 = dim.dims(1)
params = params[d0]
# repopulate module
params.to_module(module)
# run module
loss = module(torch.randn(3, 3)).sum()
# backprop
loss.backward()
```

fails with

```
RuntimeError: backward() called inside a functorch transform. This is not supported, please use functorch.grad or functorch.vjp instead or call backward() outside of functorch transforms.
```

Back in the day, @zou3519 talked a lot about the potential of non-lexical functorch transforms, one of the use cases being this kind of backward() call. I think we concluded that in principle it would be possible, but we have to implement it (which we never found time to do.)

So drat, I guess you can't actually replace the auto-batching behavior from tensordict with first class dims, what a bummer.

vmoens:
Thanks @ezyang and @zdevito for looking into this!
I don't think "wrapping" the dim.Tensor within an `nn.Parameter` is crucial for a first prototype.
Based on #526 I think that the model ensemble API will initially look like this:
```python
import torch
import torch.nn as nn
import functorch.dim as dim

module = nn.Sequential(nn.Linear(3, 4), nn.Linear(4, 1))
params = TensorDict.from_module(module)
# create a batch of params
num_models = 5
params = params.expand(num_models).clone().apply(lambda x: nn.Parameter(x))
x = torch.randn(3)

def compute(params_vals):
    # reduce the first dim
    d0 = dim.dims(1)
    for key, val_source in zip(params.keys(include_nested=True, leaves_only=True), params_vals):
        params.set(key, val_source)
    params_dims = params[d0]
    # no vmap
    y = torch.func.functional_call(module, params_dims, (x,))
    return y._tensor.sum() # or any other loss

grads = torch.func.grad(compute)(list(params.values(True, True)))
# put grads in a tensordict for clarity
grads = TensorDict(
    {key: grad for key, grad in zip(params.keys(include_nested=True, leaves_only=True), grads)}, batch_size=params.batch_size
)
print("grads of our 5 models", grads)
```
which gives you
```
grads of our 5 models TensorDict(
    fields={
        0: TensorDict(
            fields={
                bias: Tensor(shape=torch.Size([5, 4]), device=cpu, dtype=torch.float32, is_shared=False),
                weight: Tensor(shape=torch.Size([5, 4, 3]), device=cpu, dtype=torch.float32, is_shared=False)},
            batch_size=torch.Size([5]),
            device=None,
            is_shared=False),
        1: TensorDict(
            fields={
                bias: Tensor(shape=torch.Size([5, 1]), device=cpu, dtype=torch.float32, is_shared=False),
                weight: Tensor(shape=torch.Size([5, 1, 4]), device=cpu, dtype=torch.float32, is_shared=False)},
            batch_size=torch.Size([5]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([5]),
    device=None,
    is_shared=False)
```
ie, you can get your gradients using fcd and no apparent vmap

Note that this requires some modif in functorch and torch nn (see #526 which is based on this PR).

I think that this is pretty cool, what is happening is pretty apparent. Not as easy as substituting the params with their fcd counterpart but with some woodwork this could look pretty nice already. For instance, I already made `vmap` compatible with TensorDict, making the same work for `grad` and similar functorch functions would bring some value. Ideally we'd like to simplify the above to
```python
# pass directly the params
def compute(params):
    d0 = dim.dims(1)
    params_dims = params[d0]
    y = torch.func.functional_call(module, params_dims, (x,))
    return y._tensor.sum() # or any other loss

# pass directly the params
grads = torch.func.grad(compute)(params)
# grads are already in a tensordict
print("grads of our 5 models", grads)
```

ezyang:
Oh this is the thing where you wanted to override how functorch detects batching on inputs passed to grad. @zou3519 I don't remember what your objection to this was?

zou3519:
I don't think I have an objection? The code in https://github.com/pytorch-labs/tensordict/pull/525#issuecomment-1726077112 looks reasonable to me

ezyang:
@zou3519 Specifically referring to https://github.com/pytorch-labs/tensordict/pull/526/files#diff-b60cca91a0a60483cc294cdb1ae495bac71dbdb25d06e008443a41873e2a1891

vmoens:
> @zou3519 Specifically referring to https://github.com/pytorch-labs/tensordict/pull/526/files#diff-b60cca91a0a60483cc294cdb1ae495bac71dbdb25d06e008443a41873e2a1891

That and [this](https://github.com/pytorch/tensordict/blob/accd8a4a31ec749f52e75a87a875424652069163/tensordict/nn/functional_modules.py#L223-L306) too, which I think is more contentious. 
The pitch of this feature is this: 
We could just register TensorDict in PyTree and call it a day.
However, if we just "pytree" over a tensordict (ie we register it in pytree, deconstruct it and reconstruct it) we lose the ability to do this:
```python
def func(tensordict_or_tensor):
    assert tensordict_or_tensor.shape == (1, 2, 4)
    return tensordict_or_tensor
tensor = torch.randn(1, 2, 3, 4)
vmap(func, (2,))(tensor)
tensordict = TensorDict({}, batch_size=[1, 2, 3, 4])
vmap(func, (2,))(tensordict)
```
Currently, with the monkey patch, this code runs. If we just register tensordict within pytree but do not patch, it will fail. 
This is because params will be deconstructed, a batch-dimension will be added and then the tensordict will be reconstructed but then the batch-size will either mismatch or be lost.
For this, we need some ad-hoc mechanism to keep the batch-size of the tensordict consistent through vmap, hence the ad-hoc monkey-patch in tensordict.nn.

Now, I totally get that this has ramifications that go beyond this (I guess we don't want to overcharge vmap with every new class that comes into sight). To me, any solution that makes vmap possible with tensordict (ie, a tensordict passed through vmap sees its shape changed to a consistent shape within the vmap call) is great.

matteobettini (approved):
LGTM

PR #528, opened 2023-09-14
[BugFix] Fix state-dict

vmoens:
## Description

Fixes calls to state-dict to make them compatible with nn.Module.

Introduces related tests and docstring.

cc @matteobettini

### DIFF
--- a/docs/source/saving.rst
+++ b/docs/source/saving.rst
@@ -79,17 +79,17 @@ that we will re-populate with the saved data.
 Again, two lines of code are sufficient to save the data:
 
   >>> app_state = {
-  ...     "state": torchsnapshot.StateDict(tensordict=tensordict_source.state_dict())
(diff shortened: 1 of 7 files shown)


matteobettini (approved):
LGTM

maybe one test where the loaded data is not all zeros just to make sure

matteobettini:
we also may one some tests in torchrl to test the non regression of saving the state dict of loss modules

vmoens:
> LGTM
> 
> maybe one test where the loaded data is not all zeros just to make sure

We load onto zeroed data no?

vmoens:
> we also may want some tests in torchrl to test the non regression of saving the state dict of loss modules

on it already

vmoens:
> I meant something like getting the params from a linear module, saving, and reloading and checking against the original params

Ah but we do that, what is the problem with doing it with zero?

PR #539, opened 2023-10-09
[Feature] Add `pad` argument to TensorDict.where

vmoens:
## Description

Adds a `pad` argument in `TensorDict.where` to allow the method to be called on partially matching tensordicts.
The `pad` value will be used for any key that is present in one but not the other tensordict.


tensordict/tensordict.py
```diff
@@ -4802,38 +4808,87 @@ def to(tensor):
...
+                result._set_str(key, val, inplace=False, validated=True)
+                other_keys.discard(key)
+            for key in other_keys:
+                tensor = None
+                _other = other._get_str(key, default=NO_DEFAULT)
+                if _is_tensor_collection(type(_other)):
+                    try:
+                        tensor = _other.empty()
+                    except NotImplementedError:
+                        # H5 tensordicts do not support select()
```
matteobettini, line 4864: what is H5?

tensordict/tensordict.py
```diff
@@ -4802,38 +4808,87 @@ def to(tensor):
...
+                        raise KeyError(
+                            f"Key {key} not found and no pad value provided."
+                        )
+                    cond = expand_as_right(~condition, tensor)
+                elif _other is None:
+                    if pad is not None:
+                        _other = pad
+                    else:
+                        raise KeyError(
+                            f"Key {key} not found and no pad value provided."
```
matteobettini, line 4829: Before this PR was an error thrown for these cases? What was the treatment for missing keys?

matteobettini (commented):
LGTM

what was the treatment of missing keys before?

To me it could make sense that someone calls` self.where` with an `other` that has less keys and the `where` is only applied to the shared keys

tensordict/tensordict.py
vmoens: H5DF

tensordict/tensordict.py
vmoens: it was a bug in the sense that `td0.where(cond, td1)` and `td1.where(~cond, td0)` where not returning the same thing. Only the keys of the first (self) were scanned. Keys in `self` but not `other` would raise an exception, but keys in `other` but not `self` where just ignored. This is super bad IMO

PR #542, opened 2023-10-10
[BugFix] `torch.where`: Expand mask on the right in lazy stack tds

vmoens:
Masks in `torch.where` for lazy stacks need to have as many dims as the tensordict to be unbound along the stack dim. This PR makes sure that it's the case.


matteobettini (approved):
LGTM

PR #543, opened 2023-10-11
[BugFix] Better tensor allocation in memmap_like

vmoens:
The memmap_like function does not use `_set_str`, therefore a `to(...)` can be called on memmap tensors, leading them to be regular tensors.
This PR solves this issue


matteobettini (approved):
LGTM

PR #549, opened 2023-10-30
[Refactor] Graceful as_tensor

vmoens:


tensordict/tensordict.py
```diff
@@ -1811,13 +1811,12 @@ def as_tensor(self):
...
-            raise AttributeError(
-                f"{self.__class__.__name__} does not have an 'as_tensor' method "
-                f"because at least one of its tensors does not support this method."
-            ) from err
+        def as_tensor(x):
+            try:
+                return x.as_tensor()
+            except AttributeError:
+                return x
+        return self._fast_apply(as_tensor)
```
shagunsodhani, line 1819: So in this version, no `AttributeError` will be thrown ?

vmoens: The idea is that we want to deprecate this method progressively as MemmapTensor will eventually be a regular tensor thanks to @mikaylagawarecki's work!
Since we call as_tensor here and there and the backend will change, the only way to make this non bc breaking is to bypass errors in as_tensor

shagunsodhani: Thank you for the clarification!

PR #550, opened 2023-11-02
[Refactor] Faster instantiation

vmoens:


vmoens:
woohoo! +112% on creation speed!
@albertbou92 @matteobettini @BY571 @skandermolla
In practice, for torchrl it's a >10% further speedup on data collection speed (but also impacts all other modules)
Can anyone review this?

albertbou92:
I see, so defining directly the attributes as class-level attributes is faster at runtime than doing attribute assignment in the __new__ method, is that it?

vmoens:
Yes that and removing slots

albertbou92 (approved):

albertbou92:
> Yes that and removing slots

looking forward to see how it affects torchrl speed! :)

PR #562, opened 2023-11-21
[Feature] Seed workers in TensorDict.map

vmoens:
## Description

Seeds the workers in TensorDict.map pool.
The base seed can be passed to the method.
If no seed is required, the unseeded pool can be passed to the method.

The seed is aimed at having a separate seed for each worker and NOT to have reproducible results. By nature, pool's workers will pick up job on the basis of their status, hence the process is non-deterministic by nature.

The tests only verify that with torch and numpy, each worker has a predictable behaviour if the number of jobs equates the number of workers.

cc @NicolasHug

### DIFF
--- a/tensordict/base.py
+++ b/tensordict/base.py
@@ -3018,7 +3018,7 @@ class TensorDictBase(MutableMapping):
             for i in range(num_workers):
                 queue.put(i)
             with mp.Pool(
-                num_workers,
+                processes=num_workers,
                 initializer=_proc_init,
                 initargs=(seed, queue),
                 maxtasksperchild=max_tasks_per_child,
--- a/tensordict/base.py
+++ b/tensordict/base.py
@@ -3014,7 +3014,7 @@ class TensorDictBase(MutableMapping):
                 torch.empty((), dtype=torch.int64).random_(generator=generator).item()
             )
 
-            queue = mp.Queue(maxsize=num_workers)
+            queue = mp.Queue(maxsize=num_workers * 2)
             for i in range(num_workers * 2):
                 queue.put(i)
             with mp.Pool(
--- a/tensordict/base.py
+++ b/tensordict/base.py
@@ -2914,11 +2914,12 @@ class TensorDictBase(MutableMapping):
         self,
         fn: Callable,
         dim: int = 0,
-        num_workers: int = None,
-        chunksize: int = None,
-        num_chunks: int = None,
-        pool: mp.Pool = None,
-        seed: int = None,
+        num_workers: int | None = None,
+        chunksize: int | None = None,
+        num_chunks: int | None = None,
+        pool: mp.Pool | None = None,
+        seed: int | None = None,
+        maxtasksperchild: int | None = None,
     ):
         """Maps a function to splits of the tensordict across one dimension.
 
@@ -2975,6 +2976,9 @@ class TensorDictBase(MutableMapping):
                   know which worker will pick which job. However, we can make sure
                   that each worker has a different seed and that the pseudo-random
                   operations on each will be uncorrelated.
+            maxtasksperchild (int, optional): the maximum number of jobs picked
+                by every child process. Defaults to ``None``, i.e., no restriction
+                on the number of jobs.
 
         Examples:
             >>> import torch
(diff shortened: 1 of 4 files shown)


tensordict/base.py
```diff
@@ -2941,16 +2943,38 @@ def map(
...
                 into. If none is provided, the number of chunks will equate the number
                 of workers. For very large tensordicts, such large chunks
                 may not fit in memory for the operation to be done and
                 more chunks may be needed to make the operation practically
-                doable. This argument is exclusive with chunksize.
+                doable. This argument is exclusive with ``chunksize``.
             pool (mp.Pool, optional): a multiprocess Pool instance to use
                 to execute the job. If none is provided, a pool will be created
                 within the ``map`` method.
+            seed (integer, optional): the initial seed of the pool. Each member
```
NicolasHug, line 2956: `DataLoader` doesn't take a seed as input but rather[ `Generator` object](https://github.com/pytorch/pytorch/blob/8ec59d3553db4e3dae991db9aa3251558265dfd3/torch/utils/data/dataloader.py#L169C1-L171C59), which defaults to the pytorch default RNG. The `base_seed` is then [drawn internally](https://github.com/pytorch/pytorch/blob/8ec59d3553db4e3dae991db9aa3251558265dfd3/torch/utils/data/dataloader.py#L602C6-L602C6) from that `Generator`

tensordict/base.py
```diff
@@ -2941,16 +2943,38 @@ def map(
...
+                and passed to :meth:`map` directly.
+                .. note::
+                  Caution should be taken when providing a low-valued seed as
+                  this can cause autocorrelation between experiments, example:
+                  if 8 workers are asked and the seed is 4, the workers seed will
+                  range from 4 to 11. If the seed is 5, the workers seed will range
+                  from 5 to 12. These two experiments will have an overlap of 7
+                  seeds, which can have unexpected effects on the results.
+
+                .. note::
```
NicolasHug, line 2970: That kind of determinism should be enough - for users who want a more fine-grained control, `map()` could expose `worker_init_fn` like in Pytorch?

tensordict/utils.py
```diff
@@ -1721,3 +1721,12 @@ def _legacy_lazy(func):
...
     return func
+
+
+# Process initializer for map
+def _proc_init(base_seed, queue):
+    worker_id = queue.get()
+    seed = base_seed + worker_id
+    torch.manual_seed(seed)
+    np_seed = _generate_state(base_seed, worker_id)
+    np.random.seed(np_seed)
```
NicolasHug, line 1732: *If* `worker_init_fn` is exposed, then this seeding should happen *before* `worker_init_fn` gets called, so that users can override the seeds that are being used. (See https://github.com/pytorch/pytorch/blob/8ec59d3553db4e3dae991db9aa3251558265dfd3/torch/utils/data/_utils/worker.py#L250C35-L250C35)

`worker_init_fn` just takes `worker_id` as input, but users can call [`get_worker_info`](https://pytorch.org/docs/stable/data.html#torch.utils.data.get_worker_info) to retrieve more stuff like the seed they were seeded with (not sure how that would translate for TensorDict though).

NicolasHug (commented):
Just sharing pointers to the equivalent APIs in the pytorch Dataloader for ref, in case you want to align the APIs

tensordict/base.py
vmoens: thanks I will edit that!

tensordict/base.py
vmoens: We could do that, otherwise the user could just past their own Pool (which is roughly equivalent).
The issue we have is that the initializer and its args are shared by all workers of the pool. The way we go around that in this PR is to create a queue that contains all the worker ids, and each worker gets its id from the queue.

tensordict/base.py
```diff
@@ -2914,11 +2914,12 @@ def map(
...
-        chunksize: int = None,
-        num_chunks: int = None,
-        pool: mp.Pool = None,
-        seed: int = None,
+        num_workers: int | None = None,
+        chunksize: int | None = None,
+        num_chunks: int | None = None,
+        pool: mp.Pool | None = None,
+        seed: int | None = None,
+        maxtasksperchild: int | None = None,
```
shagunsodhani, line 2922: nit: but maybe max_tasks_per_child will be easier to read

vmoens: I know! But it's what mp.Pool is using so I wanted to be consistent. Readable inputs are probably less harmful though

shagunsodhani: Thanks for the clarification!

vmoens: No worry I'll edit it!

PR #582, opened 2023-11-28
[Feature] Saving metadata of tensorclass

vmoens:
A simple script to test the feature

```python
import tempfile
from tensordict import tensorclass, TensorDict
import torch
import os

def print_directory_tree(path, indent=""):
    """
    Print the directory tree starting from the specified path.

    Parameters:
    - path (str): The path of the directory to print.
    - indent (str): The current indentation level for formatting.
    """
    if os.path.isdir(path):
        print(indent + os.path.basename(path) + "/")
        indent += "    "
        for item in os.listdir(path):
            print_directory_tree(os.path.join(path, item), indent)
    else:
        print(indent + os.path.basename(path))

@tensorclass
class MyClass:
    X: torch.Tensor
    td: TensorDict
    integer: int
    string: str
    dictionary: dict

@tensorclass
class MyOtherClass:
    Y: torch.Tensor

data = MyClass(
    X = torch.randn(10, 3),
    td=TensorDict({"y": torch.randn(10)}, batch_size=[10]),
    integer=3,
    string="a string",
    dictionary={"some_data": "a"},
    batch_size=[],
)

with tempfile.TemporaryDirectory() as tmpdir:
    print(tmpdir)
    data.memmap_(tmpdir)
    print_directory_tree(tmpdir)
    data2 = MyClass.load_memmap(tmpdir)
    print(data2)

    # the original class can be recovered
    data3 = MyOtherClass.load_memmap(tmpdir)
    assert isinstance(data3, MyClass)

```

cc @janeyx99

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -1532,7 +1532,7 @@ class TensorDict(TensorDictBase):
         def load_metadata(filepath):
             with open(filepath) as json_metadata:
                 metadata = json.load(json_metadata)
-                if metadata['_type'] != cls.__name__:
+                if metadata["_type"] != cls.__name__:
                     # return early to load from another cls
                     return metadata
                 if metadata["device"] == "None":
@@ -1548,6 +1548,7 @@ class TensorDict(TensorDictBase):
         type_name = metadata["_type"]
         if type_name != cls.__name__:
             import tensordict
+
             for other_cls in tensordict.base._ACCEPTED_CLASSES:
                 if other_cls.__name__ == type_name:
                     return other_cls.load_memmap(prefix)
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -800,7 +800,7 @@ class TensorDict(TensorDictBase):
                 {
                     "shape": list(data.shape),
                     "device": str(data.device),
-                    "_type": data.__class__.__name__,
+                    "_type": str(data.__class__),
                 }
             )
             with open(filepath, "w") as json_metadata:
@@ -1466,7 +1466,7 @@ class TensorDict(TensorDictBase):
                 {
                     "shape": list(data.shape),
                     "device": str(data.device),
-                    "_type": data.__class__.__name__,
+                    "_type": str(data.__class__),
                 }
             )
             with open(filepath, "w") as json_metadata:
@@ -1532,7 +1532,7 @@ class TensorDict(TensorDictBase):
         def load_metadata(filepath):
             with open(filepath) as json_metadata:
                 metadata = json.load(json_metadata)
-                if metadata["_type"] != cls.__name__:
+                if metadata["_type"] != str(cls):
                     # return early to load from another cls
                     return metadata
                 if metadata["device"] == "None":
@@ -1546,11 +1546,11 @@ class TensorDict(TensorDictBase):
 
         metadata = load_metadata(prefix / "meta.json")
         type_name = metadata["_type"]
-        if type_name != cls.__name__:
+        if type_name != str(cls):
             import tensordict
 
             for other_cls in tensordict.base._ACCEPTED_CLASSES:
-                if other_cls.__name__ == type_name:
+                if str(other_cls) == type_name:
                     return other_cls.load_memmap(prefix)
             else:
                 raise RuntimeError(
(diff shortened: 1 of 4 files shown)


vmoens:
Another script for optimizer state_dict:
```python
import os
import tempfile
from typing import Any

import torch

from tensordict import tensorclass, TensorDict

model = torch.nn.Linear(3, 4)
optim = torch.optim.Adam(list(model.parameters()), lr=1e-2, amsgrad=True)
model(torch.randn(3)).sum().backward()
optim.step()
optim.zero_grad()

sd = optim.state_dict()
print(sd)


def print_directory_tree(path, indent=""):
    """
    Print the directory tree starting from the specified path.

    Parameters:
    - path (str): The path of the directory to print.
    - indent (str): The current indentation level for formatting.
    """
    if os.path.isdir(path):
        print(indent + os.path.basename(path) + "/")
        indent += "    "
        for item in os.listdir(path):
            print_directory_tree(os.path.join(path, item), indent)
    else:
        print(indent + os.path.basename(path))


@tensorclass
class ParamGroup:
    params: list
    lr: float
    betas: tuple
    eps: float
    weight_decay: float
    amsgrad: bool
    maximize: bool
    foreach: Any
    capturable: bool
    differentiable: bool
    fused: Any


tc = ParamGroup(**sd['param_groups'][0], batch_size=[])
print(tc)
td = TensorDict(
    {"state": {str(key): val for key, val in sd["state"].items()},
     "metadata": tc}, batch_size=[]
)
print(td)

with tempfile.TemporaryDirectory() as tempdir:
    td.memmap_(tempdir)
    print_directory_tree(tempdir)
    print("loaded", TensorDict.load_memmap(tempdir))
```

This produces a tensordict with the following structure
```
TensorDict(
    fields={
        metadata: ParamGroup(
            params=[0, 1],
            lr=0.01,
            betas=[0.9, 0.999],
            eps=1e-08,
            weight_decay=0,
            amsgrad=True,
            maximize=False,
            foreach=None,
            capturable=False,
            differentiable=False,
            fused=None,
            batch_size=torch.Size([]),
            device=None,
            is_shared=False),
        state: TensorDict(
            fields={
                0: TensorDict(
                    fields={
                        exp_avg: MemoryMappedTensor(shape=torch.Size([4, 3]), device=cpu, dtype=torch.float32, is_shared=True),
                        exp_avg_sq: MemoryMappedTensor(shape=torch.Size([4, 3]), device=cpu, dtype=torch.float32, is_shared=True),
                        max_exp_avg_sq: MemoryMappedTensor(shape=torch.Size([4, 3]), device=cpu, dtype=torch.float32, is_shared=True),
                        step: MemoryMappedTensor(shape=torch.Size([]), device=cpu, dtype=torch.float32, is_shared=True)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False),
                1: TensorDict(
                    fields={
                        exp_avg: MemoryMappedTensor(shape=torch.Size([4]), device=cpu, dtype=torch.float32, is_shared=True),
                        exp_avg_sq: MemoryMappedTensor(shape=torch.Size([4]), device=cpu, dtype=torch.float32, is_shared=True),
                        max_exp_avg_sq: MemoryMappedTensor(shape=torch.Size([4]), device=cpu, dtype=torch.float32, is_shared=True),
                        step: MemoryMappedTensor(shape=torch.Size([]), device=cpu, dtype=torch.float32, is_shared=True)},
                    batch_size=torch.Size([]),
                    device=None,
                    is_shared=False)},
            batch_size=torch.Size([]),
            device=None,
            is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```

The file path looks like this
```
tmpbf33lwvd/
    state/
        0/
            max_exp_avg_sq.memmap
            exp_avg_sq.memmap
            step.memmap
            exp_avg.memmap
            meta.json
        1/
            max_exp_avg_sq.memmap
            exp_avg_sq.memmap
            step.memmap
            exp_avg.memmap
            meta.json
        meta.json
    metadata/
        _tensordict/
            meta.json
        meta.json
    meta.json
```

vmoens:
cc @shagunsodhani this may be relevant

The goal is to explore if tensorclass could be a good backend for optimizer checkpointing.
Advantages:
- does not rely on pickle (which is "unsafe" when not saving only tensors)
- can load partial state-dict: say you build 2 groups in your optimizer, you could load just one of the 2 (since we save each tensor separately)
- The data structure on disk is self-explanatory and the hparams are accessible in json files (easy to parse from the outside)

PR #592, opened 2023-12-08
[Feature] Multithread memmap

vmoens:
TODO:
- [ ] Doc
- [x] Tests
- [x] hide the executor logit under a private `_memmap_`
- [x] Make a `memmap` version that does not change the tenosrdict inplace

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -1650,7 +1650,7 @@ class TensorDict(TensorDictBase):
             self._is_shared = False  # since they are mutually exclusive
             self._device = torch.device("cpu")
         dest._is_locked = True
-        return self
+        return dest
 
     @classmethod
     def load_memmap(cls, prefix: str) -> T:
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -1477,7 +1477,7 @@ class TensorDict(TensorDictBase):
         self,
         prefix: str | None = None,
         copy_existing: bool = False,
-        num_threads: int=0,
+        num_threads: int = 0,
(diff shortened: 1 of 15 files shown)


tensordict/_td.py
```diff
@@ -1510,21 +1518,26 @@ def save_metadata(data: TensorDictBase, filepath, metadata=None):
...
+                        filename=str(prefix / f"{key}.memmap")
+                        if prefix is not None
+                        else None,
+                        copy_existing=copy_existing,
+                        # copy_existing=copy_existing,
+                    )
+                if executor is None:
+                    _populate()
+                else:
+                    executor.submit(_populate)
```
laurencer, line 1540: Probably want to return a Future here so the caller can block on the saves actually completing if they want (or get a callback when it completes)

vmoens:
@laurencer thanks for the review.
I created the futures (execution time is roughly identical as before). I thought that `executor.__exit__()` would wait for submitted jobs to complete but after testing it, it doesn't seem to be the case. Do we want to optionally let users proceed with the main script without waiting for the futures to complete (and return the futures)?

I also prevented conflicting executor and num_threads, but eventually I think `executor` will only be an arg of a sub-call to a new _memmap private function (as well as `futures` to restrict the args of the public method to what they should be).

fegin:
@vmoens Thanks for the information. Just out of curiosity, have you try to compare the performance if the target storage is a memory file system?

vmoens:
That would be `memmap_()` (without file associated). I can check that!

PR #594, opened 2023-12-08
[Feature] Convert tf.Tensors to tensor

vmoens:


vmoens:
Incorporated in #601 for to_dict compatibility

PR #596, opened 2023-12-11
[Feature] state_dict hooks compatibility in `from_module` and `to_module`

vmoens:
Optionally calls the state_dict hooks in `state_dict` and `load_state_dict`.

cc @fegin


fegin:
@vmoens The PR looks good to me.

I'm thinking to write a DCP storage plugin with TensorDict to see how it works. But we will need TensorDict to support 1.) DTensor and 2.) optimizer state_dict.

vmoens:
@fegin I made the from-module method a little more recursive.

Let me see about DTensors and optimizers state-dict compatibility.

PR #603, opened 2023-12-21
[Feature] Remove and check for prints in codebase using flake8-print

vmoens:


.github/unittest/linux/scripts/run-clang-format.py
```diff
@@ -216,7 +217,7 @@ def print_trouble(prog, message, use_colors):
     error_text = "error:"
     if use_colors:
         error_text = bold_red(error_text)
-    print(f"{prog}: {error_text} {message}", file=sys.stderr)
+    logging.info(f"{prog}: {error_text} {message}", file=sys.stderr)
```
shagunsodhani, line 220: nit: if this is logged to the stderr, maybe we should log it as error i.e. logging.error

PR #610, opened 2024-01-05
[Feature] inplace to_module

vmoens:
Allows `to_module` to write tensors in-place.

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -308,6 +308,8 @@ class TensorDict(TensorDictBase):
     def to_module(
         self,
         module,
+        *,
+        inplace: bool | None = None,
         return_swap: bool = True,
         swap_dest=None,
         memo=None,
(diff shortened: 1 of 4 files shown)


tensordict/_td.py
```diff
@@ -308,6 +308,8 @@ def is_empty(self):
     def to_module(
         self,
         module,
+        *,
+        inplace: bool | None = None,
```
shagunsodhani, line 312: Is `None` supported for backward compatibility ?

vmoens: We want to catch inplace=True when use_state_dict=False since this is not implemented.

By making it None, and given that inplace is True for state dict, we tell users: you explicitly asked for False but we can't make that happen.

shagunsodhani: Thanks for explaining!

PR #618, opened 2024-01-15
[Feature] Memory-mapped nested tensors

vmoens:
This PR makes it possible to create memory-mapped tensors with heterogeneous (jagged) shapes using nested tensors as a backend.

The usage I have in mind with this is to amortize the cost of reading files: this can be done once and for all provided that there is enough space to store a single gigantic uint8 tensor on some scratch storage (more elaborate pipelines can be thought of if the dataset is split in several chunks etc)

Code example with torchvision where we decode a small dataset on a single tensor to cache the decoding phase, then do a similar preprocessing using resize with a resized buffer:
```python
import torchvision
from tensordict import MemoryMappedTensor, TensorDict
import tempfile
import torch
from torchvision.transforms.v2 import ToTensor, Resize

with tempfile.TemporaryDirectory() as path:
    # Create a fake dataset
    shapes = []
    for i in range(1000):
        w, h = torch.randint(low=200, high=300, size=(2,)).tolist()
        image = torch.randint(256, (3, w, h), dtype=torch.uint8)
        shapes.append(image.shape)
        torchvision.io.write_jpeg(image, path + "/" + str(i) + ".jpeg")

    # Create a memmap tensor
    mmap_tensor = MemoryMappedTensor.zeros(torch.tensor(shapes), dtype=torch.uint8)
    td = TensorDict({"image": mmap_tensor, "index": torch.arange(1000)}, batch_size=[1000])
    totensor = ToTensor()
    def load_preproc(td):
        i = td["index"].item()
        img = torchvision.io.read_image(path + "/" + str(i) + ".jpeg")
        td.set_("image", totensor(img))
        return None
    print("Loading images")
    td.map(load_preproc, chunksize=0, mp_start_method="fork", num_workers=8, pbar=True)

    print("Resizing images")
    resize = Resize((32, 32))
    # resize preproc
    def preproc_resize(td):
        td["image"] = resize(td["image"])
        return td

    out = preproc_resize(td[0]).expand(td.shape).memmap_(path + "/resized", num_threads=32)
    td.map(preproc_resize, chunksize=0, mp_start_method="fork", num_workers=8, pbar=True, out=out)
    print(out)
``` 

cc @albanD @cpuhrsch @NicolasHug @mikaylagawarecki 

Gist: https://gist.github.com/vmoens/d50dc6a7defe823444bcc80143bf37fd


vmoens:
Blocked by https://github.com/pytorch/pytorch/pull/117711

PR #624, opened 2024-01-17
[BugFix] Remove shared/memmap inheritance from clone / select / exclude

vmoens:
## Description

@shagunsodhani I may benefit from a bit of feedback on this.

## Context

When a tensordict is placed in shared memory or memmaped, it is blocked. We do this to avoid having people writing in it and hoping that these changes will be reflected in another process (which won't be the case). Locking the tensordict ensures that you must first unlock it (and hence loose the `is_shared()` attribute) before writing. 

For some operations (typically all the ops that don't change the `data_ptr()` ) I thought it was cool to keep the `is_shared()` attribute since we can be sure that the content is still shared. That means that coming from a shared / memmap tensordict all these ops would return a shared and locked tensordict:

```python
td.view(-1)
td.transpose(0, 1)
td.select("key")
td.exclude("key")
td.clone(recurse=False) # clones the tree structure but not the tensors
td.flatten_keys()
td.unflatten_keys()
```

## Problem

In the past we just copies the private `_is_memmap` and `_is_shared` but not the lock: that meant that you ended up with a shared but not locked TD (which is bad!) I solved this in #621 

Unfortunately that breaks a lot of stuff in torchrl:

Problem is that usually if you do `clone`, `select` or `exclude` you may want to modify the tensordict that you had.

So the plan now will be:
- For shape operations (unbind, view, transpose, permute, squeeze, unsqueeze) you keep the lock and the shared/memmap attribute because we assume that these ops are primarily there for you to present your data in a different format
- Key-based operations (clone(False), select, exclude, flatten_keys, unflatten_keys) do not propagate the shared and lock attribute

Thoughts?

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -1856,8 +1856,11 @@ class TensorDict(TensorDictBase):
             names=copy(self._td_dim_names),
             _run_checks=False,
         )
-        if not recurse:
-            self._maybe_set_shared_attributes(result)
+        # If this is uncommented, a shallow copy of a shared/memmap will be shared and locked too
+        # This may be undesirable, not sure if this should be the default behaviour
+        # (one usually does a copy to modify it).
+        # if not recurse:
+        #     self._maybe_set_shared_attributes(result)
         return result
 
     def contiguous(self) -> T:
@@ -1922,7 +1925,11 @@ class TensorDict(TensorDictBase):
         if inplace:
             self._tensordict = result._tensordict
             return self
-        self._maybe_set_shared_attributes(result)
+        # If this is uncommented, a shallow copy of a shared/memmap will be shared and locked too
+        # This may be undesirable, not sure if this should be the default behaviour
+        # (one usually does a copy to modify it).
+        # if set_shared:
+        #     self._maybe_set_shared_attributes(result)
         return result
 
     def _exclude(self, *keys: str, inplace: bool = False, set_shared: bool = True) -> T:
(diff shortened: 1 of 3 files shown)


vmoens:
I merged to fix torchrl's CI but happy to revert / edit if you feel changes are needed

shagunsodhani:
> For some operations (typically all the ops that don't change the data_ptr() ) I thought it was cool to keep the is_shared() attribute since we can be sure that the content is still shared. That means that coming from a shared / memmap tensordict all these ops would return a shared and locked tensordict:

Makes sense!

> So the plan now will be:
> 
> * For shape operations (unbind, view, transpose, permute, squeeze, unsqueeze) you keep the lock and the shared/memmap attribute because we assume that these ops are primarily there for you to present your data in a different format
> * Key-based operations (clone(False), select, exclude, flatten_keys, unflatten_keys) do not propagate the shared and lock attribute
> 
> Thoughts?

This seems to be roughly in-line with the idea that if the input and output of the ops share storage (e.g. with unbind or transpose), we propogate the lock and shared/memmap attributes while in other cases, we do not. This is very reasonable and should not "surprise" the user. Some things to look out for:

1. `select` is a `view-like` operator in both [PyTorch](https://pytorch.org/docs/stable/generated/torch.select.html#torch.select) and [TD] (https://pytorch.org/tensordict/reference/generated/tensordict.TensorDict.html#tensordict.TensorDict.select) but in the current proposal, the behavior is different. I think the decision of `these ops are primarily there for you to present your data in a different format` is a subjective choice while relying on the idea, that we propogate lock and shared/memmap attributes as long as storage remains the same, is more objective and consistent.

2. Some ops like `reshape` do not have a consistent behavior (in terms of copying underlying data) in [PT](https://pytorch.org/docs/stable/generated/torch.reshape.html#torch.reshape) and I assume its behavior is the same in TD. So the behavior of propogating locks etc would be a bit in-consistent (when looked at the level of op) but will still be consistent with the idea of tieing locks etc with storage.

shagunsodhani:
Makes sense!

PR #632, opened 2024-01-23
[Doc] More doc on how to set and get non-tensor data

vmoens:


tensordict/tensorclass.py
```diff
@@ -1147,6 +1147,20 @@ class NonTensorData:
...
+        are overloaded to set non-tensor data appropriately (unlike :meth:`~tensordict.TensorDictBase.set`
+        and :meth:`~tensordict.TensorDictBase.get` which are reserved for tensor-like
+        objects):
+
+        >>> td = TensorDict({"a": torch.zeros(3)}, batch_size=[3])
+        >>> td["a"]  # gets a tensor
+        >>> td["b"] = "a string!"
+        >>> assert td["b"] == "a string!"
+        >>> # indexing preserves the meta-data
+        >>> assert td[0]["b"] == "a string!"
```
shagunsodhani, line 1161: curious would this still work if the `batch_size` was 2 dimensional (say (3, 2)). Similarly, would

```
assert td[0][1]["b"] == "a string!"
```

work with a 2-dimensional batch size ?

PR #635, opened 2024-01-24
[Doc] Update readme for v0.3

vmoens:


README.md
```diff
@@ -77,6 +78,25 @@ live on a dedicated device, which will send each tensor that is written there:
 >>> assert tensordict["key 3"].device is torch.device("cuda:0")
 ```
 
+But that is not all, you can also store nested values in a tensordict:
+```python
+>>> tensordict["nested", "key"] = torch.zeros(3, 4) # the batch-size must match
+```
+and any nested tuple structure will be unravelled to make it easy to code reading and
```
shagunsodhani, line 85: nit: 

```suggestion
and any nested tuple structure will be unravelled to make it easy to read code and
```

README.md
```diff
@@ -77,6 +78,25 @@ live on a dedicated device, which will send each tensor that is written there:
...
+writing ops programmatically:
```
shagunsodhani, line 86: nit:
```suggestion
write ops programmatically:
```

README.md
```diff
@@ -126,6 +146,13 @@ If a functionality is missing, it is easy to call it using `apply()` or `apply_(
 ```python
 tensordict_uniform = tensordict.apply(lambda tensor: tensor.uniform_())
 ```
+
+``apply()`` can also be great to filter a tensordict, for instance:
+```python
+data = TensorDict({"a": 1.0, "b": 1}, [])
+data_float = data.apply(lambda x: x if x.dtype == torch.float else None) # contains only the "a" key
```
shagunsodhani, line 153: nit: this example may need more explanation as the user could expect the values of `data` to be float / int and not tensors.

README.md
```diff
@@ -146,39 +173,61 @@ When nodes share a common scratch space, the
...
-Moreover, tensordict modules are compatible with `torch.fx` and `torch.compile`,
+Moreover, tensordict modules are compatible with `torch.fx` and (soon) `torch.compile`,
 which means that you can get the best of both worlds: a codebase that is
 both readable and future-proof as well as efficient and portable!
 
+### TensorDict for parameter serialization
+
+TensorDict offers an API for parameter serialization that can be >3x faster than
+regular calls to `torch.save(state_dict)`. Moreover, because tensors will be saved
+independently on disk, you can deserialized your checkpoint on an arbitrary slice
```
shagunsodhani, line 216: nit:

```suggestion
independently on disk, you can deserialize your checkpoint on an arbitrary slice
```

README.md
vmoens: yep i'm packing many things there, let's make it more straightforward

PR #652, opened 2024-02-02
[BugFix, Feature] `pad_sequence` refactoring

### LINKED ISSUE #643: [BUG] pad_sequence raises an error when return_mask=True
## Describe the bug

While the pad_sequence function works, when the flag return_mask is set to True, it raises an error. 

## To Reproduce

```python
from tensordict import TensorDict, pad_sequence

list_td = [TensorDict({"a": torch.zeros((3,))}, []), TensorDict({"a": torch.zeros((4,))}, [])]
padded_td = pad_sequence(list_td, return_mask=True, batch_first=True)
```

```bash
  File "/Users/dtsaras/Documents/CS/circuit_gen/test.py", line 37, in <module>
    padded_td = pad_sequence(list_td, batch_first=True, return_mask=True)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/c_gen/lib/python3.10/site-packages/tensordict/tensordict.py", line 5717, in pad_sequence
    raise RuntimeError(f"pad_sequence failed for key {key}") from err
RuntimeError: pad_sequence failed for key mask
```

## Expected behavior

A TensorDict with a "mask" key that contains a bool tensor where the entries with padded_val should be the padded entries in the corresponding padded sequence. 

## System info

Describe the characteristic of your environment:
 * Install with pip install tensordict
 *  Python version 3.10

```python
```
0.2.1 1.26.2 3.10.13 (main, Sep 11 2023, 08:16:02) [Clang 14.0.6 ] darwin 2.1.2

## Additional context

Another potential improvement in the current implementation is that we might need multiple mask for every different key name {key}_mask

## Reason and Possible fixes

The issue starts in the file tensordict/tensordict.py, line 5685 (if return_mask:). The issue is that the mask is created by using the shape td.shape, where in the case of a single TensorDict, it has a size of torch.Size([]), so it create a mask that is torch.tensor(True). The fixing would be to either completely remove this option, or create a mask for every key in the TensorDict and then create a mask of the appropriate size. 

## Checklist

- [Yes] I have checked that there is no similar issue in the repo (**required**)
- [Yes] I have read the [documentation](https://github.com/pytorch/tensordict/tree/main/docs/) (**required**)
- [Yes] I have provided a minimal working example to reproduce the bug (**required**)

dtsaras:
## Description

The below changes have been discussed with @vmoens :

The current implementation fixes the pad_sequence implementation when `return_mask=True` and adds an additional key named "masks". The key "masks" contains a TensorDict with the same keys as the stacked tensordict (preserving nested keys) and every key contains the mask of valid values in the stacked tensordict. This PR removes the `batch_first` and always returns a tensordict with the batching dimension being first. Additionally, adds a parameter `pad_dim: int = 0` where the user can select in which dimension to pad the tensors. Thus, the masks will have a shape of `torch.Size([batch_size, max_seq_length])` of the given dimension. Finally, added tests for the new `pad_sequence` operator. 

## Motivation and Context

It fixes the issue:
https://github.com/pytorch/tensordict/issues/643
`close #643 `

- [Yes]  I have raised an issue to propose this change ([required](https://github.com/pytorch/tensordict/issues) for new features and bug fixes)

## Types of changes

- Bug fix (non-breaking change which fixes an issue)
- Documentation (update in the documentation)

## Checklist

- [Maybe?] I have updated the documentation accordingly. 
I have updated the example and the description in the description of `pad_sequence` but I am not sure if I need to explicitly update the wiki.


tensordict/functional.py
```diff
@@ -77,7 +77,7 @@ def pad(tensordict: T, pad_size: Sequence[int], value: float = 0.0) -> T:
 
 def pad_sequence(
     list_of_tensordicts: Sequence[T],
-    batch_first: bool = True,
+    pad_dim: int = 0,
```
vmoens, line 80: This is bc breaking, we should keep batch_first as the last learn and set its default to None. If it's not None we raise a deprecation warning.

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
 
     Args:
         list_of_tensordicts (List[TensorDictBase]): the list of instances to pad and stack.
-        batch_first (bool, optional): the ``batch_first`` correspondant of :func:`torch.nn.utils.rnn.pad_sequence`.
-            Defaults to ``True``.
+        pad_dim (int, optional): the ``pad_dim`` indicates the dimension to pad all the keys in the tensordict.
```
vmoens, line 90: Shouldn't we default to -1? Or tensordict.ndim (to pad the first dim after batch size)?
Also I would make it clear that we're talking about tensor dimensions, not tensordict's

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
+            Defaults to ``0``.
         padding_value (number, optional): the padding value. Defaults to ``0.0``.
         out (TensorDictBase, optional): if provided, the destination where the data will be
             written.
         device (device compatible type, optional): if provded, the device where the
             TensorDict output will be created.
-        return_mask (bool, optional): if ``True``, a "mask" entry will be returned.
-            It contains the mask of valid values in the stacked tensordict.
+        return_mask (bool, optional): if ``True``, a "masks" entry will be returned.
+            It contains a tensordict with the same structure as the stacked tensordict where every key contains the mask of valid values with size ``torch.Size([batch_size, max_seq_length])``.
```
vmoens, line 98:
```suggestion
            It contains a tensordict with the same structure as the stacked tensordict where every entry contains the mask of valid values with size ``torch.Size([stack_len, *new_shape])``, 
            where `new_shape[pad_dim] = max_seq_length` and the rest of the `new_shape` matches the previous shape of the contained tensors.
```

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
 
     Examples:
         >>> list_td = [
-        ...     TensorDict({"a": torch.zeros((3,))}, []),
-        ...     TensorDict({"a": torch.zeros((4,))}, []),
+        ...     TensorDict({"a": torch.zeros((3,8)), "b": torch.zeros((6,8))}, []),
+        ...     TensorDict({"a": torch.zeros((5,8)), "b": torch.zeros((6,8))}, []),
```
vmoens, line 103:
```suggestion
        ...     TensorDict({"a": torch.zeros((3, 8)), "b": torch.zeros((6, 8))}, batch_size=[]),
        ...     TensorDict({"a": torch.zeros((5, 8)), "b": torch.zeros((6, 8))}, batch_size=[]),
```

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
         TensorDict(
             fields={
-                a: Tensor(shape=torch.Size([2, 4]), device=cpu, dtype=torch.float32, is_shared=False)},
-            batch_size=torch.Size([]),
+                a: Tensor(shape=torch.Size([2, 4, 8]), device=cpu, dtype=torch.float32, is_shared=False),
+                b: Tensor(shape=torch.Size([2, 5, 8]), device=cpu, dtype=torch.float32, is_shared=False),
+                masks: TensorDict(
+                    fields={
+                        a: Tensor(shape=torch.Size([2, 4]), device=cpu, dtype=torch.bool, is_shared=False),
+                        b: Tensor(shape=torch.Size([2, 6]), device=cpu, dtype=torch.bool, is_shared=False)},
```
vmoens, line 114: What shape does the mask have if `pad_dim=-1`? I guess that here the mask shape depends on the pad_dim

tensordict/functional.py
vmoens: I realize this is only true if the mask has the same shape as the tensor but given the example below it does not seem to be the case. I guess it's ok as long as we clearly state what shape the masks will have (i.e., explain that they have the shape of the tensor up until `pad_dim`)

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
-        list_of_tensordicts = [
-            td.clone(False).set("mask", torch.ones(td.shape, dtype=torch.bool))
-            for td in list_of_tensordicts
-        ]
+        for key in _check_keys(
+            list_of_tensordicts, leaves_only=True, include_nested=True
+        ):
+            list_of_tensordicts = [
+                td.clone(False).set(
+                    tuple(["masks"] + list(key)),
```
vmoens, line 132:
```suggestion
                    ("masks", key),
```
works just fine

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
+                    torch.ones(td.get(key).shape[pad_dim], dtype=torch.bool),
```
vmoens, line 133: that won't work if the tensordict has a shape
```suggestion
                    torch.ones(td.get(key).shape[:pad_dim+1], dtype=torch.bool),
```
is an idea but you'll need to convert the `pad_dim` to non-negative value first (bc -1+1 is 0 and [:0] isn't the same as [:])

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
     if return_mask:
-        list_of_tensordicts = [
-            td.clone(False).set("mask", torch.ones(td.shape, dtype=torch.bool))
-            for td in list_of_tensordicts
-        ]
+        for key in _check_keys(
+            list_of_tensordicts, leaves_only=True, include_nested=True
+        ):
+            list_of_tensordicts = [
+                td.clone(False).set(
```
vmoens, line 131: the issue with this implementation is that you'll lose the metadata of each sub-tensordict like dim names and batch-size or devices.

Here's another implementation (untested
```python
def func(tensor):
    dim = pad_dim if pad_dim >= 0 else tensor.ndim + pad_dim
    # maybe a test here to check that dim is correct? ie not bigger than tensor.ndim-1 and not negative
    return torch.ones(tensor.shape[:dim], device=tensor.device, dtype=torch.bool)
td.apply(func)
```

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
-        shape = [
-            len(list_of_tensordicts),
-        ]
-    elif batch_first:
-        shape = [len(list_of_tensordicts), shape]
-    else:
-        shape = [shape, len(list_of_tensordicts)]
+    shape = [
+        len(list_of_tensordicts),
+    ]
```
vmoens, line 141: if we want to include the batch size (I think we should) shape should be `shape = [len(list_of_tensordicts)] + batch_size`

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
+
     if out is None:
         out = TensorDict(
             {}, batch_size=torch.Size(shape), device=device, _run_checks=False
         )
-        for key in keys:
-            try:
-                out.set(
+        set_func = out.set
```
vmoens, line 147: you can just do `set(..., inplace=True)` which is a best-intention version of set (uses inplace if the key is already there, otherwise writes a new entry)

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
-    else:
-        shape = [shape, len(list_of_tensordicts)]
+    shape = [
+        len(list_of_tensordicts),
+    ]
+
     if out is None:
         out = TensorDict(
             {}, batch_size=torch.Size(shape), device=device, _run_checks=False
         )
```
vmoens, line 146: I think 
```python
out = lsit_of_tensordicts[0].empty(recurse=True)
out = out.expand(*shape)
```
is more appropriate

tensordict/functional.py
```diff
@@ -87,76 +87,92 @@ def pad_sequence(
...
+                            for td in list_of_tensordicts
+                        ],
+                        batch_first=True,
+                        padding_value=padding_value,
+                    ).transpose(1, pad_dim + 1),
+                )
+            else:
+                set_func(
                     key,
                     torch.nn.utils.rnn.pad_sequence(
```
vmoens, line 167: I guess we can't use tensordicts with batch_size because we're relying on pad_sequence.
I thought about using `functorch.dim` here to "hide" the first dimensions and use pad_sequence but I get a `RuntimeError: Batching rule not implemented for aten::pad_sequence. We could not generate a fallback.` (cc @zdevito if you're interested: we'd like to use pad_sequence for batches)

The workaround is this: for all tensordicts, create a version of the tensordict where the batch dimensions are permuted to the back of the dims, then permuted back. We can use tensordict.apply all along:


```python
td_a = TensorDict({
    "a": torch.randn(10, 11, 3, 5),
    "b": torch.randn(10, 11, 3, 5),
}, [10, 11])
td_b = TensorDict({
    "a": torch.randn(10, 11, 4, 5),
    "b": torch.randn(10, 11, 4, 5),
}, [10, 11])

def permute_batch_size(tensor, batch_size=[10, 11, ]):
    L = len(batch_size)
    return tensor.permute(*range(L, tensor.ndim), *range(L))
td_a_p = td_a.apply(permute_batch_size, batch_size=[])
td_b_p = td_b.apply(permute_batch_size, batch_size=[])

def pad_seq(*tensors):
    return torch.nn.utils.rnn.pad_sequence(list(tensors), batch_first=True)
td_pad = td_a_p.apply(pad_seq, td_b_p)
def reset_batch_size(tensor, batch_size=[10, 11, ]):
    L = len(batch_size)
    return tensor.permute(0, *range(tensor.ndim - L, tensor.ndim), *range(1, tensor.ndim - L))
td_pad = td_pad.apply(reset_batch_size)

print(td_pad)
```
prints
```
TensorDict(
    fields={
        a: Tensor(shape=torch.Size([2, 10, 11, 4, 5]), device=cpu, dtype=torch.float32, is_shared=False),
        b: Tensor(shape=torch.Size([2, 10, 11, 4, 5]), device=cpu, dtype=torch.float32, is_shared=False)},
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```

which is pretty cool. You don't really need to reset the batch size since we called empty above so our out-container has already all the metadata set.

test/test_tensordict.py
```diff
@@ -888,33 +888,63 @@ def test_pad(self):
...
-            TensorDict({"a": torch.ones((4,)), ("b", "c"): torch.ones((4, 3))}, [4]),
-        ]
-        padded_td = pad_sequence(
-            list_td, batch_first=batch_first, return_mask=make_mask
-        )
-        if batch_first:
-            assert padded_td.shape == torch.Size([2, 4])
-            assert padded_td["a"].shape == torch.Size([2, 4])
-            assert padded_td["a"][0, -1] == 0
+    def test_pad_sequence(self, pad_dim, make_mask):
```
vmoens, line 893: note to self: review test when implementation is fixed

vmoens (commented):
Awesome stuff! thanks a mil
I think we need to account for non empty batch size, see my last comment and go backward from there :)

tensordict/functional.py
dtsaras: I think the most common application for padding tensors is in sequential data. For example, the transformer expects the input data to be in the form:
`src: (S,E) for unbatched input, (N, S, E) or (S, N, E)`
where S is the sequence dimension, E is the embedding dimension, and N is the batch dim. Since in our case we stack the tensordicts into a new dimension, I think our tensordicts are supposed to be a collection of the form (S, E). Thus, I think it would be a good option to default to `pad_dim=0`.

vmoens: My issue with that is that tensordict is designed to work with data that comes in batches of arbitrary dimensions. Defaulting to 0 will cover only a small subset of cases (to me we should raise an exception any time the dimension is within the number of batch dimensions).
We could say None by default which means defaulting to input.ndim wdyt?

dtsaras: Since we are stacking the TensorDicts all the tensors for each key must have the same dimensionality except the `pad_dim`. By stacking we create a new dimension which I interpret as the batch_size. We can consider to generate a more advanced operator which is somewhat equivalent to `torch.cat` but also pads every other dimension so we can handle more arbitrary sizes.

vmoens:
Hey @dtsaras do you want me to take this over? I can do the patch for the batch size if you'd like

test/test_tensordict.py
vmoens, line 893: I think we have two completely separated tests here (the if/else completely splits the function)

test/test_tensordict.py
```diff
@@ -888,33 +888,63 @@ def test_pad(self):
...
+            list_td = [
+                TensorDict(
+                    {"a": torch.ones((2, 8, 8)), ("b", "c"): torch.ones((2, 3))}, [2]
+                ),
+                TensorDict(
+                    {"a": torch.ones((4, 8, 8)), ("b", "c"): torch.ones((4, 3))}, [4]
+                ),
+            ]
+            padded_td = pad_sequence(list_td, pad_dim=pad_dim, return_mask=make_mask)
+            assert padded_td.shape == torch.Size([2])
```
vmoens, line 904: shouldn't it be [2, 4]?
We've lost a batch dim in the process.
I think if `dim` overlaps the batch_dims we should keep it no?

test/test_tensordict.py
```diff
@@ -888,33 +888,63 @@ def test_pad(self):
...
-            assert padded_td["a"].shape == torch.Size([2, 4])
-            assert padded_td["a"][0, -1] == 0
+    def test_pad_sequence(self, pad_dim, make_mask):
+        if pad_dim == 0:
+            list_td = [
+                TensorDict(
+                    {"a": torch.ones((2, 8, 8)), ("b", "c"): torch.ones((2, 3))}, [2]
+                ),
+                TensorDict(
+                    {"a": torch.ones((4, 8, 8)), ("b", "c"): torch.ones((4, 3))}, [4]
```
vmoens, line 900: perhaps make this guy full of 2s such that we can check that the chunks look like data from the first or second tensordict

test/test_tensordict.py
```diff
@@ -888,33 +888,63 @@ def test_pad(self):
...
-        else:
-            assert "mask" not in padded_td.keys()
+            list_td = [
+                TensorDict(
+                    {"a": torch.ones((6, 3, 8)), ("b", "c"): torch.ones((6, 3))}, []
+                ),
+                TensorDict(
+                    {"a": torch.ones((6, 5, 8)), ("b", "c"): torch.ones((6, 7))}, []
+                ),
+            ]
```
vmoens, line 929: I would add a series of tests here when the batch-size is `[]` or `[6]`
(ideally if it's [6] the end batch size should be `[2, 6]`)

test/test_tensordict.py
```diff
@@ -888,33 +888,63 @@ def test_pad(self):
...
+            if make_mask:
+                padded_td_without_masks = pad_sequence(
+                    list_td, pad_dim=pad_dim, return_mask=False
+                )
+                assert "masks" in padded_td.keys()
+                assert set(
+                    padded_td_without_masks.keys(include_nested=True, leaves_only=True)
+                ) == set(padded_td["masks"].keys(include_nested=True, leaves_only=True))
+                assert not padded_td["masks", "a"].all()
+                assert not padded_td["masks", "b", "c"].all()
```
vmoens, line 918: we should explicitly test the shapes of the masks
To me the mask shape is the shape of the tensor up until the `pad_dim` (i.e. it has the same shape as the tensor if `pad_dim=-1`)

vmoens (commented):
Looks always better! I added a few suggestions for the tests. We should also consider `pad_dim=-1`

test/test_tensordict.py
dtsaras: I have been thinking that the shape of a masks should be either identical shape of the tensor i.e. ```td[key].shape == td["masks"][key].shape``` which is a complete mask or it can be a key mask i.e. ```td["mask][key].shape = td.batch_size + torch.Size([pad_dim])```. In the case that the ```pad_dim``` is part of the ```td.batch_size```, then the mask has shape ```td.batch_size```. The current implementation follows the second case because I think it's easier to transform the key mask to any potentially desired mask.

test/test_tensordict.py
dtsaras: I already did a lot of tests to ensure the size is correct and a lot more of them, with much larger sequences of batch_sizes as well. If you think any of the below tests would he worth adding lmk. 
```python
print("############ Example 1 ############")
    list_td = [
        TensorDict({"a": torch.zeros((3,8)), "b": torch.zeros((5,8))}, []),
        TensorDict({"a": torch.zeros((4,8)), "b": torch.zeros((5,8))}, []),
    ]
    padded_td = pad_sequence(
        list_td, return_mask=True, pad_dim=0
    )
    print(padded_td)

    print("############ Example 2.1 ############")
    list_td = [
        TensorDict({"a": torch.zeros((4,3)), "b": torch.zeros((4,1))}, [4]),
        TensorDict({"a": torch.zeros((4,5))}, [4]),
    ]
    padded_td = pad_sequence(
        list_td, return_mask=True, pad_dim=1, out=TensorDict({"a": torch.ones((2, 4, 5)), ("masks", "a"): torch.ones((2,4,5), dtype=torch.bool)}, batch_size=[2,4])
    )
    print(padded_td)
    # print(padded_td["a"], padded_td["masks"]["a"])

    print("############ Example 2.2 ############")
    list_td = [
        TensorDict({"a": torch.zeros((4,3)), "b": torch.zeros((4,1))}, [4]),
        TensorDict({"a": torch.zeros((4,5))}, [4]),
    ]
    padded_td = pad_sequence(
        list_td, return_mask=True, pad_dim=-1, out=TensorDict({"a": torch.ones((2, 4, 5)), ("masks", "a"): torch.ones((2,4,5), dtype=torch.bool)}, batch_size=[2,4])
    )
    print(padded_td)
    # print(padded_td["a"], padded_td["masks"]["a"])

    print("############ Example 3 ############")
    list_td = [
        TensorDict({"a": torch.zeros((4,))}, []),
        TensorDict({"a": torch.zeros((3,)), "b": torch.zeros((5,))}, []),
        TensorDict({"a": torch.zeros((3,)), "b": torch.zeros((5,))}, []),
    ]
    padded_td = pad_sequence(
        list_td, return_mask=True
    )
    print(padded_td)
    ```

vmoens (approved):
LGTM thanks so much for this fix!
Next step: add a keys_to_pad (or similar) argument :)

PR #653, opened 2024-02-02
[Feature] Adds utils method remove_duplicates

albertbou92:
## Description

 The method detects duplicate elements in the tensor associated with the specified 'key' along the specified
'dim' and removes elements in the same indices in all other tensors within the TensorDict.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1921,7 +1921,7 @@ def remove_duplicates(
 
     # Check dim is valid
     batch_dims = input.ndim
-    if dim >= batch_dims or dim < -batch_dims:
+    if dim >= batch_dims or dim < -batch_dims or batch_dims == 0:
         raise ValueError(
             f"The specified dimension '{dim}' is invalid for a TensorDict with batch size '{input.batch_size}'."
         )
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1912,7 +1912,8 @@ def remove_duplicates(
         raise KeyError(f"The key '{key}' does not point to a tensor in the TensorDict.")
 
     # Check dim is valid
-    if dim >= len(tensordict.batch_size):
+    batch_dims = len(tensordict.batch_size)
+    if dim >= batch_dims or dim <= -batch_dims:
         raise ValueError(
             f"The specified dimension '{dim}' is invalid for a TensorDict with batch size '{tensordict.batch_size}'."
         )
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1867,18 +1867,23 @@ def _index_preserve_data_ptr(index):
 def remove_duplicates(
     tensordict: TensorDictBase, key: str, dim: int = 0
 ) -> TensorDictBase:
-    """Removes indices duplicated in 'key' along the specified dimension.
+    """Removes indices duplicated in `key` along the specified dimension.
 
-    This method detects duplicate elements in the tensor associated with the specified 'key' along the specified
-    'dim' and removes elements in the same indices in all other tensors within the TensorDict.
+    This method detects duplicate elements in the tensor associated with the specified `key` along the specified
+    `dim` and removes elements in the same indices in all other tensors within the TensorDict. It is expected for
+    `dim` to be one of the dimensions within the batch size of the input TensorDict to ensure consistency in all
+    tensors. Otherwise, an error will be raised.
 
     Args:
         tensordict (TensorDictBase): The TensorDict containing potentially duplicate elements.
-        key (NestedKey): The key of the tensor along which duplicate elements should be identified and removed.
-        dim (int): The dimension along which duplicate elements should be identified and removed. Default: 0.
+        key (NestedKey): The key of the tensor along which duplicate elements should be identified and removed. It
+            must be one of the leaf keys within the TensorDict, pointing to a tensor and not to another TensorDict.
+        dim (int): The dimension along which duplicate elements should be identified and removed. It must be one of
+            the dimensions within the batch size of the input TensorDict. Defaults to ``0``.
 
     Returns:
-        TensorDictBase: tensordict with removed elements.
+        TensorDictBase: input tensordict with the indices corrsponding to duplicated elements in tensor `key` along
+        dimension `dim` removed.
 
     Example:
         >>> tensordict = TensorDict(
@@ -1886,6 +1891,7 @@ def remove_duplicates(
         ...         "tensor1": torch.tensor([[1, 2, 3], [4, 5, 6], [1, 2, 3], [7, 8, 9]]),
         ...         "tensor2": torch.tensor([[10, 20], [30, 40], [40, 50], [50, 60]]),
         ...     }
+        ...     batch_size=[4],
         ... )
         >>> output_tensordict = remove_duplicate_elements(tensordict, key="tensor1", dim=0)
         >>> expected_output = TensorDict(
@@ -1900,6 +1906,12 @@ def remove_duplicates(
     # Check if the key is a TensorDict
     if key not in tensordict.keys(include_nested=True):
         raise KeyError(f"The key '{key}' does not exist in the TensorDict.")
+
+    # Check that the key points to a tensor
+    if not isinstance(tensordict.get(key), torch.Tensor):
+        raise ValueError(
+            f"The key '{key}' does not point to a tensor in the TensorDict."
+        )
 
     # Check dim is valid
     if dim >= len(tensordict.batch_size):
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1909,9 +1909,7 @@ def remove_duplicates(
 
     # Check that the key points to a tensor
     if not isinstance(tensordict.get(key), torch.Tensor):
-        raise ValueError(
-            f"The key '{key}' does not point to a tensor in the TensorDict."
-        )
+        raise KeyError(f"The key '{key}' does not point to a tensor in the TensorDict.")
 
     # Check dim is valid
     if dim >= len(tensordict.batch_size):
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1875,7 +1875,7 @@ def remove_duplicates(
     Args:
         tensordict (TensorDictBase): The TensorDict containing potentially duplicate elements.
         key (str): The key of the tensor along which duplicate elements should be identified and removed.
-        dim (int): The dimension along which duplicate elements should be identified and removed.
+        dim (int): The dimension along which duplicate elements should be identified and removed. Default: 0.
 
     Returns:
         TensorDictBase: tensordict with removed elements.
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1932,5 +1932,5 @@ def remove_duplicates(
 
     # Remove duplicate elements in the TensorDict
     unique_tensordict = tensordict[(slice(None),) * dim + (first_indices,)]
-    
+
     return unique_tensordict
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1894,14 +1894,14 @@ def remove_duplicates(
             input tensordict for the specified `key` along the specified `dim`. Only provided if return_index is True.
 
     Example:
-        >>> tensordict = TensorDict(
+        >>> td = TensorDict(
         ...     {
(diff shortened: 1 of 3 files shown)


tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
     if isinstance(index, slice) and (index.start == 0 or index.start is None):
         return True
     return False
+
+
+def remove_duplicates(
```
vmoens, line 1867: to be added in the doc

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+    tensordict: TensorDictBase, key: str, dim: int = 0
+) -> TensorDictBase:
+    """Removes indices duplicated in 'key' along the specified dimension.
+
+    This method detects duplicate elements in the tensor associated with the specified 'key' along the specified
+    'dim' and removes elements in the same indices in all other tensors within the TensorDict.
```
vmoens, line 1873:
```suggestion
    This method detects duplicate elements in the tensor associated with the specified `key` along the specified
    `dim` and removes elements in the same indices in all other tensors within the TensorDict.
```

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+
+    Args:
+        tensordict (TensorDictBase): The TensorDict containing potentially duplicate elements.
+        key (str): The key of the tensor along which duplicate elements should be identified and removed.
+        dim (int): The dimension along which duplicate elements should be identified and removed. Default: 0.
```
vmoens, line 1878:
```suggestion
        dim (int): The dimension along which duplicate elements should be identified and removed.
            Defaults to ``0``.
```
Also I guess dim must be within the batch-size right?

tensordict/utils.py
vmoens: We should make it clear that this all happens along the batch-size (ie, empty batch size will throw an error etc)

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+
+    Returns:
+        TensorDictBase: tensordict with removed elements.
```
vmoens, line 1881: tell more about the batch size of the resulting td

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+
+    Example:
+        >>> tensordict = TensorDict(
+        ...     {
+        ...         "tensor1": torch.tensor([[1, 2, 3], [4, 5, 6], [1, 2, 3], [7, 8, 9]]),
+        ...         "tensor2": torch.tensor([[10, 20], [30, 40], [40, 50], [50, 60]]),
+        ...     }
+        ... )
```
vmoens, line 1889: this one has no batch-size

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+        >>> output_tensordict = remove_duplicate_elements(tensordict, key="tensor1", dim=0)
+        >>> expected_output = TensorDict(
+        ...     {
+        ...         "tensor1": torch.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
+        ...         "tensor2": torch.tensor([[10, 20], [30, 40], [50, 60]]),
+        ...     },
+        ...     batch_size=[3],
+        ... )
+        >>> for key in output_tensordict.keys():
+        ...     assert torch.equal(output_tensordict[key], expected_output[key])
```
vmoens, line 1899:
```suggestion
        >>> assert (tensordict == expected_output).all()
```

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+    """
+    # Check if the key is a TensorDict
+    if key not in tensordict.keys(include_nested=True):
+        raise KeyError(f"The key '{key}' does not exist in the TensorDict.")
+
+    # Check dim is valid
+    if dim >= len(tensordict.batch_size):
```
vmoens, line 1906: we should make it work for neg dims, like -1 (last dim of the batch size). Check other methods like unbind etc to see how to deal with that

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+    tensordict: TensorDictBase, key: str, dim: int = 0
+) -> TensorDictBase:
+    """Removes indices duplicated in 'key' along the specified dimension.
+
+    This method detects duplicate elements in the tensor associated with the specified 'key' along the specified
+    'dim' and removes elements in the same indices in all other tensors within the TensorDict.
+
+    Args:
+        tensordict (TensorDictBase): The TensorDict containing potentially duplicate elements.
+        key (str): The key of the tensor along which duplicate elements should be identified and removed.
```
vmoens, line 1877: We should make it extra clear that key must point to a tensor, not a tensordict (ie it must be a leaf)

vmoens, line 1877:
```suggestion
        key (NestedKey): The key of the tensor along which duplicate elements should be identified and removed.
```

tensordict/utils.py
```diff
@@ -1862,3 +1862,65 @@ def _index_preserve_data_ptr(index):
...
+    )
+
+    # Find first occurrence of each index  (e.g. [0, 1, 3])
+    _, unique_indices_sorted = torch.sort(unique_indices, stable=True)
+    cum_sum = counts.cumsum(0)
+    cum_sum = torch.cat((torch.tensor([0]), cum_sum[:-1]))
+    first_indices = unique_indices_sorted[cum_sum]
+
+    # Remove duplicate elements in the TensorDict
+    unique_tensordict = tensordict[first_indices]
```
vmoens, line 1924: can we get the indices in the output tensordict (optionally)? like `return_indices=False` by default. The key could be `"unique_indices"`, maybe parametrizable (up to you, not sure if that makes sense). For sure we don't want to override an existing key so whatever key we choose we must check it isn't there already.

test/test_utils.py
```diff
@@ -151,6 +152,43 @@ def test_unravel_key_to_tuple():
     assert not _unravel_key_to_tuple(("a", (slice(None),), ("b",)))
 
 
+@pytest.mark.parametrize("key", ("tensor1", "tensor3"))
+@pytest.mark.parametrize("dim", (0, 1))
```
vmoens, line 156:
```suggestion
@pytest.mark.parametrize("dim", (0, 1, -1, -2))
```

test/test_utils.py
```diff
@@ -151,6 +152,43 @@ def test_unravel_key_to_tuple():
...
+def test_remove_duplicates(key, dim):
```
vmoens, line 157: I would create a td with nested keys too to make sure everything works ok

vmoens (commented):
Great work! I left some suggestions to start the discussion

tensordict/utils.py
albertbou92: We can but since the indices are an 1D tensor of length batch_size[dim], it is not guaranteed that we can add them to the tensordict. We can return them separately (optionally), which also solves the problem of potentially override an existing key.

albertbou92:
I added the option to return the indices. 
I also handled the case of negative dims, which are now accepted.
finally I added some tests for the case of TDs of more than 1 dimension.

tensordict/utils.py
```diff
@@ -1862,3 +1862,89 @@ def _index_preserve_data_ptr(index):
...
+    This method detects duplicate elements in the tensor associated with the specified `key` along the specified
+    `dim` and removes elements in the same indices in all other tensors within the TensorDict. It is expected for
+    `dim` to be one of the dimensions within the batch size of the input TensorDict to ensure consistency in all
+    tensors. Otherwise, an error will be raised.
+
+    Args:
+        tensordict (TensorDictBase): The TensorDict containing potentially duplicate elements.
+        key (NestedKey): The key of the tensor along which duplicate elements should be identified and removed. It
+            must be one of the leaf keys within the TensorDict, pointing to a tensor and not to another TensorDict.
+        dim (int): The dimension along which duplicate elements should be identified and removed. It must be one of
```
vmoens, line 1884: Int, optional

tensordict/utils.py
```diff
@@ -1862,3 +1862,89 @@ def _index_preserve_data_ptr(index):
...
+    # Check if the key is a TensorDict
+    if key not in tensordict.keys(include_nested=True):
+        raise KeyError(f"The key '{key}' does not exist in the TensorDict.")
+
+    # Check that the key points to a tensor
+    if not isinstance(tensordict.get(key), torch.Tensor):
+        raise KeyError(f"The key '{key}' does not point to a tensor in the TensorDict.")
+
+    # Check dim is valid
+    batch_dims = len(tensordict.batch_size)
```
vmoens, line 1922: Or tendordict.ndim

albertbou92:
I added tests on different devices, turned return_indices into a kwarg only and fixed the other minor comments.

Regarding adding `unique_indices` to the output tensordict, `unique_indices` will also be a tensor of size `input.batch_size[dim]`. There is no guarantee that we can add this to the output TensorDict due to the shape (only if dim=0 and len(batch_size) = 1). This is why I decided to return it as an extra output.

vmoens:
> I added tests on different devices, turned return_indices into a kwarg only and fixed the other minor comments.
> 
> Regarding adding `unique_indices` to the output tensordict, `unique_indices` will also be a tensor of size `input.batch_size[dim]`. There is no guarantee that we can add this to the output TensorDict due to the shape (only if dim=0 and len(batch_size) = 1). This is why I decided to return it as an extra output. is there any alternative?

Good point. The alternative would be to expand it on the left but it's messy. Happy with the PR in its current state then!

tensordict/utils.py
```diff
@@ -1862,3 +1862,90 @@ def _index_preserve_data_ptr(index):
...
+    # Get indices of unique elements (e.g. [0, 1, 0, 2])
+    tensor = input.get(key)
+    _, unique_indices, counts = torch.unique(
+        tensor, dim=dim, sorted=True, return_inverse=True, return_counts=True
+    )
+
+    # Find first occurrence of each index  (e.g. [0, 1, 3])
+    _, unique_indices_sorted = torch.sort(unique_indices, stable=True)
+    cum_sum = counts.cumsum(0)
+    cum_sum = torch.cat((torch.tensor([0], device=input.device), cum_sum[:-1]))
```
vmoens, line 1942:
```suggestion
    cum_sum = torch.cat((torch.zeros(1, device=input.device), cum_sum[:-1]))
```

tensordict/utils.py
```diff
@@ -1862,3 +1862,90 @@ def _index_preserve_data_ptr(index):
...
+        ... )
+        >>> assert (td == expected_output).all()
+    """
+    # Check if the key is a TensorDict
+    if key not in input.keys(include_nested=True):
+        raise KeyError(f"The key '{key}' does not exist in the TensorDict.")
+
+    # Check that the key points to a tensor
+    if not isinstance(input.get(key), torch.Tensor):
+        raise KeyError(f"The key '{key}' does not point to a tensor in the TensorDict.")
```
vmoens, line 1920:
```suggestion
    tensor = input.get(key, default=None)

    # Check if the key is a TensorDict
    if tensor is None:
        raise KeyError(f"The key '{key}' does not exist in the TensorDict.")

    # Check that the key points to a tensor
    if not isinstance(tensor, torch.Tensor):
        raise KeyError(f"The key '{key}' does not point to a tensor in the TensorDict.")
```

tensordict/utils.py
```diff
@@ -1862,3 +1862,90 @@ def _index_preserve_data_ptr(index):
...
+        raise ValueError(
+            f"The specified dimension '{dim}' is invalid for a TensorDict with batch size '{input.batch_size}'."
+        )
+
+    # Convert negative dimension to its positive equivalent
+    if dim < 0:
+        dim = batch_dims + dim
+
+    # Get indices of unique elements (e.g. [0, 1, 0, 2])
+    tensor = input.get(key)
```
vmoens, line 1934:
```suggestion
```

vmoens (commented):
some small suggestions

vmoens (approved):

PR #654, opened 2024-02-02
[Feature] Adds utils method isin

albertbou92:
## Description

A similar method to torch.isin but for TensorDicts.

## Motivation and Context

## Types of changes

## Checklist


tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
         logging.info(indent + os.path.basename(path))
 
 
+def isin(
```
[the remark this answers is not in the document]
vmoens: Maybe rename in `contains`?

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+    tensordict: TensorDictBase,
+    reference_tensordict: TensorDictBase,
+    key: str,
+    dim: int = 0,
+    invert: bool = False,
+):
```
vmoens, line 1860: can we type the output too?

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+    """Tests if each element of 'key' in tensordict 'dim' is also present in the reference_tensordict.
+
+    This function returns a boolean tensor of length  ´tensordict.batch_size[dim]´ that is True for elements in
+    'key' that are also present in the reference_tensordict. This function assumes that both tensordict and
+    reference_tensordict have the same batch size and contain the specified key, otherwise it will raise an error.
```
vmoens, line 1865:
```suggestion
    """Tests if each element of ``key`` in tensordict ``dim`` is also present in the reference_tensordict.

    This function returns a boolean tensor of length  ``input.batch_size[dim]`` that is ``True`` for elements in
    the entry ``key`` that are also present in the ``reference``. This function assumes that both ``input`` and
    ``reference`` have the same batch size and contain the specified entry, otherwise an error will be raised.
```

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+
+    Args:
+        tensordict (TensorDictBase): Input TensorDict.
+        reference_tensordict (TensorDictBase): Target TensorDict against which to test.
+        key (str): The key to test.
+        dim (int): The dimension along which to test. Default: 0
+        invert (bool):  If True, inverts the boolean return tensor, resulting in True values for elements
+            not in reference_tensordict. Default: False
```
vmoens, line 1873:
```suggestion
    Args:
        input (TensorDictBase): Input TensorDict.
        reference (TensorDictBase): Target TensorDict to compare to ``input``.
        key (NestedKey): The key to test.
        dim (int, optional): The dimension along which to test. Defaults to ``0``.
    
    Keyword Args:
        invert (bool, optional):  If ``True``, inverts the boolean return tensor, resulting in ``True`` values for elements
            not in ``reference``. Defaults to ``False``.
```

To me `invert` brings very little and should be removed. You could just to `~isin(td0, td1)` and it would give the same result for just 1 extra character. I tend to think about kwargs as bringing non-trivial functionalities.

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+
+    Returns:
+        out (Tensor): A boolean tensor of length ´tensordict.batch_size[dim]´ that is True for elements in
+            'key' that are also present in the reference_tensordict or False if invert is True.
```
vmoens, line 1877:
```suggestion
        out (Tensor): A boolean tensor of length ``input.batch_size[dim]`` that is ``True`` for elements in
            ``key`` that are also present in the ``reference`` or ``False`` if invert is ``True``.
```

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+        >>> td_ref = TensorDict(
+        ...     {
+        ...         "tensor1": torch.tensor([[1, 2, 3], [4, 5, 6], [10, 11, 12]]),
+        ...         "tensor2": torch.tensor([[10, 20], [30, 40], [50, 60]]),
+        ...     },
+        ...     batch_size=[3],
+        ... )
+        >>> in_reference = is_in_reference(td, td_ref, key="tensor1")
+        >>> expected_in_reference = torch.tensor([True, True, True, False])
+        >>> assert torch.equal(in_reference, expected_in_reference)
```
vmoens, line 1896: there's also torch.testing.assert_close IIRC

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+    """
+    # Check key is present in both tensordict and reference_tensordict
+    if key not in tensordict.keys(include_nested=True):
+        raise KeyError(f"Key '{key}' not found in tensordict.")
+    if key not in reference_tensordict.keys(include_nested=True):
+        raise KeyError(f"Key '{key}' not found in reference_tensordict.")
```
vmoens, line 1902: What I usually do (to check if the key is there only once and make things faster)
```suggestion
    reference_tensor = reference_tensordict.get(key, default=None)
    target_tensor = tensordict.get(key, default=None)

    if not isinstance(target_tensor, torch.Tensor):
        raise KeyError(f"Key '{key}' not found in tensordict or not a tensor.")
    if not isinstance(reference_tensor, torch.Tensor):
        raise KeyError(f"Key '{key}' not found in reference_tensordict or not a tensor.")
```

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+
+    # Check that both TensorDicts have the same number of dimensions
+    if len(tensordict.batch_size) != len(reference_tensordict.batch_size):
+        raise ValueError(
+            "The number of dimensions in the batch size of the tensordict and reference_tensordict must be the same."
+        )
+
+    # Check that dim is valid
+    if dim >= len(tensordict.batch_size):
```
vmoens, line 1911: let's make it work with neg dims too

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+    if key not in tensordict.keys(include_nested=True):
+        raise KeyError(f"Key '{key}' not found in tensordict.")
+    if key not in reference_tensordict.keys(include_nested=True):
+        raise KeyError(f"Key '{key}' not found in reference_tensordict.")
+
+    # Check that both TensorDicts have the same number of dimensions
+    if len(tensordict.batch_size) != len(reference_tensordict.batch_size):
+        raise ValueError(
+            "The number of dimensions in the batch size of the tensordict and reference_tensordict must be the same."
+        )
```
vmoens, line 1908: one should be greater than the other no?
like `tensordict.batch_size[dim] > reference_tensordict.batch_size[dim]`

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+
+    # Get the data
+    reference_tensor = reference_tensordict.get(key)
+    target_tensor = tensordict.get(key)
+    N = reference_tensor.shape[0]
+
+    # Find the common indices
+    cat_data = torch.cat([reference_tensor, target_tensor], dim=0)
+    _, unique_indices = torch.unique(cat_data, dim=0, sorted=True, return_inverse=True)
+    out = torch.isin(unique_indices[N:], unique_indices[:N], assume_unique=True)
```
vmoens, line 1924: I don't see where dim is used

vmoens (commented):
Thanks, great feature! I left some comments

tensordict/utils.py
albertbou92: the only important thing is that both TD have the same number of dimensions. How many elements are present in each dimension does not really matter. So we only check that `len(tensordict.batch_size) != len(reference_tensordict.batch_size)`

vmoens:
RE renaming in contains I think isin isn't too bad after all since it's the PT name

albertbou92:
> RE renaming in contains I think isin isn't too bad after all since it's the PT name

haha this is why I named it like that. I am fine with both.

tensordict/utils.py
```diff
@@ -1851,6 +1851,85 @@ def format_size(size):
...
+
+    This function returns a boolean tensor of length  ``input.batch_size[dim]`` that is ``True`` for elements in
+    the entry ``key`` that are also present in the ``reference``. This function assumes that both ``input`` and
+    ``reference`` have the same batch size and contain the specified entry, otherwise an error will be raised.
+
+    Args:
+        input (TensorDictBase): Input TensorDict.
+        reference (TensorDictBase): Target TensorDict against which to test.
+        key (Nestedkey): The key to test.
+        dim (int, optionl): The dimension along which to test. Default: 0
```
vmoens, line 1870:
```suggestion
        dim (int, optional): The dimension along which to test. Defaults to ``0``.
```

vmoens (commented):
LGTM, just 2 small changes

vmoens (approved):

PR #670, opened 2024-02-07
[BugFix] Faster to_module

vmoens:
This is a quick fix for to_module runspeed until we find a better way of doing this.

cc @matteobettini


vmoens:
image
@matteobettini this is some serious speedup

PR #693, opened 2024-02-26
[Feature] Warn when `reset_parameters_recursive` is a no-op

matteobettini:
Related to https://github.com/pytorch/rl/issues/1967

### DIFF
--- a/tensordict/nn/common.py
+++ b/tensordict/nn/common.py
@@ -13,10 +13,9 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tupl
 
 import torch
 from cloudpickle import dumps as cloudpickle_dumps, loads as cloudpickle_loads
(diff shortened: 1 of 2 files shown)


tensordict/nn/common.py
```diff
@@ -868,13 +870,16 @@ def reset_parameters_recursive(
...
+    def _reset_parameters(self, module: nn.Module) -> bool:
+        any_reset = False
         for child in module.children():
             if isinstance(child, nn.Module):
-                self._reset_parameters(child)
+                any_reset += self._reset_parameters(child)
 
             if hasattr(child, "reset_parameters"):
                 child.reset_parameters()
+                any_reset += True
```
vmoens:
```suggestion
                any_reset |= True
```

vmoens (approved):
LGTM thanks!

tensordict/nn/common.py
matteobettini: Oh fancy!

Issue #663, opened 2024-02-06
Why is Tensorclass implemented as a decorator?


khwstolle:
### Current Tensorclass interface
To define a Tensorclass, we could write:
```python
@tensorclass
class MyTensorclass:
    foo: Tensor
    bar: Tensor
```
The `@tensorclass` decorator then generates a dataclass-like container class, adding various methods to implement (parts of) the `TensorDictBase` abstraction. Notice that unlike dataclasses, where the `__init__` signature may entirely be inferred from the class attribute annotations, the current decorator also augments `__init__` with extra keyword arguments `batch_size`, `device` and `names`. 

Considering that the decorator has significant (side-)effects to the resulting class, I am confused on why the API has been defined in this way. In this issue, I ask for clarification on this design choice.

### Simplified interface (conceptual)
Specifically, would a simplified representation using subclassing not suffice? For example, consider the following notation:
```python
class MyTensorclass(Tensorclass):
    foo: Tensor
    bar: Tensor
```
The `Tensorclass` baseclass then has the same effects as `@tensorclass`, and also:
1. Enables a more straightforward implementation in `tensordict`
2. Works with builtins `issubclass` and `isinstance`
3. Is compatible with static type checking (e.g. extra keyword arguments `batch_size` etc.)
4. Is easily understood as canonical Python

### Polyfill
To further clarify the effects of the subclass above, here's a quick polyfill that allows the definition of tensorclasses using the subclass-based API above in the current decorator-based paradigm:
```python
from types import resolve_bases
from typing import dataclass_transform, TYPE_CHECKING, Sequence
from dataclasses import dataclass, KW_ONLY
import torch
from torch.types import Device
from tensordict import tensorclass

@dataclass_transform()
class _TensorclassMeta(type):
    def __new__(metacls, name, bases, ns, **kwds):
        bases = resolve_bases(bases)
        cls = super().__new__(metacls, name, tuple(bases), ns, **kwds)
        return tensorclass(cls)

@dataclass_transform()
class Tensorclass(metaclass=_TensorclassMeta):  # Or: TensorDictBase subclass
    # Demonstrates point (3) above
    if TYPE_CHECKING:  
        _: KW_ONLY
        batch_size: torch.Size | Sequence[int]
        device: Device | str | None = None
        ...  # Other  methods
```

vmoens:
Hey thanks for proposing this!
[...]

PR #699, opened 2024-03-05
[Feature] Allow non-tensordata to be shared across processes + memmap

### LINKED ISSUE #698: [BUG] Cannot override __getitem__ or __setitem__ for tensorclass with non-tensor data
## Describe the bug

As described [here](https://github.com/pytorch/tensordict/issues/145#issue-1519081305), it should be possible to override __getitem__ and __setitem__ to implement indexing for non-tensor data types.

## To Reproduce

Following the linked issue:

```python
import tensordict

@tensorclass
class MyClass:
    images: torch.Tensor
    captions: List[str]

    def __getitem__(self, item):
        c = super().__getitem__(item)
        c.captions = self.captions[item]
        return c

data = MyClass(torch.randn(2, 3, 64, 64), ["a", "b"], batch_size=[2])
print(data[0])
```

## Expected behavior

The printed class to contain `captions=["a"]`.

## Additional context

It appears that monkey patching works if I copy and modify `_getitem` from tensorclass.py.

```
MyClass.__getitem__ = _getitem
```

In addition, this might be outside the scope of this issue and a separate feature request, but it seems that even with this monkey patching, stacking/concat does not properly handle this non-tensor data. Is this something that is supported?

I very often have non-tensordata that is associated with batch elements [e.g., a caption for each image] and it would be very weird to cat/stack and have the captions be unchanged.

vmoens:
Also closes #698

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -412,8 +412,7 @@ def _memmap_(
                 metadata = {"_type": str(cls)}
                 to_pickle = {}
                 for key, value in _non_tensordict.items():
-                    if isinstance(value, multiprocessing.sharedctypes.Synchronized):
-                        value = value.value
+                    value = _from_shared_nontensor(value)
                     if _is_json_serializable(value):
                         metadata[key] = value
                     else:
@@ -1699,7 +1698,9 @@ def _share_memory_nontensor(data, manager: Manager):
         result.update(data)
         return result
     if isinstance(data, str):
-        return mp.Value(ctypes.c_wchar_p, data)
+        result = mp.Array(ctypes.c_char_p, 1)
+        result[0] = data.encode('utf-8')
+        return result
     if isinstance(data, list):
         result = manager.list()
         result.extend(data)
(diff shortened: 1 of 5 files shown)


tensordict/tensorclass.py
```diff
@@ -191,25 +195,41 @@ def __torch_function__(
...
+    if not hasattr(cls, "unbind"):
+        cls.unbind = _unbind
+    if not hasattr(cls, "state_dict"):
+        cls.state_dict = _state_dict
+    if not hasattr(cls, "load_state_dict"):
+        cls.load_state_dict = _load_state_dict
+    if not hasattr(cls, "_memmap_"):
+        cls._memmap_ = _memmap_
+    if not hasattr(cls, "share_memory_"):
+        cls.share_memory_ = _share_memory_
```
vmoens, line 217: Ideally we'd like that if the class has this attribute and the built-in calls `super().foo(...)`, then this falls back on the implementation we're setting here. 
I'm not sure of how to make this happen...

vmoens: The best if we don't do #663 (the main blocker is the python version on my side) would be to discourage usage of `super().smth()` in `tensorclasses` in favour of something different, like
```
@tensorclass
class A:
   X: torch.Tensor
   
   def __getitem__(self, item):
       # your code here
       return super_method(self, "__getitem__")(item)
```
and `super_method` would be defined within `tensorclass.py` as a method falling back on the custom `tensorclass` implementation

PR #707, opened 2024-03-08
[BugFix, Feature] tensorclass.to_dict and from_dict

vmoens:
Nested tensorclasses will need some form of autocasting to make sure that a dict within a dict is interpreted as the right class. 

The PR also proposes such auto-casting for tensor containers only (leaves are not auto-cast, as it is the case with tensordict).

The doc has been updated to account for this feature, and tests have been written for it too.

cc @maximilianigl


vmoens:
As planned this isn't as easy as anticipated lol
Will try to fix it during the weekend!

maximilianigl:
Absolutely no rush from my side, it's easy to work around. Just wanted to flag it.

PR #731, opened 2024-04-18
[BugFix] consistent use of non_blocking in tensordict and torch.Tensor

vmoens:
cc @albanD

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -168,6 +168,10 @@ class TensorDict(TensorDictBase):
             tensordict. If provided, its length must match the one of the
             ``batch_size``. Defaults to ``None`` (no dimension name, or ``None``
             for every dimension).
+        non_blocking (bool, optional): if ``True`` and a device is passed, the tensordict
+            is delivered without synchronization. If ``False``, a global synchronization
+            will be made before passing the tensordict, making sure that all the stored
+            tensors are ready to be read. Defaults to ``False``.
 
     Examples:
         >>> import torch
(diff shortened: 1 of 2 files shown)


tensordict/_td.py
```diff
@@ -168,6 +168,10 @@ class TensorDict(TensorDictBase):
             tensordict. If provided, its length must match the one of the
             ``batch_size``. Defaults to ``None`` (no dimension name, or ``None``
             for every dimension).
+        non_blocking (bool, optional): if ``True`` and a device is passed, the tensordict
+            is delivered without synchronization. If ``False``, a global synchronization
+            will be made before passing the tensordict, making sure that all the stored
```
albanD, line 173:
```suggestion
            will be made before delivering the tensordict, making sure that all the stored
```
To be consistent with with the sentence above

tensordict/_td.py
```diff
@@ -234,7 +240,14 @@ def __init__(
...
             if source is not None:
                 for key, value in source.items():
-                    self.set(key, value, non_blocking=non_blocking)
+                    self.set(key, value, non_blocking=True)
+        if not non_blocking and has_device:
+            if torch.cuda.is_available():
+                torch.cuda.synchronize()
+            elif torch.backends.mps.is_available():
+                torch.mps.synchronize()
+            elif torch.backends.mha.is_available():
```
albanD, line 249: This is not a device haha

vmoens:
Reopened in #734

PR #739, opened 2024-04-21
[Feature] use return_mask as a string in pad_sequence  

### LINKED ISSUE #726: [Feature Request] Using `pad_sequence(... return_mask=True)` leads to confusing debug sessions
## Motivation

Working on computer vision applications, I often encounter situations where certain object models contain a key called "masks", most commonly referring to something like semantic segmentation outputs.

The function `pad_sequence` has an argument `return_mask` which adds a key called also called "masks". This can lead to some confusing behavior or bugs, especially since this may go unnoticed and could be hard to trace back. 

Currently, I cannot control the key that the padding mask is being written to. In general, there is no clear way to separate/modify keys that TensorDict internally uses and my own user keys.

## Solution

Make the interface with `return_mask=True` return the valid-entry mask as a secondary output, e.g.
```python
def pad_sequence(..., return_mask=False) -> TensorDict: ...  # padded tensordict
def pad_sequence(..., return_mask=True) -> tuple[TensorDict, TensorDict]: ... # padded tensordict, valid masks
```

## Alternatives

1. Propose to have all internally used keys exposed as constants and prepended with an underscore, e.g. export a variable `KEY_MASK: Final = "_mask"`, to clearly differentiate internal keys from user keys. 
2. Propose to pass the key (with default "masks") as an argument, either by a value of `str` to `return_mask` or as a separate parameter.
3. Copy-paste the method to my own codebase and change the key.

## Additional context

I understand that this is a minor/niche request, but since PyTorch is central to many computer vision applications, I hope that the adoption of TensorDict could be helped by avoiding the most common keys that users would be inclined to use.

## Checklist

- [x ] I have checked that there is no similar issue in the repo (**required**)

dtsaras:
## Description

This PR augments the ```pad_sequence``` function by allowing the parameter ```return _mask``` to be either a ```bool ``` or ```str```. In case ```return_mask``` is a ```bool```, the function behaves as before, but if ```return_mask``` is a ```str```, then the provided string becomes the key for the padding masks instead of "masks".

Additionally, for debugging purposes, if the TensorDict contains scalar tensors i.e. ```torch.Size([])```, the function raises a runtime error that the method does not support padding scalar tensors. This is consistent with ```torch.nn.utils.rnn.pad_sequence```. In case, we would like to support this we could just unqueeze such tensors. 

## Motivation and Context

In case the user already has some masks within the TensorDict under a key "masks" this will prevent the issue of renaming the masks such that they won't be overwritten by this method.

`close #726`

## Types of changes

## Checklist

### DIFF
--- a/tensordict/functional.py
+++ b/tensordict/functional.py
@@ -97,7 +97,7 @@ def pad_sequence(
             written.
         device (device compatible type, optional): if provded, the device where the
             TensorDict output will be created.
-        return_mask (bool, optional): if ``True``, a "masks" entry will be returned.
+        return_mask (bool, str): if ``True``, a "masks" entry will be returned. If ``return_mask`` is a string, it will be return the masks and be used as the key for the masks entry.
             It contains a tensordict with the same structure as the stacked tensordict where every entry contains the mask of valid values with size ``torch.Size([stack_len, *new_shape])``,
             where `new_shape[pad_dim] = max_seq_length` and the rest of the `new_shape` matches the previous shape of the contained tensors.
 
--- a/tensordict/functional.py
+++ b/tensordict/functional.py
@@ -84,7 +84,7 @@ def pad_sequence(
     padding_value: float = 0.0,
     out: T | None = None,
     device: DeviceType | None = None,
-    return_mask: bool | None = False,
+    return_mask: bool | str = False,
 ) -> T:
     """Pads a list of tensordicts in order for them to be stacked together in a contiguous format.
 
@@ -132,6 +132,11 @@ def pad_sequence(
     if not list_of_tensordicts:
         raise RuntimeError("list_of_tensordicts cannot be empty")
 
+    masks_key = "masks"
+    if isinstance(return_mask, str):
+        masks_key = return_mask
+        return_mask = True
+
     # check that all tensordict match
     update_batch_size = True
     max_seq_length = float("-inf")
@@ -144,6 +149,10 @@ def pad_sequence(
 
         for key in keys:
             tensor_shape = td.get(key).shape
+
+            if len(tensor_shape) == 0:
+                raise RuntimeError("Cannot pad scalars")
+            
             pos_pad_dim = pad_dim if pad_dim >= 0 else len(tensor_shape) + pad_dim
 
             # track the maximum sequence length to update batch_size accordingly
@@ -160,7 +169,7 @@ def pad_sequence(
 
             if return_mask:
                 tmp_list_of_tensordicts[-1].set(
-                    ("masks", key),
+                    (masks_key, key),
                     torch.ones(mask_shape, dtype=torch.bool),
                 )
     if return_mask:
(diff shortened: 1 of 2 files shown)


tensordict/functional.py
```diff
@@ -97,7 +97,7 @@ def pad_sequence(
             written.
         device (device compatible type, optional): if provded, the device where the
             TensorDict output will be created.
-        return_mask (bool, optional): if ``True``, a "masks" entry will be returned.
+        return_mask (bool, str): if ``True``, a "masks" entry will be returned. If ``return_mask`` is a string, it will be return the masks and be used as the key for the masks entry.
```
vmoens, line 100: could be a nested key

tensordict/functional.py
```diff
@@ -132,6 +132,11 @@ def pad_sequence(
     if not list_of_tensordicts:
         raise RuntimeError("list_of_tensordicts cannot be empty")
 
+    masks_key = "masks"
+    if isinstance(return_mask, str):
+        masks_key = return_mask
+        return_mask = True
```
vmoens, line 138:
```suggestion
    masks_key = "masks"
    if not isinstance(return_mask, bool):
        masks_key = unravel_key(return_mask)
        return_mask = True
```

test/test_tensordict.py
```diff
@@ -1085,7 +1085,7 @@ def test_pad(self):
         assert torch.equal(padded_td["a"], expected_a)
         padded_td._check_batch_size()
 
-    @pytest.mark.parametrize("make_mask", [True, False])
+    @pytest.mark.parametrize("make_mask", [True, "bibbidi-bobbidi-boo", False])
```
vmoens, line 1088:
```suggestion
    @pytest.mark.parametrize("make_mask", [True, ("bibbidi", "bobbidi", "boo"), False])
```

vmoens (approved):
LGTM let's just generalize it to nested keys

vmoens:
> @vmoens I think it's good to merge, thanks for the suggestions. One thing I noticed is that there is a device parameter for the function that does not get used anywhere in the code. We should fix that maybe in a separate MR.

Definitely!

tensordict/functional.py
```diff
@@ -84,7 +84,7 @@ def pad_sequence(
     padding_value: float = 0.0,
     out: T | None = None,
     device: DeviceType | None = None,
-    return_mask: bool | None = False,
+    return_mask: bool | str = False,
```
vmoens, line 87:
```suggestion
    return_mask: bool | NestedKey = False,
```

tensordict/functional.py
vmoens, line 100:
```suggestion
        return_mask (bool or NestedKey, optional): if ``True``, a "masks" entry will be returned. If ``return_mask`` is a nested key (string or tuple of strings), it will be return the masks and be used as the key for the masks entry.
```

vmoens (approved):

PR #788, opened 2024-05-23
[Feature] to_namedtuple and from_namedtuple

vmoens:


shagunsodhani (approved):
Awesome - thank you!

vmoens:
I guess here a similar thing for tensorclasses could also be handy, I don't think `a_tensorclass.from_namedtuple` will do what it should

vmoens:
My bad this works OOB
```python
import torch

from tensordict import tensordict, tensorclass

@tensorclass
class tc:
    X: torch.Tensor
    y: str

data = tc(X=torch.rand(()), y="a string!")
nt = data.to_namedtuple()
print(data.from_namedtuple(nt))
```
I have no idea how though haha

PR #800, opened 2024-05-30
[Quality] Better nested detection in numpy()

vmoens:


tensordict/base.py
```diff
@@ -6698,9 +6698,11 @@ def numpy(self):
         as_dict = self.to_dict()
 
         def to_numpy(x):
+            if isinstance(x, torch.Tensor):
+                if x.is_nested:
+                    return tuple(_x.numpy() for _x in x)
+                return x.numpy()
             if hasattr(x, "numpy"):
-                if getattr(x, "is_nested", False):
```
shagunsodhani, line 6702: Curious but isnt this a better check for nesting ?

vmoens: IDK
I usually try to avoid `hasattr` bc it can run code under the hood (imagine a very expensive property, hasattr will run it all until it may get to an `AttributeError`)
Happy to revert if you think it was better the other way around

PR #809, opened 2024-06-10
[Feature] map_names for composite dists

vmoens:


vmoens:
Solves https://github.com/pytorch/rl/issues/2167

PR #814, opened 2024-06-12
[Feature] TensorDict.consolidate

vmoens:
## Description

Consolidates the storage of a tensordict in a single storage.
This is aimed at making serialization of a tensordict faster with mmap and pickle.

Micro-benchmarks: num_threads doesn't seem to help on devgpu (but there is some potential benefit if working with DTensors!)

```python
import pickle
import tempfile

import torch
import tqdm
from torch.utils.benchmark import Timer

data = {}
d = data
for _ in range(10):
    for i in range(10):
        d[str(i)] = torch.randn(500, 50)
    d["nested"] = {}
    d = d["nested"]

data = TensorDict(data)

with open("data.pickle", "wb") as f:
    print(Timer("pickle.dump(data, f)", globals=globals()).adaptive_autorange())
# <torch.utils.benchmark.utils.common.Measurement object at 0x7f11cc954550>
# pickle.dump(data, f)
#   Median: 21.36 ms
#   IQR:    1.04 ms (20.66 to 21.70)
#   4 measurements, 1 runs per measurement, 1 thread

data_c = data.consolidate()
print(Timer("data.consolidate()", globals=globals()).adaptive_autorange())
# <torch.utils.benchmark.utils.common.Measurement object at 0x7f11ed792560>
# data.consolidate()
#   Median: 7.17 ms
#   IQR:    0.46 ms (7.02 to 7.48)
#   4 measurements, 10 runs per measurement, 1 thread

print(Timer("data.consolidate(num_threads=8)", globals=globals()).adaptive_autorange())
# <torch.utils.benchmark.utils.common.Measurement object at 0x7f11ed792110>
# data.consolidate(num_threads=8)
#   Median: 36.11 ms
#   IQR:    0.53 ms (35.84 to 36.37)
#   4 measurements, 10 runs per measurement, 1 thread

assert torch.tensor([v.untyped_storage().data_ptr() for v in data_c.values(True, True)]).unique().numel() == 1

with open("data_c.pickle", "wb") as f:
    print(Timer("pickle.dump(data_c, f)", globals=globals()).adaptive_autorange())
# <torch.utils.benchmark.utils.common.Measurement object at 0x7f11cc955e40>
# pickle.dump(data_c, f)
#   Median: 6.97 ms
#   IQR:    0.11 ms (6.92 to 7.04)
#   4 measurements, 10 runs per measurement, 1 thread

with tempfile.NamedTemporaryFile() as f:
    print(Timer("data.consolidate(filename=f.name)", globals=globals()).adaptive_autorange())
# <torch.utils.benchmark.utils.common.Measurement object at 0x7f11ed792110>
# data.consolidate(filename=f.name)
#   Median: 15.30 ms
#   IQR:    0.29 ms (15.12 to 15.40)
#   4 measurements, 10 runs per measurement, 1 thread

with tempfile.NamedTemporaryFile() as f:
    print(Timer("data.consolidate(filename=f.name, num_threads=8)", globals=globals()).adaptive_autorange())
# <torch.utils.benchmark.utils.common.Measurement object at 0x7f11cc956230>
# data.consolidate(filename=f.name, num_threads=8)
#   Median: 46.35 ms
#   IQR:    2.13 ms (44.84 to 46.97)
#   4 measurements, 1 runs per measurement, 1 thread

```

This benchmark is a bit more realistic. 
https://gist.github.com/vmoens/205f5173b3cba297389916655d07d1ce
The outcome is that the fastest serialization of the model is `tensordict.memmap_(filepath, num_threads=8)` with 150ms execution time. The second fastest is `td.consolidate(filename=filepath)`. Unfortunately `torch.save` performs equally bad (about 450-500ms) for both state-dicts and consolidated TDs in this example.

cc @shagunsodhani @albanD @mikaylagawarecki @dstaay-fb @jsuarez5341 @dtsaras @jbschlosser


tensordict/base.py
```diff
@@ -2374,6 +2452,110 @@ def saved_path(self):
...
+
+        flat_metadata = {}
+
+        offsets = torch.tensor([0] + flat_size).cumsum(0)
+
+        def assign(k, v, start, stop, storage=storage):
+            offset = v.storage_offset()
+            stride = v.stride()
+            if offset or stride != 1:
+                content = v.clone().untyped_storage()
```
vmoens, line 2518: It would be nice to be able to get a view of the storage directly
Do we have that?
cc @albanD @mikaylagawarecki

vmoens:
TODO:
- [ ] Reduction for lazy stacks
- [ ] Tests for tensorclass
- [ ] Work on optimizing runtime of consolidate (esp given non tensor data)
- [ ] Edge cases (non-tensor data that doesn't have a .data attribute)

tensordict/base.py
albanD: Well a "view of a storage" has a name, it's a "Tensor" :D 
If you need a contiguous chunk of memory containing these values, you will need a copy indeed and clone() is the way to go!

Note that you might want to pass in memory_format=torch.contiguous_format if you want a row major storage. Otherwise, you might get a channel last here as the default is to preserve the memory format.

tensordict/_reductions.py
```diff
@@ -0,0 +1,152 @@
...
+        }
+        for key, _ in leaves.items():
+            total_key = (key,) if prefix is None else prefix + (key,)
+            if total_key[-1].startswith("<NJT>"):
+                nested_values = flat_key_values[total_key]
+                continue
+            if total_key[-1].startswith("<NJT_OFFSETS"):
+                offsets = flat_key_values[total_key]
+                key = key.replace("<NJT_OFFSETS>", "")
+                value = torch.nested.nested_tensor_from_jagged(nested_values, offsets)
```
jbschlosser, line 43: nit: eventually it'll likely be useful to support NJT's `lengths` as well, which is an optional set of metadata present for non-contiguous NJTs

vmoens: Do you have an example of such a thing?

jbschlosser: sure thing, here's a test case demonstrating construction with lengths: https://github.com/pytorch/pytorch/blob/d52684e9a8b95f8b0d06f0bf08e6a39846cb3ae6/test/test_nestedtensor.py#L4242

When both `offsets` (shape `B + 1` including the end offset) and `lengths` (shape `B`) are specified, the idea is that `offsets` point to the beginning of each batch item sequence and `lengths` indicates the length of each. This allows for representing NJTs that are "non-contiguous with holes". If only `offsets` is specified, the assumption is that the entire length between offsets is part of each batch item, which of course doesn't allow for such holes.

PR #824, opened 2024-06-20
[Feature] DETERMINISTIC interaction mode

vmoens:
cc @matteobettini

### DIFF
--- a/tensordict/nn/probabilistic.py
+++ b/tensordict/nn/probabilistic.py
@@ -28,6 +28,19 @@ __all__ = ["ProbabilisticTensorDictModule", "ProbabilisticTensorDictSequential"]
 
 
 class InteractionType(Enum):
+    """A list of possible interaction types with a distribution.
+
+    MODE, MEDIAN and MEAN point to the property / attribute with the same name.
+    RANDOM points to ``rsample()`` if that method exists or ``sample()`` if not.
+
+    DETERMINISTIC can be used as a generic fallback if ``MEAN`` or ``MODE`` are not guaranteed to
+    be analytically tractable. In such cases, a rude deterministic estimate can be used
+    in some cases even if it lacks a true algebraic meaning.
+    This value will trigger a query to the ``deterministic_sample`` attribute in the distribution
+    and if it does not exist, the ``mean`` will be used.
+
+    """
+
     MODE = auto()
     MEDIAN = auto()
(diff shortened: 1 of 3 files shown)


tensordict/nn/probabilistic.py
```diff
@@ -453,8 +454,26 @@ def _dist_sample(
             interaction_type = self.default_interaction_type
 
         if interaction_type is InteractionType.MODE:
```
matteobettini, line 456:
```suggestion
        if interaction_type is InteractionType.DETERMINISTIC:
```

tensordict/nn/probabilistic.py
```diff
@@ -453,8 +454,26 @@ def _dist_sample(
...
+                    warnings.warn(
+                        "deterministic_sample wasn't found when queried. "
+                        f"{type(self).__name__} is falling back on mean instead. "
+                        f"For better code quality and efficiency, make sure to either "
+                        f"provide a distribution with a deterministic_sample attribute or "
+                        f"to change the InteractionMode to the desired value.",
+                        category=UserWarning,
+                    )
+
+        if interaction_type is InteractionType.MODE:
```
matteobettini, line 476:
```suggestion
        elif interaction_type is InteractionType.MODE:
```

tensordict/nn/probabilistic.py
```diff
@@ -453,8 +454,26 @@ def _dist_sample(
...
-            if hasattr(dist, "get_mode"):
-                return dist.get_mode()
+            try:
+                return dist.deterministic_sample
+            except AttributeError:
+                try:
+                    return dist.mean
+                except AttributeError:
+                    raise NotImplementedError(
+                        f"method {type(dist)}.mode is not implemented."
```
matteobettini, line 464: maybe say you asked deterministic and we fall back to mean and did not find it

vmoens:
```suggestion
                        f"method {type(dist)}.deterministic_sample is not implemented."
```

vmoens: @matteobettini it's in the finally below

tensordict/nn/probabilistic.py
vmoens: not necessary, the function either returns or raises just before

PR #826, opened 2024-06-24
[Feature] Add TensorDict storage

mjlaali:
## Description

TensorDict storage is a utility function that allows to use external storage (e.g. storage in cpu memory) with in other modules 
in torch. 

## Motivation and Context

This helps to implement algorithms such MCTS (see https://github.com/mjlaali/torchrl_mcts) or RAG modules. 

cc @shagunsodhani @dtsaras


tensordict/nn/storage.py
```diff
@@ -0,0 +1,346 @@
...
+        num_bits: int,
+        device: torch.device,
+        dtype: torch.dtype,
+        convert_to_binary: bool,
+    ):
+        super().__init__()
+        self.convert_to_binary = convert_to_binary
+        self.bases = 2 ** torch.arange(num_bits - 1, -1, -1, device=device, dtype=dtype)
+        self.num_bits = num_bits
+        self.zero_tensor = torch.zeros((1,))
```
vmoens, line 148: shouldn't we put this on device? What's the purpose of this tensor?

tensordict/nn/storage.py
```diff
@@ -0,0 +1,346 @@
...
+        >>> value = TensorDict(
+        ...     {"index": torch.Tensor([[10], [20], [30], [40]])}, batch_size=(4,)
+        ... )
+        >>> tensor_dict_storage[index] = value
+        >>> tensor_dict_storage[index]
+        >>> assert torch.sum(tensor_dict_storage.contain(index)).item() == 4
+        >>> new_index = index.clone(True)
+        >>> new_index["key3"] = torch.Tensor([[4], [5], [6], [7]])
+        >>> retrieve_value = tensor_dict_storage[new_index]
+        >>> assert (retrieve_value["index"] == value["index"]).all()
```
vmoens, line 273: This example doesn't run. I think part of the problem is that aggregation_module in QueryModule falls back onto hash_module if not provided, but I fail to see the logic there...

vmoens (commented):
I did a very quick initial review but overall it looks marvellous!

Do we need all of these class attributes to be public?

Can we have a short example in each docstring?

tensordict/nn/storage.py
```diff
@@ -0,0 +1,341 @@
...
+        raise NotImplementedError
+
+    def __setitem__(self, key: K, value: V) -> None:
+        raise NotImplementedError
+
+    def __len__(self) -> int:
+        raise NotImplementedError
+
+    def contain(self, item: K) -> torch.Tensor:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
mjlaali: Yes, I will add that.

tensordict/nn/storage.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+class DynamicStorage(TensorStorage[torch.Tensor, torch.Tensor]):
+    """A Dynamic Tensor Storage.
+
+    This is a storage that save its tensors in cpu memories. It
+    expands as necessary.
+    """
+
+    def __init__(self, default_tensor: torch.Tensor):
```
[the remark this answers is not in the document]
mjlaali: we can switch to TensorDict, no strong feeling.

My thought process that torch.Tensor is more granular than TensorDict. In particular, we may have different storage for different keys of TensorDict.

tensordict/nn/storage.py
```diff
@@ -0,0 +1,346 @@
...
+        self.tensor_dict: Dict[int, torch.Tensor] = {}
+        self.default_tensor = default_tensor
+
+    def clear(self) -> None:
+        self.tensor_dict.clear()
+
+    def __getitem__(self, indices: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
mjlaali: Good questions. 

We can put some restriction on tensors, but this means we make some assumptions on storage. I hesitate to do so, because there are some storage (soft memory like attentions) that indices can be floats. 

I left this open for future extendibility.

tensordict/nn/storage.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def __len__(self) -> None:
+        return len(self.tensor_dict)
+
+    def contain(self, indices: torch.Tensor) -> torch.Tensor:
+        res: List[bool] = []
+        for index in indices.tolist():
+            res.append(index in self.tensor_dict)
+
+        return torch.tensor(res, dtype=torch.int64)
```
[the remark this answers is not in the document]
mjlaali: Fair point, I will update that

tensordict/nn/storage.py
mjlaali: yes, we should, will fix it.

We could instantiate it in L158, but I thought instantiate it in init would improve efficiency. 

Let me know your thoughts and I can move it there.

tensordict/nn/storage.py
mjlaali: I have copied this from a unit test, but I did some changes afterward that may break the code. Let me copy / past the `test_storage` unit test again.

mjlaali:
> Do we need all of these class attributes to be public?
I am not sure if I understood the question. If you mean abstraction over the storage, yes, the list of methods is extracted from what I need in MCTS implementation. 

> Can we have a short example in each docstring?
Sure

vmoens:
@mjlaali 

> I am not sure if I understood the question. If you mean abstraction over the storage, yes, the list of methods is extracted from what I need in MCTS implementation.

I just mean that in all these classes, any attrbute we think the public will / should not care about should be named with a leading underscore to let us change / remove them at will.
Making all attributes public is stringent as it means that we will need deprecation warnings every time we want to change the internals of a class.

mjlaali:
Only methods define in TensorStorage interface should be public, the rest is private. 

Regarding classes, I have developed two types of storage (Dynamic/Fixed Storage). For MCTS, we only need one of them.

tensordict/nn/storage.py
```diff
@@ -0,0 +1,421 @@
...
+    @abstractmethod
+    def __len__(self) -> int:
+        raise NotImplementedError
+
+    @abstractmethod
+    def contain(self, item: K) -> torch.Tensor:
+        raise NotImplementedError
+
+
+class DynamicStorage(TensorStorage[torch.Tensor, torch.Tensor]):
```
vmoens, line 48: what about making this or the parent class a nn.Module?
I'm suggesting this because there is a tensor `default_tensor` in it, and it may be useful to register it as a buffer such that calling `dyn_storage.to("cuda")` sends the tensor to the device.

tensordict/nn/storage.py
```diff
@@ -0,0 +1,421 @@
...
+    """A Dynamic Tensor Storage.
+
+    This is a storage that save its tensors in cpu memories. It
+    expands as necessary.
+
+    Examples:
+        >>> storage = DynamicStorage(default_tensor=torch.zeros((1,)))
+        >>> index = torch.randn((3,))
+        >>> value = torch.rand((2, 1))
```
vmoens, line 57: Do you want to broadcast this?
I changed the first dim such that it matches the index's but if we want broadcasting we should make sure it works that way.

vmoens:
I made some improvements in https://github.com/pytorch/tensordict/pull/826/commits/da61af843f56504aa2b99e61d44bd6f1f5e068f5, LMK if you think some should be reverted.

I wonder if `TensorDictStorage` could not be simplified if there is only one index key in the query_module:

```python
        >>> import torch
        >>> from typing import cast
        >>> query_module = QueryModule(
        ...     in_keys=["key1", "key2"],
        ...     index_key="index",
        ...     hash_module=SipHash(),
        ... )
        >>> embedding_storage = DynamicStorage(
        ...     default_tensor=torch.zeros((1,)),
        ... )
        >>> tensor_dict_storage = TensorDictStorage(
        ...     query_module=query_module,
        ...     # key_to_storage={"index": embedding_storage},  # We know it's "index" since there is only one index key
        ...     key_to_storage= embedding_storage,
        ... )
        >>> index = TensorDict(
        ...     {
        ...         "key1": torch.Tensor([[-1], [1], [3], [-3]]),
        ...         "key2": torch.Tensor([[0], [2], [4], [-4]]),
        ...     },
        ...     batch_size=(4,),
        ... )
        >>> # value = TensorDict({"index": torch.Tensor([[10], [20], [30], [40]])}, batch_size=(4,)) # We know it's "index" since there is only one index key
        >>>  # Use this instead
        >>> value = torch.Tensor([[10], [20], [30], [40]])
        >>> tensor_dict_storage[index] = value
        >>> # indexing returns a single tensor / tensordict since there is only one key
		>>> tensor_dict_storage[index]
        Tensor(shape=torch.Size([4, 1]), device=cpu, dtype=torch.float32, is_shared=False)},
```

but maybe I'm missing the point and in most cases we'll have way more than one index key?

mjlaali:
> I made some improvements in [da61af8](https://github.com/pytorch/tensordict/commit/da61af843f56504aa2b99e61d44bd6f1f5e068f5), LMK if you think some should be reverted.

Great work, thanks for the edits.

> I wonder if `TensorDictStorage` could not be simplified if there is only one index key in the query_module:
I think there is a confusion because my bad naming, I should not use `index` in the example. 

In particular, key_to_storage is not related to query_module. key_to_storage is about output tensordict (the storage of each output tensordict key) and query module is about input tensordict (how to create an indices from input tensordict). 

Let me know if this makes sense?

vmoens:
> In particular, `key_to_storage` is not related to `query_module`. `key_to_storage` is about _output_ tensordict (the storage of each output tensordict key) and `query_module` is about _input_ tensordict (how to create an indices from input tensordict).
> 
> Let me know if this makes sense?

Not sure I'm following sorry. Can you maybe give an example where "index" is used properly (as you say it's a bad naming) and where there are more keys in the `key_to_storage` such that it's apparent why we need a dict?

mjlaali:
This is an example that how key_to_storage allows us to have different storage with different output key. 

```
query_module = QueryModule(
  in_keys=["index_key1", "index_key2"],
  index_key="index",
  hash_module=SipHash(),
)

emd_1d = FixedStorage(
    torch.nn.Embedding(num_embeddings=50, embedding_dim=1),
)
emd_2d = FixedStorage(
    torch.nn.Embedding(num_embeddings=100, embedding_dim=2),
)

tensor_dict_storage = TensorDictStorage(
    query_module=query_module,
    key_to_storage={"out_key1": emd_1d, "out_key2": emd_2d}, 
)
index = TensorDict(
   {
        "index_key1": torch.Tensor([[-1], [1], [3], [-3]]),
        "index_key2": torch.Tensor([[0], [2], [4], [-4]]),
    },
   batch_size=(4,),
)
value = TensorDict(
   {
        "out_key1": torch.Tensor([[-1], [1], [3], [-3]]),
        "out_key2": torch.Tensor([[0, 1], [1, 2], [2, 3], [3, 4]]),
    },
   batch_size=(4,),
)
tensor_dict_storage[index] = value
```

In this case, we have to use different storage because the dim is forced to be the same for every index when we use FixedStorage.
DynamicStorage would not have this issue, thought, we have to internally keep different storage per output key (like what key_to_storage provides).

vmoens:
Ok got it, I'll play a bit with it to make sure I fully grasp that feature and come back to you!

vmoens:
Ok @mjlaali I think that I would pitch this feature as this:
TensorDictStorage allows you to use a tensordict as an index, to build a bijective function
```
TensorDict -> TensorDict
```
The bijective is important: we want that for a specific tensordict (e.g. a value of "observation" and a value of "action") we get one tensor out (with entries "a", "b" and "c" always having corresponding values). 
In principle the opposite should hold: as every {"a", "b", "c"} combination there should be one {"observation", "action" associated.
I think that the example should have 2 entries on one side and 3 entries on the other to make it clear that it's not that input0 points to output0 and input1 to output1, but (input0, input1) point to (output0, output1).

For this feature to work we need to parse the tensordict used as index into a "real" long tensor index. We provide a SipHash class to do that along with a query module. This is our TensorDict2Index pipeline.

Then we need to find a way to store values and associate them with an long-tensor index. We can do that via DynamicStorage which writes a tensor for each index value (we assume that the index is a single integer). 
A FixedStorage relies on nn.embedding and will write on the weights of that module. 
We can have multiple storages associated with our module, and if we have a nested key associated with each of these we can combine the results in a TensorDict. 
This is the Index2TensorDict part. Joining the TensorDict2Index and Index2TensorDict gives us a TensorDict2TensorDict.

The most crucial thing I can think of right now is what are the limitations and assumptions, and in which situations will this fail to be efficient. 
Are we limited in size, ie can we use billions of tensordicts as indices or will this break? Which class will work best in large dimensional setting? These info should be put in the docstrings (like: "this module is particularly fit for settings where there are less than X observations" or "this module is particularly fit for settings where the observations have fewer than X elements").
What are the constrains in term of input dimension? Can I use an image as in_keys in my tensordict storage? Should we just flatten it? If we have images, maybe we should also give the opportunity to people to write their own embedding (maybe it's trivial using TensorDictModules...)

After giving a thorough look at all this I think that TensorDictMap is more appropriate:
1. There seems to be a confusion about storage because we use Storage in the Index2Tensor part but the main module is called TensorDictStorage
2. The term `storage` in pytorch usually refers to the content of a tensor. Since we now have `tensordict.consolidate()` which plays a lot with the storage of its content, using the term storage here may be confusing.

About that, I also think that the term is a bit overloaded here as DynamicStorage is about the Index2TensorDict part whereas TensorDictStorage refers to the whole `TensorDict2TensorDict` pipeline. I think that using Map could clarify that.

tensordict/nn/storage.py
```diff
@@ -0,0 +1,684 @@
...
+            TensorDict(
+                fields={
+                    done: Tensor(shape=torch.Size([35, 1]), device=cpu, dtype=torch.bool, is_shared=False),
+                    observation: Tensor(shape=torch.Size([35, 4]), device=cpu, dtype=torch.float32, is_shared=False),
+                    reward: Tensor(shape=torch.Size([35, 1]), device=cpu, dtype=torch.float32, is_shared=False),
+                    terminated: Tensor(shape=torch.Size([35, 1]), device=cpu, dtype=torch.bool, is_shared=False),
+                    truncated: Tensor(shape=torch.Size([35, 1]), device=cpu, dtype=torch.bool, is_shared=False)},
+                batch_size=torch.Size([35]),
+                device=None,
+                is_shared=False)
```
vmoens, line 608: I'd like to have a better, standalone example, but this one is nice bc it's a bit "real-life" and makes the point clear about the storage

cc @dtsaras for info

mjlaali: > TensorDictStorage allows you to use a tensordict as an index, to build a bijective function
The bijective function is an interesting view of TensorDictStorage, though originally the reason I needed it in MCTS was to have a way to manipulate the output of a function (not necessarily the bijective function).

TensorDictModule is also a function, but it is difficult to change its output. I needed something like a memory/storage to save, retrieve, and *update* a value.

> For this feature to work we need to parse the tensordict used as index into a "real" long tensor index.
This is true for this implementation, but I hope we can extend this in the future so that we can support KNN search with the same abstraction, such as retrieving the K-closest items to the input. This can be quite useful for RAG LLM.

> TensorDict2Index and Index2TensorDict
I like these names, they are more intuitive 👍 

> The most crucial thing I can think of right now is what are the limitations and assumptions, and in which situations will this fail to be efficient.
To me, the biggest limitation is that it does not allow the gradient to flow to memory, at least in this implementation. The second issue could be performance-related, in particular, DynamicStorage and SipHash have both been implemented in CPU memory, so tensors have to move from GPU. (SipHash could have been implemented with PyTorch operations, so this can be alleviated in the future.) 

Finally, concurrency is another issue, especially in distributed training jobs.

> Are we limited in size, ie can we use billions of tensordicts as indices or will this break?
This a good question, it is quite interesting to see if we can wrap [FAISS](https://github.com/facebookresearch/faiss) in this api. 

> After giving a thorough look at all this I think that TensorDictMap is more appropriate

While I originally thought about this API as a memory/storage API (the main added value is the ease in *updating* a storage/memory value), I have no strong opinion here. Either name works.

mjlaali:
> TensorDictStorage allows you to use a tensordict as an index, to build a bijective function

The bijective function is an interesting view of TensorDictStorage, though originally the reason I needed it in MCTS was to have a way to manipulate the output of a function (not necessarily the bijective function).

TensorDictModule is also a function, but it is difficult to change its output. I needed something like a memory/storage to save, retrieve, and *update* a value.

> For this feature to work we need to parse the tensordict used as index into a "real" long tensor index.

This is true for this implementation, but I hope we can extend this in the future so that we can support KNN search with the same abstraction, such as retrieving the K-closest items to the input. This can be quite useful for RAG LLM.

> TensorDict2Index and Index2TensorDict

I like these names, they are more intuitive 👍 

> The most crucial thing I can think of right now is what are the limitations and assumptions, and in which situations will this fail to be efficient.

To me, the biggest limitation is that it does not allow the gradient to flow to memory, at least in this implementation. The second issue could be performance-related, in particular, DynamicStorage and SipHash have both been implemented in CPU memory, so tensors have to move from GPU. (SipHash could have been implemented with PyTorch operations, so this can be alleviated in the future.) 

Finally, concurrency is another issue, especially in distributed training jobs.

> Are we limited in size, ie can we use billions of tensordicts as indices or will this break?

This a good question, it is quite interesting to see if we can wrap [FAISS](https://github.com/facebookresearch/faiss) in this api. 

> After giving a thorough look at all this I think that TensorDictMap is more appropriate

While I originally thought about this API as a memory/storage API (the main added value is the ease in *updating* a storage/memory value), I have no strong opinion here. Either name works.

tensordict/nn/storage.py
```diff
@@ -280,13 +284,49 @@ class SipHash(torch.nn.Module):
...
+    def __init__(
+        self,
+        n_components=16,
+        projection_type: str = "gaussian",
+        dtype_cast=torch.float16,
+        **kwargs,
+    ):
+        super().__init__()
+        from sklearn.random_projection import (
+            GaussianRandomProjection,
```
mjlaali, line 354: Is this the same as MLP where the weight are initialized from gaussian distribution?

vmoens: No I believe there is some fitting to the data involved, where it's trying to find the best separation for your examples using the projection.

mjlaali: I did a quick check and GaussianRandomProjection does not seem to use X in the fit function, it only uses it to get input shape:
https://github.com/scikit-learn/scikit-learn/blob/70fdc843a4b8182d97a3508c1a426acc5e87e980/sklearn/random_projection.py#L364

tensordict/nn/storage.py
```diff
@@ -220,11 +243,36 @@ def __setitem__(self, item: torch.Tensor, value: torch.Tensor) -> None:
...
         return torch.sum(self.flag).item()
 
-    def contains(self, item: torch.Tensor) -> torch.Tensor:
-        index = self._to_index(item)
+    def contains(self, value: torch.Tensor) -> torch.Tensor:
+        index = self._to_index(value)
         return self.flag[index]
 
 
+class LazyFixedStorage(FixedStorage):
```
mjlaali, line 251: Does it make to convert FixedStorage implementation such that by default it is lazy-initialized?

Is there any use case that we don't want lazy-initialization?

vmoens: I had the same thought, and actually that train of thought went even further... let me elaborate.

I refactored a bit FixedStorage such that we don't use modulo to get the index transform but a dict that maps the hash value to an int from 0 to M-1 where M is the max number of items we are going to store.

Also I think we don't need to rely on the Embedding class, we could just have a single tensor (akin to the weight that we're modifying) and assign values to it. That would make one less "code dependence" and clarify things IMO.

Now let's put these things together: (1) we need to convert any sparse index from 0 to N-1 with N >> M to an index from 0 to M-1 and (2) once we have that "dense" index we need to write on a storage (possibly lazily).
That looks a lot like torchrl's TensorStorage. The DynamicStorage looks a lot like torchrl's ListStorage (provided that you add the index-to-index transform explained in (1)).

So I'm starting to wonder: we are coding something with the purpose of applying it to MCTS, and recoding something very similar to what we have in torchrl (at list the second block, the Index2TensorDict part). What's missing in torchrl is
1. The TensorDict2Index (how to map a tensordict to a hash value)
2. How to map the hash value to an index from 0 to M-1
but we have the storage part, and it's pretty neat (works with prioritized values, N-dimension, arbitrary pytrees...)

Not sure what the way forward is, but here are various versions of what we could do
1. Make sure that whatever API we propose here matches TorchRL's storages one, such that we can have a minimal amount of storages coded here and just use TorchRL's when we need to
2. move things to torchrl and make an API that leverages the storages but provides more flexibility in terms of indexing. Something like:
```python
num_elts = 100
t2t_storage = TD2TDStorage(
    storage=LazyTensorStorage(num_elts), 
    in_keys=["obs", "action"], 
    out_keys=None, 
    map=TensorToIndex()) # out_keys dynamically determined
```

(1) would be a bit weird if we don't make things well bc there will be some implicit dependency on TorchRL's API that will make little sense here, but it could make sense if we have other usages of this kind of storages than MCTS (can you name any?)
(2) would make more sense if we are only thinking about MCTS

mjlaali: ListStorage is pretty neat, I missed it in TorchRL. 

My typical strategy is to optimize for simplicity, in this case having the code in torchrl make many decisions more simple.  So my vote would be moving this PR to TorchRL.

At this moment, I don't have any other use case than MCTS. If future, if the same abstraction can be used to implement K-NN search, then we may have other use cases (e.g. RAG), thought I would not optimize for this potential use case.

vmoens:
Closing in favor of https://github.com/pytorch/rl/pull/2283

PR #838, opened 2024-06-25
[Test] Test FC of memmap save and load

vmoens:
cc @albanD @mikaylagawarecki


MateuszGuzek (changes requested):
I quickly checked and it seems that memmaps have write access for owner 
```
-rw-r--r--   test/artifacts/mmap_example/nested/bfloat16.memmap
```
I think it will be better to set it to read only to enable testing this functionality (should work as long as test are not run by root).

Otherwise LGTM!

vmoens:
Good point @MateuszGuzek thanks!

Issue #717, opened 2024-03-21
[BUG] Tensorclass.key() doesn't list non-tensor data.


maximilianigl:
## Describe the bug

Tensorclasses (in contrast to tensordicts) don't list non-tensor data when iterating through it with `.keys()` or `.items()`. This also affects e.g. `apply()`. 

## To Reproduce

```python
from tensordict import TensorDict
from tensordict.prototype import tensorclass
import torch

tensordict = TensorDict({"tensor": torch.ones(3), "string": "string"})

print(sorted(tensordict.keys()))
# Output: ['string', 'tensor']

@tensorclass
class MyTensorClass:
    tensor: torch.Tensor
    string: str


my_tensor_class = MyTensorClass(tensor=torch.ones(3), string="string")

print(sorted(my_tensor_class.keys()))
# Output: ['tensor'], i.e. it ignores non-tensor data.

print(my_tensor_class.apply(lambda x: print(x)))
# Output: tensor([1., 1., 1.])
```

## Expected behavior

For `keys()` and `items()` I'd expect it to iterate over non-tensor data. 
For `apply()` probably as well, but whether or not it iterates over non-tensor data could also be dependent on an input flag (as one might explicitly want to only apply something to tensors).

## System info

tensordict version: '0.4.0+b4c91e8'

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)
- [x] I have read the [documentation](https://github.com/pytorch/tensordict/tree/main/docs/) (**required**)
- [x] I have provided a minimal working example to reproduce the bug (**required**)

vmoens:
This used to be a design choice before we introduced NonTensorData

## Context

We initially thought it could be interesting to let tensorclass carry non-tensor data, but thought it was better to exclude it from the keys since things like `apply` or even any other op that iterates through the keys (`reshape`, `gather`, ...) would have been meaningless with non-tensor data.

Then we introduces the `NonTensorData` which is a simple subclass of the a tensorclass that can only carry non-tensor data. If you call `tensordict.keys(include_nested=True)` you will have a `NonTensorData` node appearing as if it had no leaves, but in reality it is a leaf. If you call `apply` over NonTensorData it's ok because it does not access the `data` field (it's not part of the keys). But becaue we hack through `__getitem__` and `__setitem__` to access `NonTensorData.data`, the situation is less clear (one would imagine that the non-tensor data is now part of the keys since we use the key to set or access the value).

## Solution

[...]

PR #841, opened 2024-06-27
[Refactor] Make all leaves in tensorclass part of `_tensordict`, except for NonTensorData

vmoens:
This is bc-breaking in the following ways:
1. non-tensor data are proper leaves in the tensorclass
```python
data = MyData(X=X, y=y, z="a string", batch_size=batch_size)
assert "z" in data.keys() # used to break
```

2. non-tensor data will be compared if a tensorclass is compared to another tc / td
```python
# Previously
z = "a striing!"
tensordict = TensorDict(
            {
                "X": X,
                "y": y,
            },
            batch_size=[3, 4],
)
data = MyData(X=X, y=y, z=z, batch_size=batch_size)
assert (tensordict == data).all()
# Now z needs to be part of the tensordict as it won't be ignored during comparison
tensordict = TensorDict(
            {
                "X": X,
                "y": y,
            },
            batch_size=[3, 4],
)
```
3. Non-tensor data following comparison is not None
```python
data0 = MyData(X=X, y=y, z="a string", batch_size=batch_size)
data1 = MyData(X=X, y=y, z="another string", batch_size=batch_size)
(data0 == data1).z # used to be None, now a TD with boolean values
```
4. when setting non-tensor values in-place will now return a `ValueError`, not `RuntimeError`
5. This now works BUT it will convert any `NonTensorData` in `data` in a `NonTensorStack` (since values depend on their location in the batch):
```python
data0 = MyData(X=X, y=y, z="a string", batch_size=batch_size)
data1 = MyData(X=X, y=y, z="another string", batch_size=batch_size)
data0[:2] = data1[:2]
data0.z # used to be a string, bc ignored by __setitem__, now a list
```

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -797,16 +797,6 @@ class TensorDict(TensorDictBase):
             subtd = None
             for value_key, item in value.items():
                 if value_key in keys:
-                    if is_non_tensor(item):
-                        dest = self._get_str(value_key, NO_DEFAULT)
-                        if (dest[index] != item).any():
-                            dest_stack = dest.maybe_to_stack()
-                            if dest_stack is not dest:
-                                dest_stack[index] = item
-                                self._set_str(
-                                    value_key, dest_stack, validated=True, inplace=False
-                                )
-                                continue
                     self._set_at_str(
                         value_key, item, index, validated=False, non_blocking=False
                     )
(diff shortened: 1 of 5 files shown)


vmoens:
@maximilianigl
I think this will resolve #717 but the price to pay is some bc-breaking changes - hopefully for the better.

PR #872, opened 2024-07-11
Removed global set_default_validate_args

mctigger:
## Description
Removed global set_default_validate_args

## Motivation and Context
Having it true deviates from the default PyTorch behavior and may result in errors not raised.

## Types of changes

## Checklist

### DIFF
--- a/tensordict/nn/distributions/continuous.py
+++ b/tensordict/nn/distributions/continuous.py
@@ -21,9 +21,6 @@ __all__ = [
     "Delta",
 ]
 
-# speeds up distribution construction
-D.Distribution.set_default_validate_args(False)
-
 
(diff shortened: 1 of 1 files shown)


vmoens (commented):
Thanks for this.

In principle I support this change!
The reason we had this was that within `ProbabilisticTensorDictModule`, setting this to False can speed up the construction of the dists.

What would you think of raising a warning if this is `True` in `ProbabilisticTensorDictModule.__init__` to let users know that they could do things faster? Maybe it's an overkill, not sure. Another option would be to put a `.. note::` in `ProbabilisticTensorDictModule` docstrings to let users know that setting this to `False` can speed up the construction of the dists at the expense of safety.

If you could add one of those I'd be happy to merge this!

PR #873, opened 2024-07-11
[Feature] Compile integration - basics

vmoens:
* #876
* #875
* #874
* __->__ #873


tensordict/_lazy.py
```diff
@@ -1552,7 +1548,7 @@ def _multithread_rebuild(
         # We know batch_size is None, this has been checked earlier
         batch_size: Sequence[int] | None = None,
         device: torch.device | None = NO_DEFAULT,
-        names: Sequence[str] | None = None,
+        names: Sequence[str] | None = NO_DEFAULT,
```
shagunsodhani, line 1551: Since the default value is not `None` anymore, maybe we can drop the `| None` bit ?

tensordict/base.py
```diff
@@ -4709,8 +4766,12 @@ def _values_list(
                 is_leaf=is_leaf,
                 collapse=collapse,
             )
-            source = dict(zip(keys, vals))
-            return [source[key] for key in sorting_keys]
+            if torch.compiler.is_dynamo_compiling():
+                key_to_index = {key: i for i, key in enumerate(keys)}
```
shagunsodhani, line 4770: Interesting so the `dict` construct does not work when compiling?

tensordict/_torch_func.py
```diff
@@ -537,21 +546,24 @@ def stack_fn(key, values, is_not_init, is_tensor):
                     return torch.stack(values, dim)
                 with _ErrorInteceptor(
                     key, "Attempted to stack tensors on different devices at key"
-                ):
+                ) if not torch.compiler.torch.compiler.is_dynamo_compiling() else contextlib.nullcontext():
```
shagunsodhani, line 549: typo, "torch.compiler" bit is repeated

vmoens: And still, the tests pass 🤪

tensordict/base.py
vmoens: @anijain2305 may have more context but my understanding is just that this isn't registered. There's an item in the TD tracker for this!

PR #883, opened 2024-07-12
[Feature] Compile - benchmarks

vmoens:
* __->__ #883


vmoens:
@jisaacso this is the last PR of the compile stack

It includes benchmarks for tensordict vs pytree in eager and compile. Ops like `copy` (test: `test_compile_copy_flat`) run faster in eager and compile with TD compared to pytree.

PR #884, opened 2024-07-12
[BugFix] Fix `pad_sequence` behavior for non-tensor attributes of tensorclass

### LINKED ISSUE #783: [BUG] `tensordict.pad_sequence` silently ignores non-tensor attributes in `tensorclass`es or `TensorDict`s
## Describe the bug

I have some `tensorclass`es that store an audio file, with some metadata including speaker id and utterance id. I would like to collate these `tensorclass`es to form a batch, however when I do so, the metadata is lost (the metadata from the first tensordict is kept for every item in the batch) and the user is not warned about this either.

## To Reproduce

Steps to reproduce the behavior.

```python
from tensordict import pad_sequence, TensorDict

d1 = TensorDict({'a': torch.tensor([0]), 'b': ['asd']})
d2 = TensorDict({'a': torch.tensor([0]), 'b': ['efg']})

pad_sequence([d1, d2])
```

```bash
TensorDict(
    fields={
        a: Tensor(shape=torch.Size([2, 1]), device=cpu, dtype=torch.int64, is_shared=False),
        b: NonTensorData(data=asd, batch_size=torch.Size([2]), device=None)},
    batch_size=torch.Size([2]),
    device=None,
    is_shared=False)
```

## Expected behavior

I should either get a properly joined tensordict, e.g., 
```
TensorDict(
    fields={
        a: Tensor(shape=torch.Size([2, 1]), device=cpu, dtype=torch.int64, is_shared=False),
        b: NonTensorData(data=['asd', 'efg'], batch_size=torch.Size([2]), device=None)}, # in a list same shape as the batch_size
    batch_size=torch.Size([2]),
    device=None,
    is_shared=False)
```
or `tensordict.pad_sequence` should warn the user that the metadata is being discarded.

## System info

```
tensordict-nightly              2024.5.18
```

## Reason and Possible fixes

If you know or suspect the reason for this bug, paste the code lines and suggest modifications.

## Checklist

kurtamohler:
## Description

Fix the way `pad_sequence` checks for non-tensor attributes of tensorclasses. Also raise an error if `pad_sequence` is given tensorclasses with `return_mask=True`, since we cannot add the "masks" attribute with `<tensorclass>.set()`.

## Motivation and Context

close #783

## Types of changes

## Checklist

### DIFF
--- a/tensordict/functional.py
+++ b/tensordict/functional.py
@@ -20,7 +20,14 @@ from tensordict.base import (
     T,
     TensorDictBase,
 )
-from tensordict.utils import _check_keys, _shape, DeviceType, is_non_tensor, unravel_key
+from tensordict.utils import (
+    _check_keys,
+    _shape,
+    DeviceType,
+    is_non_tensor,
+    is_tensorclass,
+    unravel_key,
+)
 
 
 def pad(tensordict: T, pad_size: Sequence[int], value: float = 0.0) -> T:
@@ -149,6 +156,13 @@ def pad_sequence(
     if not list_of_tensordicts:
         raise RuntimeError("list_of_tensordicts cannot be empty")
 
+    if return_mask and is_tensorclass(list_of_tensordicts[0]):
+        raise RuntimeError(
+            "Expected 'return_mask=False' when list_of_tensordicts contains "
+            "tensorclasses, but got 'return_mask=True'. If you want masks, "
+            "plase convert the tensorclasses to TensorDicts first."
+        )
+
     masks_key = "masks"
     if not isinstance(return_mask, bool):
         masks_key = unravel_key(return_mask)
@@ -161,9 +175,10 @@ def pad_sequence(
     list_of_dicts = [{} for _ in range(len(list_of_tensordicts))]
     keys_copy = list(keys)
     for i, td in enumerate(list_of_tensordicts):
+        is_tensorclass_ = is_tensorclass(td)
 
         for key in keys:
-            item = td.get(key)
+            item = td._tensordict.get(key) if is_tensorclass_ else td.get(key)
             list_of_dicts[i][key] = item
             if is_non_tensor(item):
                 continue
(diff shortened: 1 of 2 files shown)


tensordict/functional.py
```diff
@@ -149,6 +156,13 @@ def pad_sequence(
     if not list_of_tensordicts:
         raise RuntimeError("list_of_tensordicts cannot be empty")
 
+    if return_mask and is_tensorclass(list_of_tensordicts[0]):
+        raise RuntimeError(
```
vmoens, line 160: Yep that's also my take on this! 👍

tensordict/functional.py
```diff
@@ -163,7 +177,7 @@ def pad_sequence(
     for i, td in enumerate(list_of_tensordicts):
 
         for key in keys:
-            item = td.get(key)
+            item = td._tensordict.get(key) if is_tensorclass(td) else td.get(key)
```
vmoens, line 180: is_tensorclass is a bit too expensive, let's take this check out of the loop (isolate the tensordict we'll be using before iterating through the keys)

vmoens (commented):
Just one little thing to fix but otherwise LGTM!

tensordict/functional.py
```diff
@@ -161,9 +175,10 @@ def pad_sequence(
     list_of_dicts = [{} for _ in range(len(list_of_tensordicts))]
     keys_copy = list(keys)
     for i, td in enumerate(list_of_tensordicts):
+        is_tensorclass_ = is_tensorclass(td)
```
vmoens, line 178:
```suggestion
        if is_tensorclass(td):
            td = td._tensordict
```

tensordict/functional.py
```diff
@@ -161,9 +175,10 @@ def pad_sequence(
...
 
         for key in keys:
-            item = td.get(key)
+            item = td._tensordict.get(key) if is_tensorclass_ else td.get(key)
```
vmoens, line 181:
```suggestion
            item = td.get(key)
```

vmoens (approved):
Thanks for this!

PR #901, opened 2024-07-18
[Feature] `zero_grad` and `requires_grad_`

vmoens:


tensordict/base.py
```diff
@@ -1343,6 +1343,23 @@ def grad(self):
...
+        Args:
+            set_to_none (bool, optional): if ``True``, tensor.grad will be ``None``,
+                otherwise ``0``.
+                Defaults to ``True``.
+
+        """
+        if set_to_none:
+            for val in self._values_list(True, True, is_leaf=_NESTED_TENSORS_AS_LISTS):
+                val.grad = None
+            return
```
shagunsodhani, line 1358: shouldnt we return self here too ?

PR #966, opened 2024-08-14
[BugFix] Allow context-manager for to_module

vmoens:
Fixes the torch.compile compatibility with functional calls with tensordict using CMs:
```python
def call(x, td):
    with td.to_module(module):
       return module(x)
call_compiled = torch.compile(vmap(call, (None, 0)))
call_compiled(x, td)
```

cc @zou3519


vmoens:
Integrated in #975

PR #981, opened 2024-09-02
[Feature] Unify composite dist method signatures with other dists

albertbou92:
## Description

## Motivation and Context

## Types of changes

## Checklist

Add a method to compute the entropy for composite distributions


vmoens:
As mentioned [here](https://github.com/pytorch/rl/pull/2391#issuecomment-2326451634), for consistency we should have entropy and log_prob return tensors IMO

tensordict/nn/distributions/composite.py
```diff
@@ -147,19 +152,67 @@ def rsample(self, shape=None) -> TensorDictBase:
             shape + self.batch_shape,
         )
 
-    def log_prob(self, sample: TensorDictBase) -> TensorDictBase:
-        """Writes a ``<sample>_log_prob entry`` for each sample in the input tensordict, along with a ``"sample_log_prob"`` entry with the summed log-prob."""
+    def log_prob(self, sample: TensorDictBase) -> torch.Tensor:
```
vmoens, line 155: This is BC-breaking.
I would suggest the following:
- add a `aggregate_probabilities` argument in the constructor, default it to `None`
- if `None`, raise a warning saying that as of now (v0.5) we will default it to `False` but as of v0.7 it will be `True`. People can remove the warning by passing this value explicitly.
- if `False`, `log_prob` calls `log_prob_composite`. If `True`, it behaves like your implementation.
- The same argument controls the behaviour of `log_prob` and `entropy`.

tensordict/nn/distributions/composite.py
```diff
@@ -147,19 +167,71 @@ def rsample(self, shape=None) -> TensorDictBase:
...
+        if not self.aggregate_probabilities:
+            return self.entropy_composite(samples_mc, include_sum=True)
+        se = 0.0
+        for _, dist in self.dists.items():
+            try:
+                e = dist.entropy()
+            except NotImplementedError:
+                x = dist.rsample((samples_mc,))
+                e = -dist.log_prob(x).mean(0)
+            while e.ndim > len(self.batch_shape):
```
vmoens, line 209: this seems to lead to a different result than log-prob, which makes sense because we don't pass a TD. I think it should be documented though: each docstring should tell what shape of tensor output is to be expected.

tensordict/nn/distributions/composite.py
```diff
@@ -147,19 +167,71 @@ def rsample(self, shape=None) -> TensorDictBase:
...
+            lp = dist.log_prob(sample.get(name))
+            while lp.ndim > sample.ndim:
+                lp = lp.sum(-1)
+            slp = slp + lp
+        return slp
+
+    def log_prob_composite(
+        self, sample: TensorDictBase, include_sum=True
+    ) -> TensorDictBase:
+        """Writes a ``<sample>_log_prob`` entry for each sample in the input tensordict, along with a ``"sample_log_prob"`` entry with the summed log-prob."""
```
vmoens, line 185: This should point to `log_prob` and vice-versa (to let users know what else exists)

tensordict/nn/distributions/composite.py
```diff
@@ -147,19 +167,71 @@ def rsample(self, shape=None) -> TensorDictBase:
...
+            except NotImplementedError:
+                x = dist.rsample((samples_mc,))
+                e = -dist.log_prob(x).mean(0)
+            while e.ndim > len(self.batch_shape):
+                e = e.sum(-1)
+            se = se + e
+        return se
+
+    def entropy_composite(self, samples_mc=1, include_sum=True) -> TensorDictBase:
+        """Writes a ``<sample>_entropy`` entry for each sample in the input tensordict, along with a ``"entropy"`` entry with the summed entropies."""
```
vmoens, line 215: this should point to `entropy`

tensordict/nn/distributions/composite.py
```diff
@@ -147,19 +166,91 @@ def rsample(self, shape=None) -> TensorDictBase:
...
         return sample
 
+    def entropy(self, samples_mc=1) -> torch.Tensor | TensorDictBase:
+        """Computes and returns the summed entropies.
+
+        If ``self.aggregate_probabilities`` is ``True``, this method will return a single tensor with
+        the summed entropies. If ``self.aggregate_probabilities`` is ``False``, this method will call
+        the `:meth:`~.entropy_composite` method and return a tensordict with the entropies of each sample
+        in the input tensordict along with an ``entropy`` entry with the summed entropy. In both cases,
+        the output shape will match the shape of the distribution sample, without a batch dimension.
```
vmoens, line 214: isn't it the opposite, we only have the batch dimension, no feature dim?

albertbou92: So, for a single distribution, the entropy method returns a tensor with the shape of the distribution samples. 
In this case we have one tensor for each of the distributions of the composite distribution. 
Each output can have different shapes.
What we can do is to assume all outputs are independent and sum all entropies, returning then a tensor of size `torch.Size([])`.
And to retrieve the entropies of individual dists with their specific shape `composite_entropy` can be used.

tensordict/nn/distributions/composite.py
```diff
@@ -147,19 +166,91 @@ def rsample(self, shape=None) -> TensorDictBase:
...
         return sample
 
+    def entropy(self, samples_mc=1) -> torch.Tensor | TensorDictBase:
+        """Computes and returns the summed entropies.
+
+        If ``self.aggregate_probabilities`` is ``True``, this method will return a single tensor with
+        the summed entropies. If ``self.aggregate_probabilities`` is ``False``, this method will call
+        the `:meth:`~.entropy_composite` method and return a tensordict with the entropies of each sample
+        in the input tensordict along with an ``entropy`` entry with the summed entropy. In both cases,
+        the output shape will match the shape of the distribution batch_shape.
```
vmoens, line 214:
```suggestion
        the output shape will match the shape of the distribution ``batch_shape``.
```

vmoens (commented):
LGTM thanks!

vmoens:
Looking at the errors, i realize that before we were adding both the single log-probs and the aggregate, now people have to choose (which again is bc-breaking).
So we should default to a third possibility which is "both".
I'll arrange that

PR #1004, opened 2024-09-20
[BugFix] Fix parsing integer batch size in AOT

vmoens:
* __->__ #1004

### DIFF
--- a/test/test_compile.py
+++ b/test/test_compile.py
@@ -774,16 +774,17 @@ class TestFunctional:
 
 
 @pytest.mark.skipif(not _v2_5, reason="Requires PT>=2.5")
+@pytest.mark.parametrize("strict", [False, True])
 class TestExport:
-    def test_export_module(self):
+    def test_export_module(self, strict):
         torch._dynamo.reset_code_caches()
         tdm = Mod(lambda x, y: x * y, in_keys=["x", "y"], out_keys=["z"])
         x = torch.randn(3)
         y = torch.randn(3)
-        out = torch.export.export(tdm, args=(), kwargs={"x": x, "y": y})
+        out = torch.export.export(tdm, args=(), kwargs={"x": x, "y": y}, strict=strict)
         assert (out.module()(x=x, y=y) == tdm(x=x, y=y)).all()
 
-    def test_export_seq(self):
+    def test_export_seq(self, strict):
         torch._dynamo.reset_code_caches()
         tdm = Seq(
             Mod(lambda x, y: x * y, in_keys=["x", "y"], out_keys=["z"]),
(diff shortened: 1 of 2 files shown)


test/test_compile.py
```diff
@@ -774,26 +774,65 @@ def call(x, td):
...
+            }
+        else:
+            kwargs = {}
+
+        result = torch.export.export(test, args=(x, y), strict=False, **kwargs)
+        export_mod = result.module()
+        x_new, y_new = torch.zeros(5, 100), torch.zeros(5, 100)
+        export_test = export_mod(x_new, y_new)
+        eager_test = test(x_new, y_new)
+        assert eager_test.batch_size == export_test.batch_size
```
vmoens, line 833: @ezyang this seems to be lost when using dynamic shape.
The batch size `[s0]` becomes `[]` when using dynamic shapes and when the 2nd output shape mismatches the 1st. 
We do get a warning though
```
W0920 10:19:28.564000 20340 torch/fx/experimental/symbolic_shapes.py:5136] Ignored guard Eq(s0, 5) == False, this could result in accuracy problems
```

ezyang: So, there's something a bit nontrivial going on here. In torch.compile eager, if we produce a fresh TensorDict and that TensorDict holds a list of dynamic ints, then in the residual bytecode we have to construct the TensorDict and also put in the freshly computed dynamic shapes from the FX graph (that has some int outputs now). So actually building a TensorDict isn't just a matter of putting in the right tensors, you also have to put some ints in too. Does this work?

Assuming this does work, export also has to be setup to do the same thing as well. It wouldn't be surprising if it didn't. In particular, if all export is doing is a pytree unflatten on Tensor leaves, the batch size won't be modified at all. To address this, we need to fix the export bug. But I also saw the comment about TensorDict not being pytree-able, so I am uncertain about the status there.

If you want to workaround, perhaps batch size can store rank instead of size and lazily compute it from tensor if it's not set? Better to fix things though. Just not sure what you expect to work and not work.

vmoens: > Assuming this does work, export also has to be setup to do the same thing as well. It wouldn't be surprising if it didn't. In particular, if all export is doing is a pytree unflatten on Tensor leaves, the batch size won't be modified at all. To address this, we need to fix the export bug. But I also saw the comment about TensorDict not being pytree-able, so I am uncertain about the status there.



TensorDict is pytreeable but you can deactivate it, this is what the comment is about (don't do it or the test will fail)

vmoens: Here's what works and what doesn't
```python
        class Test(torch.nn.Module):
            def forward(self, x: torch.Tensor, y: torch.Tensor):
                return TensorDict(
                    {
                        "x": x,
                        "y": y,
                    },
                    batch_size=x.shape[0],
                )
     x, y = torch.zeros(5, 100), torch.zeros(5, 100)
     result = torch.export.export(test, args=(x, y), strict=False, dynamic_shapes={
                    "x": {0: torch.export.Dim("batch"), 1: torch.export.Dim("time")},
                    "y": {0: torch.export.Dim("batch"), 1: torch.export.Dim("time")},
                })
    result = torch.export.export(test, args=(x, y), strict=False, **kwargs)
    export_mod = result.module()
    x_new, y_new = torch.zeros(5, 100), torch.zeros(5, 100)
    export_test = export_mod(x_new, y_new)
    eager_test = test(x_new, y_new)
    assert torch.Size([5]) == eager_test.batch_size == export_test.batch_size # Works because x and x_new have the same shape

    x_new, y_new = torch.zeros(2, 100), torch.zeros(2, 100)
    export_test = export_mod(x_new, y_new)
    eager_test = test(x_new, y_new)
    assert torch.Size([2]) == eager_test.batch_size == export_test.batch_size # Fails! now export_test.batch_size is torch.Size([])
```
So it's a weird behaviour, the SymInt just vanished into thin air in the second case

ezyang (approved):

PR #1013, opened 2024-09-30
[Feature] better sync and instantiation of cudagraphs

vmoens:


tensordict/nn/cudagraphs.py
```diff
@@ -208,20 +208,33 @@ def _call(
...
                         if tensordict_out is not None:
                             kwargs["tensordict_out"] = tensordict_out
                         out = self.module(tensordict, *args, **kwargs)
                         if self._out_matches_in is None:
                             self._out_matches_in = out is tensordict
                     self.counter += self._has_cuda
-                    if self._warmup_stream is not None:
-                        torch.cuda.current_stream().wait_stream(self._warmup_stream)
+                    if self._has_cuda:
+                        torch.cuda.synchronize()
```
MichaelFishmanBDAII, line 235: @vmoens What's the purpose of this change? IIUC this is forcing more syncs than the old version. Was the old version causing bugs?

vmoens: I think so, data could be read before it was fully updated 
We can make this optional if you want to experiment

MichaelFishmanBDAII: Thanks for the offer. I don't think I'll need to experiment with this for now.

PR #1015, opened 2024-09-30
[Quality] Fewer recompiles with tensordict

vmoens:


tensordict/nn/cudagraphs.py
```diff
@@ -222,6 +222,8 @@ def _call(
                     return result
 
                 if not self._has_cuda or self.counter < self._warmup - 1:
+                    # We must clone the data because providing non-contiguous data will fail later when we clone
+                    tensordict = self._tensordict = tensordict.clone()
```
shagunsodhani, line 226: In this statement, we are making a clone of `tensordict` and assigning it to `tensordict` ? (ignoring the assignment to `self._tensordict` for now)

shagunsodhani: ohh interesting! Thank you for the explanation.

PR #1060, opened 2024-10-25
[Performance] Faster _is_tensor_collection in eager mode

vmoens:
* __->__ #1060

### DIFF
--- a/tensordict/base.py
+++ b/tensordict/base.py
@@ -11097,8 +11097,20 @@ def _register_tensor_class(cls):
     _ACCEPTED_CLASSES = tuple(_ACCEPTED_CLASSES)
 
 
-def _is_tensor_collection(datatype):
-    return issubclass(datatype, TensorDictBase) or _is_tensorclass(datatype)
+_TENSOR_COLLECTION_MEMO = {}
+
+
+def _is_tensor_collection(datatype: type) -> bool:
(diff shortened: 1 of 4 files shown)


tensordict/base.py
```diff
@@ -11097,8 +11097,20 @@ def _register_tensor_class(cls):
...
-def _is_tensor_collection(datatype):
-    return issubclass(datatype, TensorDictBase) or _is_tensorclass(datatype)
+_TENSOR_COLLECTION_MEMO = {}
+
+
+def _is_tensor_collection(datatype: type) -> bool:
+    is_dynamo = is_dynamo_compiling()
+    out = None
+    if not is_dynamo:
+        out = _TENSOR_COLLECTION_MEMO.get(datatype)
```
shagunsodhani, line 11107: Wondering why do we need the if-conditon. Does compile not support dict inserts ?

vmoens: It's to avoid re-compiles. 

In #1015 I completely removed these checks but the benchmarks in eager mode suffered a lot from this (basically any iteration over tensordict leaves got 10x slower).

So I'm partially reverting this. The compiler shouldn't care too much about this kind of optimization, + it would add a guard on `_TENSOR_COLLECTION_MEMO` which would be counter productive.

shagunsodhani: Cool - thank you!

PR #1062, opened 2024-10-26
[Feature] Ordinal distribution

louisfaury:
## Description

Adds an ordinally parametrised distribution from [[Tang & Agrawal, 2020]](https://arxiv.org/pdf/1901.10500).

## Motivation and Context

This parametrisation showed useful when learning distribution on finite sets that were obtained by discretising continuous sets. 

Note: We can provide an example, although it should probably live in the `torchrl` repository. 

## Types of changes

## Checklist

- [] My change requires a change to the documentation.

### DIFF
--- a/tensordict/nn/distributions/discrete.py
+++ b/tensordict/nn/distributions/discrete.py
@@ -9,6 +9,7 @@ from typing import Sequence
 
 import torch
 from torch import distributions as D
+from torch.nn import functional as F
 
 __all__ = [
     "OneHotCategorical",
@@ -87,3 +88,55 @@ class OneHotCategorical(D.Categorical):
         out = d.rsample(sample_shape)
         out.data.copy_((out == out.max(-1)[0].unsqueeze(-1)).to(out.dtype))
         return out
+
+
+class Ordinal(D.Categorical):
+    """
+    A discrete distribution for learning to sample from finite ordered sets.
+    It is defined in contrast with the `Categorical` distribution, which does
+    not impose any notion of proximity or ordering over its support's atoms.
+    The `Ordinal` distribution explicitly encodes those concepts, which is
+    useful for learning discrete sampling from continuous sets. See §5 of
+    [Tang & Agrawal, 2020](https://arxiv.org/pdf/1901.10500.pdf) for details.
+
+    Notes:
+        This class is mostly useful when you want to learn a distribution over
+        a finite set which is obtained by discretising a continuous set.
+    """
+    def __init__(self, scores: torch.Tensor):
+        """
+        Args:
+            scores: a tensor of shape [..., N] where N is the size of the set which supports the distributions.
+            Typically, the output of a neural network parametrising the distribution.
+        """
+        logits = _generate_ordinal_logits(scores)
+        super().__init__(logits=logits)
+
+
+class OneHotOrdinal(OneHotCategorical):
+    """The one-hot version of the :class:`~.Ordinal` distribution."""
+    def __init__(self, scores: torch.Tensor):
+        """
+        Args:
+            scores: a tensor of shape [..., N] where N is the size of the set which supports the distributions.
+            Typically, the output of a neural network parametrising the distribution.
+        """
+        logits = _generate_ordinal_logits(scores)
+        super().__init__(logits=logits)
+
+
+def _generate_ordinal_logits(scores: torch.Tensor) -> torch.Tensor:
+    """Implements Eq. 4 of [Tang & Agrawal, 2020](https://arxiv.org/pdf/1901.10500.pdf)."""
+    # Assigns Bernoulli-like probabilities for each class in the set
+    log_probs = F.logsigmoid(scores)
+    complementary_log_probs = F.logsigmoid(-scores)
+
+    # Total log-probability for being "larger than k"
+    larger_than_log_probs = log_probs.cumsum(dim=-1)
+
+    # Total log-probability for being "smaller than k"
+    smaller_than_log_probs = (
+        complementary_log_probs.flip(dims=[-1]).cumsum(dim=-1).flip(dims=[-1]) - complementary_log_probs
+    )
+
+    return larger_than_log_probs + smaller_than_log_probs
--- a/tensordict/nn/distributions/discrete.py
+++ b/tensordict/nn/distributions/discrete.py
@@ -97,26 +97,34 @@ class Ordinal(D.Categorical):
     not impose any notion of proximity or ordering over its support's atoms.
     The `Ordinal` distribution explicitly encodes those concepts, which is
     useful for learning discrete sampling from continuous sets. See §5 of
-    [Tang & Agrawal, 2020](https://arxiv.org/pdf/1901.10500.pdf) for details.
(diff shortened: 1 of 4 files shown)


tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+
+
+class Ordinal(D.Categorical):
+    """A discrete distribution for learning to sample from finite ordered sets.
+
+    It is defined in contrast with the `Categorical` distribution, which does
+    not impose any notion of proximity or ordering over its support's atoms.
+    The `Ordinal` distribution explicitly encodes those concepts, which is
+    useful for learning discrete sampling from continuous sets. See §5 of
+    [Tang & Agrawal, 2020](https://arxiv.org/pdf/1901.10500.pdf) for details.
```
vmoens, line 100: Not sure this will render in sphinx
```suggestion
    `Tang & Agrawal, 2020<https://arxiv.org/pdf/1901.10500.pdf>`_ for details.
```

tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+
+    Notes:
```
vmoens, line 102:
```suggestion
    .. note::
```

tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+        This class is mostly useful when you want to learn a distribution over
+        a finite set which is obtained by discretising a continuous set.
+    """
+
+    def __init__(self, scores: torch.Tensor):
+        """A discrete distribution for learning to sample from finite ordered sets.
+
+        Args:
+            scores: a tensor of shape [..., N] where N is the size of the set which supports the distributions.
+            Typically, the output of a neural network parametrising the distribution.
```
vmoens, line 112:
```suggestion
            scores (torch.Tensor): a tensor of shape [..., N] where N is the size of the set which supports the distributions.
                Typically, the output of a neural network parametrising the distribution.
```

tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+        """
+        logits = _generate_ordinal_logits(scores)
+        super().__init__(logits=logits)
+
+
+class OneHotOrdinal(OneHotCategorical):
+    """The one-hot version of the :class:`~.Ordinal` distribution."""
```
vmoens, line 119: I think this link will look for `OneHotOrdinal.Ordinal` but i could be wrong (sphinx is still a unpredictable to me)
```suggestion
    """The one-hot version of the :class:`~tensordict.nn.distributions.Ordinal` distribution.
    
    Args:
        scores (torch.Tensor): a tensor of shape [..., N] where N is the size of the set which supports the distributions.
            Typically, the output of a neural network parametrising the distribution.
    """
```

tensordict/nn/distributions/discrete.py
vmoens, line 112: We usually put the docstrings of `__init__` in the main dosctrings (see above)

tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+    It is defined in contrast with the `Categorical` distribution, which does
+    not impose any notion of proximity or ordering over its support's atoms.
+    The `Ordinal` distribution explicitly encodes those concepts, which is
+    useful for learning discrete sampling from continuous sets. See §5 of
+    [Tang & Agrawal, 2020](https://arxiv.org/pdf/1901.10500.pdf) for details.
+
+    Notes:
+        This class is mostly useful when you want to learn a distribution over
+        a finite set which is obtained by discretising a continuous set.
+    """
```
vmoens, line 105:
```suggestion
    Args:
        scores (torch.Tensor): a tensor of shape [..., N] where N is the size of the set which supports the distributions.
            Typically, the output of a neural network parametrising the distribution.
    """
```

tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+class OneHotOrdinal(OneHotCategorical):
+    """The one-hot version of the :class:`~.Ordinal` distribution."""
+
+    def __init__(self, scores: torch.Tensor):
+        """The one-hot version of the :class:`~.Ordinal` distribution.
+
+        Args:
+            scores: a tensor of shape [..., N] where N is the size of the set which supports the distributions.
+            Typically, the output of a neural network parametrising the distribution.
+        """
```
vmoens, line 127: ditto

tensordict/nn/distributions/discrete.py
```diff
@@ -87,3 +88,60 @@ def rsample(
...
+        logits = _generate_ordinal_logits(scores)
+        super().__init__(logits=logits)
+
+
+def _generate_ordinal_logits(scores: torch.Tensor) -> torch.Tensor:
+    """Implements Eq. 4 of [Tang & Agrawal, 2020](https://arxiv.org/pdf/1901.10500.pdf)."""
```
vmoens, line 133:
```suggestion
    """Implements Eq. 4 of `Tang & Agrawal, 2020<https://arxiv.org/pdf/1901.10500.pdf>`__."""
```
although it wont be rendered anyway

vmoens (commented):
Thanks a mil for this, lovely feature.

We need to add the new classes to the docs

https://github.com/pytorch/tensordict/blob/main/docs/source/reference/nn.rst

Random thought: why choosing tensordict to host these? I'm not against it but I'd like to make sure we put this in the right place straight away, bc moving primitives from one repo to another can be a painful surgery.

I know we have some legacy distributions in tensordict (like one hot) and tooling (like NormalParamExtractor) but I don't think it really makes sense. 
Wouldn't torchrl be a better fit?

I guess the question we need to ask ourselves is: say a user wants to get one of these distributions, are they more likely to install torchrl or tensordict and think "yeah that makes sense".

Otherwise LGTM!

louisfaury:
> We need to add the new classes to the docs

Missed it, will take care of it. 

> why choosing tensordict to host these?

is a question I asked myself before pushing this. I went for it almost "by default" since, as you mention, I found some support for other db here. I agree with you that it doesn't feel perfectly right. 

> say a user wants to get one of these distributions, are they more likely to install torchrl or tensordict and think "yeah that makes sense"

Afaik, the best place would actually be `torch.distributions`. The `Ordinal` distribution is not tied to RL; it can be very relevant in, _e.g._, a supervised learning context. That being said, I do agree that torchrl would already be a better place than tensordict. I'll merge your suggestions (thx, btx :) ) and then maybe close this one and move it to torchrl ? Wdyt ?

vmoens:
Yeah let's move it to rl then, sorry you have to do the work twice!
torch.distributions is slightly unattended atm and we're struggling to make it compatible with compile so I can't guarantee on a timeline for this to be merged there. Let's keep it on the back of our minds and maybe make a PR against pytorch/pytorch later!

louisfaury:
No worries at all -- actually, for some reason I missed / forgot that torchrl had a `modules/distributions` too, feeling pretty silly I didn't put that piece of code there directly. Closing now.

PR #1100, opened 2024-11-21
[Refactor] Avoid TDParams parameters and buffers construction when obvious + new constructor

vmoens:
* __->__ #1100

### DIFF
--- a/docs/source/distributed.rst
+++ b/docs/source/distributed.rst
@@ -131,7 +131,7 @@ Although the call to :obj:`rpc.rpc_sync` involved passing the entire tensordict,
 updating specific indices of this object and return it to the original worker,
 the execution of this snippet is extremely fast (even more so if the reference
 to the memory location is already passed beforehand, see `torchrl's distributed
-replay buffer documentation <https://github.com/pytorch/rl/blob/main/examples/distributed/distributed_replay_buffer.py>`_ to learn more).
+replay buffer documentation <https://github.com/pytorch/rl/blob/main/examples/distributed/replay_buffers/distributed_replay_buffer.py>`_ to learn more).
 
 The script contains additional RPC configuration steps that are beyond the
 purpose of this document.
--- a/docs/source/reference/tensorclass.rst
+++ b/docs/source/reference/tensorclass.rst
@@ -282,6 +282,7 @@ Here is an example:
     TensorClass
     NonTensorData
     NonTensorStack
+    from_dataclass
 
 Auto-casting
 ------------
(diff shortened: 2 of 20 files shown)


vmoens:
@kurt-stolle 
I'm investigating some issues with compile and I'd appretiate some help if you have time (ofc!)

This works:
```python
from tensordict import from_module, TensorDictParams, TensorDict
import torch.nn

module = torch.nn.Module()
module.params = torch.nn.Parameter(torch.randn(3))
params2 = from_module(module).data.clone()
params2 *= 0
params2 = TensorDictParams(params2)

@torch.compile(fullgraph=True)
def func(z, params2):
    with params2.to_module(module):
        out = z + module.params
    return out

print(func(torch.zeros(()), params2))
```

All of these don't
1. Using a plain tensordict: _dynamo doesn't like that we set attribute to a MutableMapping
```python
from tensordict import from_module, TensorDictParams, TensorDict
import torch.nn

module = torch.nn.Module()
module.params = torch.nn.Parameter(torch.randn(3))
params2 = from_module(module).data.clone()
params2 *= 0
params2 = TensorDictParams(params2)
# Isolate the inner tensordict
params2 = params2._param_td

@torch.compile(fullgraph=True)
def func(z, params2):
    with params2.to_module(module):
        out = z + module.params
    return out

print(func(torch.zeros(()), params2))
```

2. This doesn't work because we have a TensorDIctParams in our module (the error is somewhat similar to the one above)
```python
from tensordict import from_module, TensorDictParams, TensorDict
import torch.nn

module = torch.nn.Module()
module.params = TensorDictParams(
    # string="a string!",
    TensorDict(a=0.0)
)
params2 = from_module(module).data.clone()
params2 *= 0
params2 = TensorDictParams(params2)

@torch.compile(fullgraph=True)
def func(z, params2):
    with params2.to_module(module):
        out = z + module.params["a"]
    return out

print(func(torch.zeros(()), params2))
```

It's crucial that _dyanmo works fine with this kinds of ops. I suspect that just handling `setattr` for `MutableMapping` might be enough but I'm not sure.


(commenting on this PR as it's part of the effort to rationalize TensorDictParams, see also https://github.com/pytorch/pytorch/issues/141118 for a related issue)
cc @anijain2305

kurtamohler:
> @kurtamohler I'm investigating some issues with compile and I'd appretiate some help if you have time (ofc!)

I found a fix the first broken case (plain TensorDict). I'll push a PR to PyTorch by tomorrow--just need to figure out how to write a test that doesn't depend on tensordict. There are already a few tests in PyTorch that I can base it on

Issue #1099, opened 2024-11-20
[Feature Request] Support tuple values in TensorDictModule `in_keys` arguments.


bachdj-px:
## Motivation

One limitation that is found when passing a dictionary as `in_keys` to the current `TensorDictModule` is that a specific key in the input `TensorDict` cannot be used more than once in the wrapped function (or wrapped `Module`'s `forward`).

For instance, one can do:

```python
module = TensorDictModule(lambda x, *, y: x+y, in_keys={'1': 'x', '2': 'y'}, out_keys=['z'])
module(TensorDict({'1': torch.ones(()), '2': torch.ones(())*2}, []))
print(td['z']) # tensor(3.)
```
but the following will crash as expected:

```python
module = TensorDictModule(lambda x, *, y, t : x+y+t, in_keys={'1': 'x', '2': ('y', 't')}, out_keys=['z'])
module(TensorDict({'1': torch.ones(()), '2': torch.ones(())*2}, []))
# TypeError: keywords must be strings
```

I would clearly see cases where passing the same input value to 2 different arguments in the wrapped function/module would be desirable.

Could you clarify whether that is a specific design choice or whether there is any reason this feature would cause a problem?

## Solution

A simple solution would be to allow tuples and lists a values for the `in_keys` argument and parse accordingly. This can be [achieved](https://github.com/pytorch/tensordict/compare/main...bachdj-px:tensordict:support_tuple_in_keys_dispatch) by iterating over said values and building the lists of in_keys and keywords:

```python
if isinstance(in_keys, dict):
            # write the kwargs and create a list instead
            _in_keys = []
            self._kwargs = []
            for key, value in in_keys.items():
                self._kwargs.append(value)
                _in_keys.append(key)
                if isinstance(value, tuple) or isinstance(value, list):
                    for _value in value:
                        self._kwargs.append(_value)
                        _in_keys.append(key)
                else:
                    self._kwargs.append(value)
                    _in_keys.append(key)
            in_keys = _in_keys
```

In our second example, this would result in:

```python
in_keys = ["1", "2", "2"]
self._kwargs = ["x", "y", "t"]
```

## Additional context

I haven't identified undesired side effects that would come with this new support of various in_keys.

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

vmoens:
Completed by #1101

PR #1101, opened 2024-11-21
[FEATURE] TensorDictModule in_keys allowed as Dict[str, tuple | list] to enable multi use of a sample feature

bachdj-px:
## Description

See feature request #1099 

## Motivation and Context

Close issue #1099 

## Types of changes

- [x] Example inside the code documentation

## Checklist


test/test_nn.py
```diff
@@ -161,6 +161,14 @@ def fn(a, b=None, *, c=None):
             td = TensorDict({"1": torch.ones(1)}, [])
         assert (module(td)["a"] == 2).all()
 
+    def test_tuple_input_keys(self):
+        in_keys = {"1": "first", "2": ("second", "third")}
+        fn = lambda x, y, z: x + y + z
+        module = TensorDictModule(fn, in_keys=in_keys, out_keys=["out"])
+
+        kword_in_keys = sorted(zip(module._kwargs, module.in_keys))
+        assert kword_in_keys == [("first", "1"), ("second", "2"), ("third", "2")]
```
vmoens, line 170: I don't really get this example:
the modue is not run, what would the input tensordict look like?
Could the example be a bit more extensive?

tensordict/nn/common.py
```diff
@@ -912,13 +927,18 @@ class TensorDictModule(TensorDictModuleBase):
...
 
     def __init__(
         self,
         module: Callable,
-        in_keys: NestedKey | List[NestedKey] | Dict[NestedKey:str],
+        in_keys: (
+            NestedKey
+            | List[NestedKey]
+            | Dict[NestedKey:str]
+            | Dict[NestedKey:str, Tuple[str] | List[str]]
```
vmoens, line 940:
```suggestion
            | Dict[NestedKey:str, List[str]]
```

vmoens (commented):
Hello, thanks for this PR!

IIUC your change is to allow one element of the input tensordict to be used multiple times within the module inputs.

That would be a good addition.

I think the tests would need to be a bit more extensive. We also need to check this works with other TD modules, like `ProbabilisticTensorDictModule`.

test/test_nn.py
bachdj-px: Thanks, it's a good point - I updated that test. I also added another test that tests the integration of TensorDictModule within a ProbabilisticTensorDictSequential.

bachdj-px: @vmoens Thanks for your feedback.
Could you clarify how you would see the change with `ProbabilisticTensorDictModule`? It seems to me that this feature will only impact `TensorDictModule`. `ProbabilisticTensorDictModule` inherits from `TensorDictModuleBase` and is not affected. Also I don't see the use of implementing my suggested feature in `ProbabilisticTensorDictModule` - its input have to be unique as far I see it.
Please let me know your thoughts

vmoens: Sure
What if I want to do something along the line of 
```python
ProbabilisticTensorDictModule(
   in_keys={"hidden": ("concentration", "rate")},
   out_keys=["other"],
   distribution_class=torch.distributions.Gamma
)
```
with the [`gamma`](https://pytorch.org/docs/stable/distributions.html#torch.distributions.gamma.Gamma) distribution taking params `concentration` and `rate`.

As per the utility of doing such thing - I think it's up to the users. To me as developer is more a matter of keeping the API as uniform as we can. You could for instance decide to build 2 different Normal distributions with a `CompositeDistribution` class that share the same scale?

bachdj-px: Thanks for the explanations @vmoens . 
Sorry, I am still a bit confused - the API documentation states that ` If in_keys is a dictionary, the keys are the keys of the distribution and the values are the keys in the tensordict that will get match to the corresponding distribution keys.`. In your example, it looks to me that you should pass the in_keys dictionary in the reverse order. Passing the same tensordict key to 2 different distribution parameters seems to be already supported, for instance via:

```python
ProbabilisticTensorDictModule(
   in_keys={"concentration": "value_in_input_td", "rate": "value_in_input_td" },
   out_keys=["other"],
   distribution_class=torch.distributions.Gamma
)
```

We could change the API to match your example but I suppose that it would be a breaking API change, wouldn't it?

vmoens: Yeah i just realized that, so we are asking dicts that are flipped between TensorDictModule and ProbabilisticTensorDictModule?
That ain't good!
If we flip the dicts in TensorDictModule (to match Proba), maybe that would solve your problem altogether?

bachdj-px: Absolutely, that would change the problem altogether! And it's a relatively change to make (docs and tests I assume)
However, that would be a breaking change for users already using dictionary with TensorDictModule. How would you like to proceed?

vmoens: Yep that's BC breaking but it's for the best
Here is how I would procede in this case: what we want is that people passing `in_keys` now with a dict don't see things breaking but they get a warning, and we also want to make your life easy if you're using the new API. 
So I'd suggest to add a `out_to_in_map: bool | None = None` keyword arg in the module that says:
- if `None`, the current default is used but if `in_keys` is a dict, then a deprec warning is raised
- If `False` we use the current behaviour `input-key: output-key`, no warning is raised.
- If `True`, we use the new behaviour.

The warning must be explicit and tell the user that the default will change 2 versions from now (next is 0.7 so it will be in 0.9). To silence the warning pass `out_to_in_map=False` or `True`.

Does that makes sense? Should I take care of it?

bachdj-px: Yes, thanks, I like this approach, it's safe. I can push the change today and add some tests for the 3 cases (check the warning is raised etc)

tensordict/nn/common.py
```diff
@@ -923,17 +941,35 @@ def __init__(
...
+        if out_to_in_map is not None and not isinstance(in_keys, dict):
+            warnings.warn(
+                "out_to_in_map is not None but is only used when in_key` is a dictionary."
+            )
+
         if isinstance(in_keys, dict):
+            if out_to_in_map is None:
+                warnings.warn(
+                    "Using a dictionary in_keys without specifying out_to_in_map is deprecated."
+                    "Use out_to_in_map to indicate the ordering of the input keys.",
```
vmoens, line 958: I think we want to change the behavior - currently `out_to_in_map` is `False` by default but we would like to make it `True` eventually. This should be mentioned in the warning, as well as the version where the change will occur (v0.9).

tensordict/nn/common.py
```diff
@@ -773,11 +773,17 @@ class TensorDictModule(TensorDictModuleBase):
...
             to be read in the tensordict and its values must match the name of
-            the keyword argument in the function signature.
+            the keyword argument in the function signature. if `out_to_in_map`,
+            the mapping gets inverted so that the keys correspond to the keyword
+            arguments in the function signature.
         out_keys (iterable of str): keys to be written to the input tensordict. The length of out_keys must match the
             number of tensors returned by the embedded module. Using "_" as a key avoid writing tensor to output.
 
     Keyword Args:
+        out_to_in_map (bool or None, optional): if ``True``, `in_keys` is read as if the keys are the arguments keys of
```
vmoens, line 783:
```suggestion
        out_to_in_map (bool, optional): if ``True``, `in_keys` is read as if the keys are the arguments keys of
```

tensordict/nn/common.py
```diff
@@ -773,11 +773,17 @@ class TensorDictModule(TensorDictModuleBase):
...
+            the :meth:`~.forward` method and the values are the keys in the input :class:`~tensordict.TensorDict`. If
+            `False`, keys are considered go be the input keys and values the method's arguments. If `None` (default), the
+            behaviour is the same as for `False` but a deprecation warning is raised.
```
vmoens, line 786:
```suggestion
            `False`, keys are considered go be the input keys and values the method's arguments. Deaults to ``False``.

            .. warning:: The default value of `out_to_in_map` will change from `False` to `True` in the v0.9 release.
                        
```

vmoens (commented):
Thanks! Looks great! I'd just make the doc / warning a bit more informative. 

I don't think we should mention in the docstrings that the value can be None as this is just used internally to check if the user passed a value or not (documenting it as public API defies its purpose IMO).

bachdj-px:
Thanks for your feedback @vmoens ! I applied your suggestions - please let me know your thoughts
I am running the tests locally but many CUDA-related fail - they arise from `test_tensordict.py` and they seem not to be linked to the new feature. Could we run the CI to double check?

tensordict/nn/common.py
```diff
@@ -773,11 +773,19 @@ class TensorDictModule(TensorDictModuleBase):
             order given by the in_keys iterable.
             If ``in_keys`` is a dictionary, its keys must correspond to the key
             to be read in the tensordict and its values must match the name of
-            the keyword argument in the function signature.
+            the keyword argument in the function signature. If `out_to_in_map` is True,
```
vmoens, line 776:
```suggestion
            the keyword argument in the function signature. If `out_to_in_map` is ``True``,
```

tensordict/nn/common.py
```diff
@@ -773,11 +773,19 @@ class TensorDictModule(TensorDictModuleBase):
...
             number of tensors returned by the embedded module. Using "_" as a key avoid writing tensor to output.
 
     Keyword Args:
+        out_to_in_map (bool, optional): if ``True``, `in_keys` is read as if the keys are the arguments keys of
+            the :meth:`~.forward` method and the values are the keys in the input :class:`~tensordict.TensorDict`. If
+            `False` or `None` (default), keys are considered to be the input keys and values the method's arguments keys.
+
+            .. warning::
+                The default value of `out_to_in_map` will change from `False` to `True` in the v0.9 release.
         inplace (bool or string, optional): if ``True`` (default), the output of the module are written in the tensordict
```
vmoens, line 789:
```suggestion
                The default value of `out_to_in_map` will change from `False` to `True` in the v0.9 release.

        inplace (bool or string, optional): if ``True`` (default), the output of the module are written in the tensordict
```

tensordict/nn/common.py
```diff
@@ -923,17 +943,37 @@ def __init__(
...
         super().__init__()
 
+        if out_to_in_map is not None and not isinstance(in_keys, dict):
+            warnings.warn(
+                "out_to_in_map is not None but is only used when in_key is a dictionary."
+            )
+
         if isinstance(in_keys, dict):
+            if out_to_in_map is None:
+                warnings.warn(
```
vmoens, line 958:
```suggestion
            if out_to_in_map is None:
                out_to_in_map = False
                warnings.warn(
```

tensordict/nn/common.py
```diff
@@ -923,17 +943,37 @@ def __init__(
...
+                    "Using a dictionary in_keys without specifying out_to_in_map is deprecated."
+                    "By default, out_to_in_map is False (`in_keys` keys as tensordict keys), but from "
+                    "version>=0.9, default will be True (`in_keys` as func arg keys)."
+                    "Please use explicit out_to_in_map to indicate the ordering of the input keys. ",
+                    DeprecationWarning,
+                    stacklevel=2,
```
vmoens, line 964:
```suggestion
                    "Using a dictionary in_keys without specifying out_to_in_map is deprecated. "
                    "By default, out_to_in_map is `False` (`in_keys` keys as tensordict pointers, values as kwarg name), but from "
                    "version>=0.9, default will be `True` (`in_keys` keys as func kwarg name, values as tensordict pointers). "
                    "Please use explicit out_to_in_map to indicate the ordering of the input keys. ",
                    DeprecationWarning,
                    stacklevel=2,
```

vmoens (commented):
Almost there thanks so much

vmoens:
Thanks for this!

PR #1112, opened 2024-11-26
[Refactor] Make CompositeDistribution a tensordict-exclusive class

vmoens:
* #1114
* #1113
* __->__ #1112

In this PR, I propose a new vision for `CompositeDistributions` where in the future:
- log_prob will return a tensordict with log-probs if not asked otherwise
- the aggregate log_prob will not be computed. Users will have the opportunity to do that manually, for instance through 
  ```python
  lp = dist.log_prob(sample, inplace=False)
  lp = lp.sum(reduce=False)
  ```
- the tensordict returned will not be the one used as input but another one.

This implies:
- reverting the announced deprecation (FutureWarning) for `aggregate_probabilities` (cc @albertbou92). The idea is that we actually want `aggregate_probabilities` to be False because we want this class to be an exclusively tensordict-to-tensordict class and `aggregate_probabilities=True` by default would mean that we get tensors from log-prob/entropy if not asked otherwise.
- New params, `inplace` and `include_sum` now control the respective behaviours. Since we want these behaviours to change in the future, we ask users to pass the value explicitly to make sure everything will hold when we make these changes. 

cc @louisfaury

### DIFF
--- a/tensordict/nn/distributions/composite.py
+++ b/tensordict/nn/distributions/composite.py
@@ -314,7 +314,8 @@ class CompositeDistribution(d.Distribution):
                 "The current default is ``True`` but from v0.9 it will be changed to ``False``. Please adapt your call to `log_prob_composite` accordingly.",
                 category=DeprecationWarning,
             )
-        slp = 0.0
+        if include_sum:
+            slp = 0.0
         d = {}
         for name, dist in self.dists.items():
             d[_add_suffix(name, "_log_prob")] = lp = dist.log_prob(sample.get(name))
--- a/tensordict/nn/distributions/composite.py
+++ b/tensordict/nn/distributions/composite.py
@@ -13,44 +13,38 @@ from torch import distributions as d
 
 
 class CompositeDistribution(d.Distribution):
-    """A composition of distributions.
+    """A composite distribution that groups multiple distributions together using the TensorDict interface.
 
-    Groups distributions together with the TensorDict interface. Methods
-    (``log_prob_composite``, ``entropy_composite``, ``cdf``, ``icdf``, ``rsample``, ``sample`` etc.)
-    will return a tensordict, possibly modified in-place if the input was a tensordict.
+    This class allows for operations such as `log_prob_composite`, `entropy_composite`, `cdf`, `icdf`, `rsample`, and `sample`
+    to be performed on a collection of distributions, returning a TensorDict. The input TensorDict may be modified in-place.
 
     Args:
-        params (TensorDictBase): a nested key-tensor map where the root entries
-            point to the sample names, and the leaves are the distribution parameters.
-            Entry names must match those of ``distribution_map``.
-
-        distribution_map (Dict[NestedKey, Type[torch.distribution.Distribution]]):
-            indicated the distribution types to be used. The names of the distributions
-            will match the names of the samples in the tensordict.
+        params (TensorDictBase): A nested key-tensor map where the root entries correspond to sample names, and the leaves
+            are the distribution parameters. Entry names must match those specified in `distribution_map`.
+        distribution_map (Dict[NestedKey, Type[torch.distribution.Distribution]]): Specifies the distribution types to be used.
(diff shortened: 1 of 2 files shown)


tensordict/nn/distributions/composite.py
```diff
@@ -223,16 +221,32 @@ def rsample(self, shape=None) -> TensorDictBase:
...
 
         Args:
-            sample (TensorDictBase): the sample to compute the log probability.
+            sample (TensorDictBase): The input sample to compute the log probability for.
 
         Keyword Args:
             aggregate_probabilities (bool, optional): if provided, overrides the default ``aggregate_probabilities``
                 from the class.
+            include_sum (bool, optional): Whether to include the summed log-probability in the output TensorDict.
+                Defaults to ``self.inplace`` which is set through the class constructor (``True`` by default).
```
louisfaury, line 240:
```
              Defaults to ``self.include_sum`` which is set through the class constructor (``True`` by default).
```

tensordict/nn/distributions/composite.py
```diff
@@ -253,47 +269,105 @@ def log_prob(
...
+            warnings.warn(
+                "`inplace` wasn't set when building the `CompositeDistribution` or when calling log_prob_composite. "
+                "The current default is ``True`` but from v0.9 it will be changed to ``False``. Please adapt your call to `log_prob_composite` accordingly.",
+                category=DeprecationWarning,
+            )
+        if include_sum:
+            slp = 0.0
         d = {}
         for name, dist in self.dists.items():
             d[_add_suffix(name, "_log_prob")] = lp = dist.log_prob(sample.get(name))
```
louisfaury, line 321: Question: if I am to use `include_sum=False` to go through a custom aggregation downstream, should I also be in charge of setting the key for the log-probs -- instead of the default `_add_suffix(name, "_log_prob")` ?

tensordict/nn/distributions/composite.py
```diff
@@ -320,9 +427,10 @@ def entropy_composite(self, samples_mc=1, include_sum=True) -> TensorDictBase:
                 x = dist.rsample((samples_mc,))
                 e = -dist.log_prob(x).mean(0)
             d[_add_suffix(name, "_entropy")] = e
```
louisfaury, line 429: Ditto

tensordict/nn/distributions/composite.py
```diff
@@ -253,47 +269,105 @@ def log_prob(
...
-        """Writes a ``<sample>_log_prob`` entry for each sample in the input tensordict, along with a ``"sample_log_prob"`` entry with the summed log-prob.
+        """Computes the log-probability of each component in the input sample and return a TensorDict with individual log-probabilities.
+
+        Args:
+            sample (TensorDictBase): The input sample to compute the log probabilities for.
 
-        This method is called by the :meth:`~.log_prob` method when ``self.aggregate_probabilities`` is ``False``.
+        Keyword Args:
+            include_sum (bool, optional): Whether to include the summed log-probability in the output TensorDict.
+                Defaults to ``self.inplace`` which is set through the class constructor (``True`` by default).
```
louisfaury, line 285:
```
Defaults to ``self.include_sum`` which is set through the class constructor (``True`` by default).
```
instead a `self.inplace`

louisfaury (commented):
lgtm, thanks!

tensordict/nn/distributions/composite.py
vmoens: ok correcting these typos in follow-up PR

tensordict/nn/distributions/composite.py
vmoens: if you want different names and not the sum, I guess you could do `inplace=False`, `include_sum=False` and then rename?
Or we could allow you to pass a function that maps the name to a new log-prob name?
Open to suggestions

tensordict/nn/distributions/composite.py
vmoens: ditto

PR #1196, opened 2025-01-30
[BugFix] strong instead of weakref in compiled regions

vmoens:
* #1197
* __->__ #1196


vmoens:
That's actually a terrible solution: in the following code, the output of the `func` carries the orginal tensordict in its `_last_op` even though it's never going to be used:

```python
# file: tensordict_cycles.py
#
# /// script
# requires-python = "==3.12"
# dependencies = [
#     "tensordict==0.6.2",
#     "packaging",
# ]
# ///
import gc

import torch
from torch.utils.viz._cycles import warn_tensor_cycles

warn_tensor_cycles()

@torch.compile()
def func():
    for _ in range(5):
        td = TensorDict({"a": torch.zeros(42)}, device="cuda", batch_size=[42]).view(6, 7)
    return td

for _ in range(5):
    func()

gc.collect()
```

vmoens:
Ok corrected by adding a method that allows us to clear the weakrefs explicitly. 
I think that the proper way to solve this should be:
- On tensordict side, use weakref.finalize rather than/in addition to weakref.ref to keep track of the previous TD (finalize could tell TD to clear its ref automatically when it gets out of scope)
- On dynamo side, handle `weakref.ref` to dead objects in `reconstruct` as well as `weakref.finalize` (if at all possible)

cc @anijain2305

PR #1197, opened 2025-01-30
[BugFix] Same behaviour in compiled and non-compiled versions of _new_unsafe

vmoens:
* (to be filled)


vmoens:
@anijain2305 Just reporting a very minor gb: `self = cls.__new__(cls)` (which is a very hacky way of creating a python object I admit) causes a graph break. In general we don't care too much but in a very few cases people want to subclass TensorDict and then we need that call. It's far from being a high-pri issue :)

PR #1230, opened 2025-02-22
[BugFix] Fix non-deterministic key order in stack

vmoens:
* __->__ #1230

### DIFF
--- a/tensordict/_torch_func.py
+++ b/tensordict/_torch_func.py
@@ -626,7 +626,6 @@ def _stack(
                 key: stack_fn(key, values, is_not_init, is_tensor)
                 for key, (values, is_not_init, is_tensor) in out.items()
             }
-
             result = clz._new_unsafe(
                 out,
                 batch_size=LazyStackedTensorDict._compute_batch_size(
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1790,7 +1790,7 @@ def _check_keys(
                     f"got keys {keys} and {set(td.keys())} which are incompatible"
                 )
     if strict:
-        return keys
+        return list(keys)
     return keys_set
 
 
--- a/tensordict/utils.py
+++ b/tensordict/utils.py
@@ -1757,7 +1757,7 @@ def _check_keys(
     strict: bool = False,
     include_nested: bool = False,
     leaves_only: bool = False,
-) -> set[str]:
+) -> set[str] | list[str]:
     from tensordict.base import _is_leaf_nontensor
 
     if not len(list_of_tensordicts):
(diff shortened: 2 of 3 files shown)


tensordict/utils.py
```diff
@@ -1769,27 +1769,29 @@ def _check_keys(
...
             else:
                 k = set(k)
-            if k != keys:
+            if k != keys_set:
                 raise KeyError(
                     f"got keys {keys} and {set(td.keys())} which are incompatible"
                 )
-    return keys
+    if strict:
+        return keys
```
vmoens, line 1793: We should actually make it a list

tensordict/utils.py
```diff
@@ -1769,27 +1769,29 @@ def _check_keys(
...
+    return keys_set
```
vmoens, line 1794: If keys can be exclusive, their order becomes arbitrary

tensordict/utils.py
```diff
@@ -1769,27 +1769,29 @@ def _check_keys(
     )
     # TODO: compile doesn't like set() over an arbitrary object
     if is_compiling():
-        keys = {k for k in keys}  # noqa: C416
+        keys_set = {k for k in keys}  # noqa: C416
     else:
-        keys: set[str] = set(keys)
+        keys_set: set[str] = set(keys)
```
priba, line 1774: Out of curiosity, is it much more efficient using `set` rather than using always the other option? Or there is another reason?

tensordict/_torch_func.py
```diff
@@ -626,7 +626,6 @@ def stack_fn(key, values, is_not_init, is_tensor):
                 key: stack_fn(key, values, is_not_init, is_tensor)
                 for key, (values, is_not_init, is_tensor) in out.items()
             }
-
```
louisfaury, line 629: Not sure if this added space is on purpose.

tensordict/utils.py
louisfaury, line 1793:
```suggestion
        return list(keys)
```
pretty sure that's what you mean with your comment, but just to be on the safe side. Rn, the return type is not consistent with typing.

tensordict/utils.py
louisfaury: By curiosity, what are the downstream functions that would be impacted by this? In other words, in which context is `_check_keys(strict=False)` used?

test/test_tensorclass.py
```diff
@@ -1887,6 +1887,20 @@ class MyDataNested:
...
+    def test_stack_keyorder(self):
+
+        class MyTensorClass(TensorClass):
+            foo: Tensor
+            bar: Tensor
+
+        tc1 = MyTensorClass(foo=torch.zeros((1,)), bar=torch.ones((1,)))
+
+        for _ in range(10000):
+            assert list(torch.stack([tc1, tc1], dim=0)._tensordict.keys()) == [
```
louisfaury, line 1899:
```suggestion
            assert list(torch.stack([tc1, tc1], dim=0).keys()) == [
```

tensordict/utils.py
vmoens: torch.compile used to not understand set() that's all. I should check if it's still the case

tensordict/utils.py
vmoens: Yes when using lazy stacks iirc

test/test_tensorclass.py
vmoens: This was on purpose to avoid any artifacts caused by @tensorclass (if there had been any)

PR #1242, opened 2025-02-28
[Bugfix] Allow non Module in TensorDictModule when method is passed

mikaylagawarecki:
relax this check as [LLM](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/llm.py#L53) in vLLM does not subclass nn.Module and this should be allowed when method is a callable

* __->__ #1242


vmoens (approved):
LGTM thanks!

PR #1259, opened 2025-03-12
[Feature] Add init value option to AddStateIndependentNormalScale

lin-erica:
## Description

When using AddStateIndependentNormalScale (in our case specifically with scale_mapping relu) we'd like the option of initializing the scale to any value based on our configurations. I added the option of taking in an initial value. Since it's set to 0.0 for default, it should be the same as initializing to torch.zeros.

## Motivation and Context

## Types of changes

## Checklist


tensordict/nn/distributions/continuous.py
```diff
@@ -148,11 +151,11 @@ def __init__(
         self.scale_mapping = scale_mapping
         if make_param:
             self.state_independent_scale = torch.nn.Parameter(
-                torch.zeros(scale_shape, device=device)
+                torch.ones(scale_shape, device=device) * init_value
             )
         else:
             self.state_independent_scale = torch.nn.Buffer(
-                torch.zeros(scale_shape, device=device)
+                torch.ones(scale_shape, device=device) * init_value
```
vmoens, line 158:
```suggestion
                torch.full(scale_shape, init_value, dtype=torch.get_default_dtype(), device=device)
```

vmoens (commented):
nit: let's use full instead of ones

otherwise lgtm!

PR #1280, opened 2025-03-27
[Performance] tensor_only for tensorclass

vmoens:
* __->__ #1280

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -517,18 +517,20 @@ class _tensorclass_dec:
     frozen: bool
     nocast: bool
     shadow: bool
+    tensor_only: bool
 
     def __new__(
         cls,
         autocast: bool = False,
         frozen: bool = False,
         nocast: bool = False,
-        shadow: bool = False,
+            shadow: bool = False,
+            tensor_only: bool = False,
     ):
         if not isinstance(autocast, bool):
             clz = autocast
             self = super().__new__(cls)
-            self.__init__(autocast=False, frozen=False, nocast=False, shadow=False)
+            self.__init__(autocast=False, frozen=False, nocast=False, shadow=False, tensor_only=False)
             return self.__call__(clz)
         return super().__new__(cls)
 
@@ -537,7 +539,8 @@ class _tensorclass_dec:
         autocast: bool = False,
         frozen: bool = False,
         nocast: bool = False,
-        shadow: bool = False,
+            shadow: bool = False,
+            tensor_only: bool = False,
     ):
         if autocast and nocast:
             raise ValueError("autocast is exclusive with nocast.")
@@ -545,14 +548,16 @@ class _tensorclass_dec:
         self.frozen = frozen
         self.nocast = nocast
         self.shadow = shadow
+        self.tensor_only = tensor_only
 
     @dataclass_transform()
     def __call__(self, cls: T) -> T:
-        clz = _tensorclass(cls, frozen=self.frozen, shadow=self.shadow)
+        clz = _tensorclass(cls, frozen=self.frozen, shadow=self.shadow, tensor_only=self.tensor_only)
         clz._autocast = self.autocast
         clz._nocast = self.nocast
         clz._shadow = self.shadow
         clz._frozen = self.frozen
+        clz._tensor_only = self.tensor_only
         return clz
 
 
@@ -568,6 +573,7 @@ def from_dataclass(
     nocast: bool = False,
     inplace: bool = False,
     shadow: bool = False,
+tensor_only: bool=False,
     device: torch.device | None = None,
 ) -> Any:
     """Converts a dataclass instance or a type into a tensorclass instance or type, respectively.
@@ -587,6 +593,7 @@ def from_dataclass(
         frozen (bool, optional): If ``True``, the resulting class or instance will be immutable. Defaults to ``False``.
         autocast (bool, optional): If ``True``, enables automatic type casting for the resulting class or instance. Defaults to ``False``.
         nocast (bool, optional): If ``True``, disables any type casting for the resulting class or instance. Defaults to ``False``.
+        tensor_only (bool, optional): TODO
         inplace (bool, optional): If ``True``, the dataclass type passed will be modified in-place. Defaults to ``False``.
             Without effect if an instance is provided.
         device (torch.device, optional): The device on which the TensorDict will be created. Defaults to ``None``.
@@ -654,6 +661,7 @@ def from_dataclass(
         clz._nocast = nocast
         clz._shadow = shadow
         clz._frozen = frozen
+        clz._tensor_only = tensor_only
         return clz
 
     if not is_dataclass(obj):
@@ -669,6 +677,7 @@ def from_dataclass(
         clz._nocast = nocast
         clz._shadow = shadow
         clz._frozen = frozen
+        clz._tensor_only = tensor_only
     else:
         clz = dest_cls
     result = clz(**asdict(obj), batch_size=batch_size, device=device)
@@ -690,6 +699,7 @@ def tensorclass(
     frozen: bool = False,
     nocast: bool = False,
     shadow: bool = False,
+tensor_only: bool=False,
 ) -> T | None:
     """A decorator to create :obj:`tensorclass` classes.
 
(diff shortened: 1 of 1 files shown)


vmoens:
```python
import time

import torch
from tensordict import TensorClass

class A(TensorClass["tensor_only"]):
    x: torch.Tensor
    y: torch.Tensor


class B(TensorClass):
    x: torch.Tensor
    y: torch.Tensor

a = A(x=torch.randn(()), y=torch.randn(()))
b = B(x=torch.randn(()), y=torch.randn(()))
from torch.utils.benchmark import Timer
ta = Timer("a.x", globals=globals())
tb = Timer("b.x", globals=globals())
print(tb.adaptive_autorange())
print(ta.adaptive_autorange())

# double check
t0 = time.time()
for _ in range(1_000_000):
    a.x
print(time.time() - t0)
t0 = time.time()
for _ in range(1_000_000):
    b.x
print(time.time() - t0)

t0 = time.time()
for _ in range(1_000_000):
    a.mean
print(time.time() - t0)
t0 = time.time()
for _ in range(1_000_000):
    b.mean
print(time.time() - t0)
```

prints
```
<torch.utils.benchmark.utils.common.Measurement object at 0x32d9bcca0>
b.x
  Median: 496.79 ns
  IQR:    6.25 ns (494.38 to 500.62)
  21 measurements, 1000 runs per measurement, 1 thread
<torch.utils.benchmark.utils.common.Measurement object at 0x32d9bd420>
a.x
  Median: 306.46 ns
  IQR:    6.79 ns (303.08 to 309.88)
  33 measurements, 1000 runs per measurement, 1 thread
0.31830406188964844
0.5060598850250244
0.04136085510253906
0.04162406921386719

Process finished with exit code 0
```

tensordict/tensorclass.py
```diff
@@ -1235,13 +1298,43 @@ def get_parent_locals(cls, localns=localns):
             localns=localns,
             # globalns=globals(),
         )
+        if tensor_only:
+
+            def is_tensor_or_optional_tensor(type_hint):
```
antoinebrl, line 1303: I wonder if the localization of functions can also have a certain impact. Here, `is_tensor_or_optional_tensor` and `get_parent_locals` are redefined on every call to `_get_type_hints`. Maybe they should leave outside, but maybe it would only bring a marginal improvement.

tensordict/tensorclass.py
```diff
@@ -1252,22 +1345,22 @@ def get_parent_locals(cls, localns=localns):
...
+    #     # We simply differ the warning till _set() is called
+    #     cls._set_dict_warn_msg = (
+    #         "A TypeError occurred when trying to retrieve a type annotation. "
+    #         "This may be caused by annotations that use plain `|` instead of typing.Union "
+    #         "or typing.Optional which are supported. If you wish to use the feature "
+    #         "of setting dict as attributes with automapping to tensordict/tensorclass "
+    #         "(`my_obj.attr = dict(...)`), consider re-writing the tensorclass with "
+    #         "traditional type annotations."
+    #     )
+    #     cls._type_hints = None
```
antoinebrl, line 1363: Is this removal intentional?

tensordict/tensorclass.py
```diff
@@ -1493,11 +1602,13 @@ def _getattr(self, item: str, **kwargs) -> Any:
                 ):
                     return _from_shared_nontensor(out)
                 return out
+        _tensordict = self._tensordict
         out = _tensordict._get_str(item, NO_DEFAULT, **kwargs)
```
antoinebrl, line 1606: nit: can be inlined
```suggestion
        out = self._tensordict._get_str(item, NO_DEFAULT, **kwargs)
```

tensordict/tensorclass.py
```diff
@@ -1493,11 +1602,13 @@ def _getattr(self, item: str, **kwargs) -> Any:
...
         if is_non_tensor(out):
             return out.data if not isinstance(out, NonTensorStack) else out.tolist()
         return out
 
+    _tensordict = self._tensordict
     out = getattr(_tensordict, item, NO_DEFAULT)
```
antoinebrl, line 1612: ditto
```suggestion
    out = getattr(self._tensordict, item, NO_DEFAULT)
```

test/test_tensorclass.py
```diff
@@ -2937,6 +2943,65 @@ def __setattr__(self, key, value):
...
+            a: torch.IntTensor
+            b: torch.LongTensor
+            c: torch.Tensor | None = None
+            d: torch.Tensor | Union[torch.IntTensor, torch.LongTensor] | None = (
+                None  # noqa
+            )
+            e: Optional[torch.IntTensor] = None  # noqa
+            f: Optional[torch.IntTensor | None] = None  # noqa
+
+        with pytest.raises(TypeError):
```
antoinebrl, line 2983: Do you want to capture part of the message to guarantee it's the right error being raised?
```suggestion
        with pytest.raises(TypeError, match="..."):
```

test/test_tensorclass.py
```diff
@@ -2937,6 +2943,65 @@ def __setattr__(self, key, value):
         assert (s.b == 2).all()
 
 
+class TestTensorOnly:
```
antoinebrl, line 2946: Should this be valid?
```python
    class NestedTensorOnly(TensorClass["tensor_only"]):
        a: torch.Tensor
        b: torch.Tensor | None
        c: TensorOnly
```
Either way, supported or not, should a test be added for this scenario?

test/test_tensorclass.py
antoinebrl, line 2983: Do you want to capture part of the message to guarantee it's the right error being raised?
```suggestion
        with pytest.raises(TypeError, match="..."):
```

antoinebrl (approved):
Looking good! Thanks for adding benchmarks too 👍

tensordict/tensorclass.py
vmoens: this is within the definition of your tensorclass, which you presumably run only once in your code
```
@tensorclass(tensor_only=True)  # <-- here
class X:
     ....
if __name__ == "__main__":
     x = X(...)  # <-- not here
     x.attr  # <-- not here
```

tensordict/tensorclass.py
vmoens: removing

test/test_tensorclass.py
vmoens: good catch, i think we can allow tensorclasses

tensordict/tensorclass.py
```diff
@@ -856,20 +910,34 @@ def __torch_function__(
...
     if not hasattr(cls, "__torch_function__"):
         cls.__torch_function__ = classmethod(__torch_function__)
     cls.__getstate__ = _getstate
     cls.__setstate__ = _setstate
 
-    cls.__getattr__ = _getattr
+    if tensor_only:
+        cls.__getattr__ = _getattr_tensor_only
+    else:
+        cls.__getattr__ = _getattr
```
antoinebrl, line 933: very minor nit: as this is short the condition can be only on the value and not the assignment. Feel free to ignore
```suggestion
    cls.__getattr__ = _getattr_tensor_only if tensor_only else _getattr
```

tensordict/tensorclass.py
```diff
@@ -856,20 +910,34 @@ def __torch_function__(
...
+
+    cls.__setattr_parent__ = object.__setattr__
     if "__setattr__" not in cls.__dict__:
-        cls.__setattr__ = _setattr_wrapper(cls.__setattr__, expected_keys)
+        if not tensor_only:
+            cls.__setattr__ = _setattr
+        else:
+            cls.__setattr__ = _setattr_tensor_only
```
antoinebrl, line 940: ditto:
```suggestion
        cls.__setattr__ = _setattr if not tensor_only else _setattr_tensor_only
```

antoinebrl (approved):
Nice work! 🚀

benchmarks/common/common_ops_test.py
```diff
@@ -2,6 +2,7 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+from __future__ import annotations
```
antoinebrl, line 5: does this import need to be added to all files?

vmoens: In general it doesn't hurt and I'm always pissed when python 3.9 tests fail bc it's missing

PR #1282, opened 2025-04-01
[Refactor] Setting lists in tensordict instances

vmoens:
* __->__ #1282

### DIFF
--- a/docs/source/reference/tensordict.rst
+++ b/docs/source/reference/tensordict.rst
@@ -243,3 +243,5 @@ Utils
     parse_tensor_dict_string
     set_capture_non_tensor_stack
     set_lazy_legacy
+    set_list_to_stack
+    list_to_stack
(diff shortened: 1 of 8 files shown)


vmoens:
To use this feature:
```python
from tensordict import *
set_list_as_stack(True).set()
td = LazyStackedTensorDict(TensorDict(), TensorDict())
td["a"] = ["0", "1"]
td["b"] = [torch.rand((1,)), torch.rand((2,))]
```

I think this will not be a footgun. Very few tests had to be adapted.

Only lists (not iterables) are considered as batches - so a passed string will not be unbound. 

If you have a lazy stack (or a tensorclass containing a lazy stack) and you want to get ragged tensors, you can do
```python
data.get(key, as_nested_tensor=True, layout=torch.strided) # gives a nested tensor
data.get(key, as_padded_tensor=True, padding_side="left", padding_value=0)
data.get(key, as_list=True)
```


cc @Darktex @mikaylagawarecki

PR #1285, opened 2025-04-11
[BugFix] Fix TDParams compatibility with export

vmoens:
* __->__ #1285


vmoens:
Restrictions:
- Seems like vmap within exported code with functional calls is not supported yet
- using strict=True with functional calls (without vmap) also fails as the exported code fails to see that we swapped the params.

cc @matteobettini @jeguzzi

PR #1314, opened 2025-05-08
[BugFix] Fix CudaGraphModule streams

vmoens:
## Description

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/tensordict/nn/cudagraphs.py
+++ b/tensordict/nn/cudagraphs.py
@@ -233,6 +233,7 @@ class CudaGraphModule:
                         out = self.module(tensordict, *args, **kwargs)
                         if self._out_matches_in is None:
                             self._out_matches_in = out is tensordict
+                    torch.cuda.current_stream().wait_stream(self._warmup_stream)
                     self.counter += self._has_cuda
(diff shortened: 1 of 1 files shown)


tensordict/nn/cudagraphs.py
```diff
@@ -233,6 +233,7 @@ def _call(
                         out = self.module(tensordict, *args, **kwargs)
                         if self._out_matches_in is None:
                             self._out_matches_in = out is tensordict
+                    torch.cuda.current_stream().wait_stream(self._warmup_stream)
```
MichaelFishmanBDAII, line 236: Do we also need `self._warmup_stream_cm.wait_stream(torch.cuda.current_stream())` before the `with self._warmup_stream_cm`? In the [torch cuda semantics docs](https://docs.pytorch.org/docs/stable/notes/cuda.html#whole-network-capture), it looks like they make the warmup stream wait for the current stream before doing the warmup.

vmoens: could be!
Have a look at #1315, I'll add it there

PR #1384, opened 2025-07-18
[Feature] Allow in-place modification of lazy stacks

vmoens:
@vermouth1992 can you comment on this?


vermouth1992:
Thank you for such a prompt feature support! Really appreciate!

PR #1388, opened 2025-07-24
[Feature] TensorCollection parent class and better type annotation

vmoens:
@louisfaury I know you guys care about type annotation - is this sound?


louisfaury:
We do love type hinting – thanks for this!

vmoens:
Ok still a long way to go but the mypy complains have shrunk substantially now!

PR #1394, opened 2025-07-27
[BugFix] Fix MetaData assignment in tensorclasses

vmoens:


nightdr:
Thanks for the quick fix!

PR #1398, opened 2025-07-28
[BugFix] Fix ClassVar support in tensorclass

vmoens:
cc @alex-bene


alex-bene:
Seems to work for me!

PR #1400, opened 2025-07-28
[Feature] Added correct indexing for NonTensorData

### LINKED ISSUE #1399: [BUG] Indexing NonTensorData returns an unindexed tensordict
## Describe the bug

This issue is related to a weird behaviour while indexing tensordicts. It has some serious implications in pytorch/rl's MinariExperienceReplay dataset class, where this weird behaviour has some indirect implications in the incorrect download of NonTensorData for every dataset.

More details below.

## To Reproduce

Take the following code, for example:

```python
a = NonTensorData(np.array(["cuac", "woof"], dtype="object"), batch_size=2)
```

Now it is where things get a bit weird. If we wanted to index this NonTensorData instance, one would expect to get the first element of the data attribute, but it only addapts the batch_size, it does not index the contained data.

```markdown
In [45]: a
Out[45]: NonTensorData(data=['cuac' 'woof'], batch_size=torch.Size([2]), device=None)
In [46]: a.data
Out[46]: array(['cuac', 'woof'], dtype=object)
In [47]: a[0]
Out[47]: NonTensorData(data=['cuac' 'woof'], batch_size=torch.Size([]), device=None)
In [48]: a[0].data
Out[48]: array(['cuac', 'woof'], dtype=object)
```

## Expected behavior

One would expect this to happen. And some code already implemented in pytorch/rl incorrectly assumes that this was how tensordict behave.

```markdown
In [47]: a[0]
Out[47]: NonTensorData(data=['cuac'], batch_size=torch.Size([]), device=None)
In [48]: a[0].data
Out[48]: array(['cuac'], dtype=object)
```

## Additional context

This error causes problems in pytorch/rl, exactly [here](https://github.com/pytorch/rl/blob/93fcb0222bd90e950692ca305d0f18930ceaeba0/torchrl/data/datasets/minari_data.py#L329) and [over here too](https://github.com/pytorch/rl/blob/93fcb0222bd90e950692ca305d0f18930ceaeba0/torchrl/data/datasets/minari_data.py#L345)

When dealing with these type of objects and its indexing, I would also expect torch.zeros_like(val[0]) to return a NonTensorData of the same type, but with 0s or null values instead of the original ones.

## Reason and Possible fixes

I have a proposed PR to solve this issue.

## Checklist

marcosgalleterobbva:
## Description

The reason for this PR is explained in depth [here](https://github.com/pytorch/tensordict/issues/1399)

## Motivation and Context

`close #1399`

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -3296,13 +3296,7 @@ class NonTensorDataBase(TensorClass):
         """Calling a NonTensorDataBase falls back to a call of its data."""
         return self.data(*args, **kwargs)
 
-    def __getitem__(self, idx):
-        if isinstance(self.data, list):
-            new_data = [self.data[i] for i in torch.as_tensor(idx).tolist()]
-        else:
-            new_data = self.data[idx]
-        new_batch_size = torch.Size(torch.Size(self.batch_size)[idx]) if self.batch_size else torch.Size([])
-        return NonTensorData(new_data, new_batch_size, self.device)
+
 
     def update(
         self,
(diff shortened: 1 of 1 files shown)


tensordict/tensorclass.py
```diff
@@ -3296,6 +3296,14 @@ def __call__(self, *args, **kwargs):
         """Calling a NonTensorDataBase falls back to a call of its data."""
         return self.data(*args, **kwargs)
 
+    def __getitem__(self, idx):
+        if isinstance(self.data, list):
+            new_data = [self.data[i] for i in torch.as_tensor(idx).tolist()]
```
vmoens, line 3301: Unfortunately, no, a list should not be treated differently than any other non-tensor data. 
For example, what happens if you set a list that has only one element in a tensordict of batch-size 3? 
Why list and not tuples? Or other indexable sequences?

`NonTensorStack` is the way to implement this behaviour.

You can have a non tensor data item shared among many different elements of a stack, this is why you can set a batch size to `NonTensorData` with no requirement on the data characteristics.

marcosgalleterobbva:
I answered to this problem in a bit more depth [in the original issue.](https://github.com/pytorch/tensordict/issues/1399#issuecomment-3131682808)

marcosgalleterobbva:
This PR was incorrectly opened to handle an issue that should have been resolved in a completely different way.

PR #1401, opened 2025-07-28
[BugFix] Args for TC with ClassVar

vmoens:
@alex-bene

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1190,7 +1190,20 @@ def _init_wrapper(
             lock = frozen
         if not is_compiling():
             # zip not supported by dynamo
-            for value, key in zip(args, type(self).__dataclass_fields__):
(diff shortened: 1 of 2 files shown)


alex-bene:
This seems to have broken something:

```python
import torch
from tensordict.tensorclass import TensorClass

class PoC_bug0(TensorClass):
    boring_variable: torch.Tensor | None = None

try:
    poc_bug0 = PoC_bug0()
except ValueError as e:
    assert str(e) == "zip() argument 2 is longer than argument 1"
try:
    poc_bug0 = PoC_bug0(boring_variable=torch.tensor([12, 12]))
except ValueError as e:
    assert str(e) == "zip() argument 2 is longer than argument 1"
```

vmoens:
yes we can actually have more expected items than args, that was silly.
I just edited that

alex-bene:
Okay, nice, I don't see any obvious issue now 👌

PR #1421, opened 2025-08-19
[BugFix,TypeHint] Fix type annotations in tensorclass stub file

vmoens:
cc @zhou13 -- can you review this?


zhou13:
It does not help to return `Self | TensorCollection | torch.Tensor`.  The type checker will still complain that `Self | TensorCollection | torch.Tensor` is not compatible with `YourTensorClass`, since it might return a `torch.Tensor`. Is there any reason not to return just `Self`?

vmoens:
> It does not help to return `Self | TensorCollection | torch.Tensor`. The type checker will still complain that `Self | TensorCollection | torch.Tensor` is not compatible with `YourTensorClass`, since it might return a `torch.Tensor`. Is there any reason not to return just `Self`?

I corrected the `Self | TensorCollection` but in some cases (when we can reduce the tensordict) we need the `torch.Tensor` (when a reduction is possible).

zhou13:
@vmoens Is it possible to keep `reduce` as a separate function, e.g., `max_reduce`? That will clean up the typing system. But I guess it is not backward compatible. Alternatively, maybe it is a good idea just to return `Any` for such cases to disable type checking on the return value to avoid assertion.

vmoens:
> @vmoens Is it possible to keep `reduce` as a separate function, e.g., `max_reduce`? That will clean up the typing system. But I guess it is not backward compatible. Alternatively, maybe it is a good idea just to return `Any` for such cases to disable type checking on the return value to avoid assertion.

Maybe...

Another option would be to use overload:
```
@overload
def mean(self, ...) -> Self: ...
@overload
def mean(self, ..., reduce: blah=bleh) -> Self | torch. Tensor: ...
def mean(self, ..., reduce: blah=bleh) -> Self | torch. Tensor:
    ...
```
Would that help?

zhou13:
@vmoens I did some experiments and it seems that overload did work this in case! I was not aware that `@overload` can be used on optional arguments before.

```python
from __future__ import annotations

from collections.abc import Sequence
from typing import overload


@overload
def double(input_: int) -> int: ...


@overload
def double(input_: int, reduce: bool) -> int | list[int]: ...


def double(input_, reduce: bool = False):
    ...

y = double(10)
y_int_orlist = double(10, False)
```

vmoens:
Then I'll do a PR with this and include it in the next major (hopefully coming this week)

PR #1425, opened 2025-08-26
[BugFix] consistent use of non_blocking in tensordict and torch.Tensor on non-cuda devices

ji-huazhong:
## Description

## Motivation and Context

## Types of changes

## Checklist


tensordict/base.py
```diff
@@ -14258,14 +14259,17 @@ def _has_mps(self):
...
+
+        if device_type == "cuda":
             # TODO: dynamo doesn't like torch.cuda.is_initialized
             if not is_compiling() and torch.cuda.is_initialized():
                 torch.cuda.synchronize()
-        elif self._has_mps:
-            mps = getattr(torch, "mps", None)
-            if mps is not None:
-                mps.synchronize()
+        else:
```
vmoens, line 14270: Double indent

vmoens (commented):
Thanks for this PR!
Can you further explain why this is needed and what problem is solved?
Thanks!

ji-huazhong:
> Thanks for this PR! Can you further explain why this is needed and what problem is solved? Thanks!

Hi @vmoens We have recently been using VERL for post-training our model on the Ascend NPUs. We observed that as the training progresses, the gradient norm (grad norm) consistently diverges (runs out of control). In contrast, all metrics converge well when training on NV GPUs.

Coincidentally, we found that adding synchronization (torch.npu.synchronize) at the connection points between different components in the RL workflow resolved this issue, and it no longer recurs.

In VERL, a single controller distributes data(organized in the form of TensorDict) to various workers (each associated with a device) via Ray for computation. After the computation is completed, `td.to("cpu")` is called to move the data from the device to the CPU (see [here](https://github.com/volcengine/verl/blob/9f0f8b0e7c227bedacd16eacaf662f00d2de4367/verl/workers/fsdp_workers.py#L738)). The data is then aggregated back to the single controller through Ray, which then proceeds to execute the next phase of computation jobs. It was precisely by adding synchronization after `td.to("cpu")` that the gradient norm divergence issue ceased to occur.

Through further investigation, we discovered that the `to` operation of TensorDict appears to be non-blocking. Meanwhile, it ensures that when the target device is specified as `to('cpu')`, the `_sync_all` method is invoked to guarantee that the data transfer from the device to the CPU is successfully completed before the data is used. see: https://github.com/pytorch/tensordict/blob/5e781517a1d638b55c3c250d990921fea9ba3c69/tensordict/base.py#L14130-L14136 

https://github.com/pytorch/tensordict/blob/5e781517a1d638b55c3c250d990921fea9ba3c69/tensordict/base.py#L14260-L14268

However, this guarantee only covers CUDA and MPS devices. PyTorch also supports third-party hardware such as Intel XPU and Ascend NPU, and this PR  aims to extend this guarantee to cover these third-party hardware platforms as well.

vmoens (approved):
LGTM I like it!

FightingZhen:
Great to see this PR merged! When will the next version containing this feature be released? @vmoens

vmoens:
Hoping to release this by end of week but I have a bunch of CI issues...

PR #1434, opened 2025-09-05
[Feature] Support for mixed `TensorDictBase` types in `torch.autograd.grad`

Xmaster6y:
## Description

Support:

```py
inputs = TensorDict(a=torch.randn(2, 3, requires_grad=True))
outputs = inputs + 1
grad_outputs = LazyStackedTensorDict(TensorDict(a=torch.ones(3)), TensorDict(a=torch.ones(3)), stack_dim=0)
grads = torch.autograd.grad(outputs, inputs, grad_outputs)
```

## Motivation and Context

- Mentionned in #1417 

## Types of changes

## Checklist

### DIFF
--- a/tensordict/_torch_func.py
+++ b/tensordict/_torch_func.py
@@ -929,15 +929,11 @@ def _grad(
         )
 
     if grad_outputs is not None:
-        tup_grad_outputs = tuple(
-            grad_outputs._values_list(True, True, is_leaf=_NESTED_TENSORS_AS_LISTS)
-        )
+        tup_grad_outputs = tuple(grad_outputs[k] for k in outputs.keys(True, True))
     else:
         tup_grad_outputs = None
 
-    tup_outputs = tuple(
-        outputs._values_list(True, True, is_leaf=_NESTED_TENSORS_AS_LISTS)
-    )
+    tup_outputs = tuple(outputs[k] for k in outputs.keys(True, True))
 
     keys, all_inputs = inputs._items_list(True, True, is_leaf=_NESTED_TENSORS_AS_LISTS)
 
--- a/test/test_tensordict.py
+++ b/test/test_tensordict.py
@@ -10608,6 +10608,13 @@ class TestLazyStackedTensorDict:
         assert td.batch_size == (2, 4)
         assert td.batch_size == td2.batch_size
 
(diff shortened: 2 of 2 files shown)


tensordict/_torch_func.py
```diff
@@ -929,15 +929,11 @@ def _grad(
         )
 
     if grad_outputs is not None:
-        tup_grad_outputs = tuple(
-            grad_outputs._values_list(True, True, is_leaf=_NESTED_TENSORS_AS_LISTS)
-        )
+        tup_grad_outputs = tuple(grad_outputs[k] for k in outputs.keys(True, True))
```
vmoens, line 932: When using a lazy stack, `grad_outputs._values_list(..., is_leaf=_NESTED_TENSORS_AS_LISTS)` will return the leaf tensors (not their stack).
It's important if they can't be stacked (ragged dimensions).
The new implementation will break this I think (try with a stack of tensordicts that have different tensor shapes to convince yourself!)

vmoens (commented):
Thanks a mil! Not sure the changes really fix the issue -- see my comment

Xmaster6y:
What about now? Not sure we can do better. Feel free to close the PR in case of a wontfix.

Xmaster6y:
@vmoens bump

vmoens (commented):
Yeah not sure that's exactly what I had in mind. 
I think we should avoid things like 
```
tup_outputs = tuple(outputs[k] for k in outputs.keys(True, True))
```
which are inefficient (we could just use `values(...)`)
But more in general, the idea I had was to check if the various inputs all shared the same type (possibly recursively because you can have a Lazy stack within a regular TD) and if not, specialize the ones that don't have the right type. 
It's a bit convoluted though:
- you would need to call `td0.apply(check, td1, td2, is_leaf=_is_tensor_collection)` where `check` returns `True` if all types match
- if check returns False, then you need to cast one of these guys
- then we should implement the logic to do that. I guess something like
  ```
  # td: a TD with a LazyStack within
  # plain_td: a TD without lazy stack
   new_td = td.copy().update(plain_td)
  ```


With that kind of thing we can keep the rest of the function exactly as it it

Xmaster6y:
Ok I see, thanks for the heads-up.

PR #1460, opened 2025-10-14
[Feature] Extended support to include NPU use cases in addition to existing CUDA-compatible scenarios

UserChen666:
Extended support to include NPU use cases in addition to existing CUDA-compatible scenarios.

## Description

Based on the existing UT on CUDA, this PR supports running UT on NPU.

## Types of changes

## Checklist

### DIFF
--- a/test/_utils_internal.py
+++ b/test/_utils_internal.py
@@ -32,8 +32,10 @@ def prod(sequence):
     else:
         return int(np.prod(sequence))
 
+
 def is_npu_available():
     return hasattr(torch, "npu") and torch.npu.is_available()
+
 
 def get_available_devices():
     devices = [torch.device("cpu")]
--- a/tensordict/_torch_func.py
+++ b/tensordict/_torch_func.py
@@ -534,7 +534,7 @@ def _stack(
                 if not maybe_dense_stack:
                     raise RuntimeError(
                         "stacking tensordicts requires them to have congruent batch sizes, "
-                        f"got td[{i+1}].batch_size={td.batch_size} and td[0].batch_size="
+                        f"got td[{i + 1}].batch_size={td.batch_size} and td[0].batch_size="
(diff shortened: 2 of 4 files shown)


UserChen666:
> Thanks for this PR! Do you have an idea of how we can test this in the CI? Is there a public runner in GHA that has NPUs? I don't think pytorch provides such runners, so this feature would be mostly untested which isn't great as it'll give false reassurance that we properly cover NPU tests.

As far as I know, there is truely no public npu env for CI test. I think I can upload some 'Testcase passed' screenshot for temprary reassurance. Out team is trying to solve this situation, and maybe we can build a CI env for public use in the near future.

test/_utils_internal.py
```diff
@@ -40,6 +42,13 @@ def get_available_devices():
             devices += [torch.device(f"cuda:{i}")]
             if i == 1:
                 break
+    if is_npu_available():
+        n_npu = torch.npu.device_count()
+        if n_npu > 0:
+            for i in range(n_npu):
+                devices += [torch.device(f"npu:{i}")]
+                if i == 1:
+                    break
```
vmoens: I'm open to this but I would put big warning as a comment and a real warning as well to tell users that this is an experimental and untested (as in: not part of the CI) feature. Can you do that?

test/_utils_internal.py
```diff
@@ -40,6 +43,16 @@ def get_available_devices():
             devices += [torch.device(f"cuda:{i}")]
             if i == 1:
                 break
+    if is_npu_available():
+        warnings.warn(
+            "torch_npu is an experimental and untested feature currently."
```
vmoens:
```suggestion
            "torch_npu is an experimental feature and not currently included in tensordict CI/CD."
```

UserChen666:
I've already seen your commit. Do you have more suggestions for this PR?

PR #1477, opened 2025-11-17
[BugFix] Change is_diff check to use is_diff.any()

### LINKED ISSUE #1476: [BUG] _td.py uses comparison of lists as truth value without calling any/all
## Describe the bug

https://github.com/pytorch/tensordict/blob/f7b5e059899f7bfb18aeb6b0f29ce68489121265/tensordict/_td.py#L2581-L2582

## To Reproduce

This is a logic error nothing required to reproduce just look at the code.

## Reason and Possible fixes

I assume using any() would recreate the intended behavior.

## Checklist

pancan21:
## Description

Previous implementation using list of boolean values resulted in ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

I assume using any() instead of all() is the intended behavior.

## Motivation and Context
Why is this change required? ValueError when setting td.set_at_()
close #1476

## Types of changes

## Checklist

Not sure how this logical error should be tested. Please help

### DIFF
--- a/tensordict/_td.py
+++ b/tensordict/_td.py
@@ -2579,7 +2579,7 @@ class TensorDict(TensorDictBase):
                     ignore_lock=True,
                 )
             is_diff = dest[idx].tolist() != value.tolist()
-            if is_diff:
+            if is_diff.any():
                 dest_val = dest.maybe_to_stack()
                 dest_val[idx] = value
                 if dest_val is not dest:


tensordict/_td.py
```diff
@@ -2579,7 +2579,7 @@ def _set_at_str(self, key, value, idx, *, validated, non_blocking: bool):
                     ignore_lock=True,
                 )
             is_diff = dest[idx].tolist() != value.tolist()
-            if is_diff:
+            if is_diff.any():
```
vmoens: This should be a list. If it's not, we should fix that!

PR #1516, opened 2026-01-06
[Feature] More copy() refactors for compile friendliness

vmoens:


vmoens:
unrelated failing tests

PR #1611, opened 2026-02-26
[Feature] Add UCXX transport for fast node-to-node TensorDict transfers

vmoens:
## Summary

- Adds a UCXX-based transport layer for high-performance point-to-point TensorDict communication between nodes. UCXX auto-selects the fastest available transport (InfiniBand, shared memory, TCP, CUDA IPC/GPUDirect RDMA).
- Introduces `TensorDictPipe` with a two-phase protocol: first send transmits metadata + consolidated buffer; subsequent same-schema sends transmit only the raw buffer (zero-alloc steady-state on the receiver).
- Introduces `TensorDictServer` for one-to-many listener setups, plus standalone `send_tensordict`/`recv_tensordict` functions.
- Overloads `td.send(pipe)` / `td.recv(pipe)` so TensorDict remains the entry point, and adds `td.asend(pipe)` / `td.arecv(pipe)` async variants.
- CUDA-direct transfers are supported when tensors are on GPU.
- `ucxx` is a fully optional dependency (lazy import, clear error message if missing).

## API

```python
from tensordict._ucxx import TensorDictPipe

# Sender
pipe = await TensorDictPipe.connect("node-b", 13337)
await td.asend(pipe)        # first: handshake + data
await td.asend(pipe)        # subsequent: raw buffer only (zero-alloc)

# Receiver
pipe = await TensorDictPipe.listen(13337)
td = await TensorDict.recv(pipe)   # first: builds TD from metadata
await td.arecv(pipe)               # subsequent: in-place overwrite

# Sync variants also work
td.send(pipe)
td.recv(pipe)
```

## Test plan

- [x] Unit tests for metadata hashing (deterministic, collision-free, key-order-invariant)
- [x] Unit tests for total bytes computation from metadata
- [x] Protocol flag tests (new schema vs same schema detection)
- [x] Full roundtrip tests with fake endpoints (consolidated, nested TDs, steady-state in-place)
- [x] `td.send(pipe)` / `td.recv(pipe)` overload dispatch tests
- [ ] Integration tests with real UCXX endpoints (skipped without ucxx installed)
- [ ] CUDA-direct transfer tests (skipped without CUDA)
- [ ] `TensorDictServer` multi-client test (skipped without ucxx installed)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/scripts/check_rst_titles.py
+++ b/scripts/check_rst_titles.py
@@ -10,6 +10,7 @@ RST_SECTION_CHARS = frozenset("!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~")
 
 logger = getLogger(__name__)
 
+
 def _strip_bom(line: str) -> str:
     """Strip the Unicode BOM (U+FEFF) that some generators emit."""
     return line.lstrip("\ufeff")
--- a/tensordict/_ucxx.py
+++ b/tensordict/_ucxx.py
@@ -125,7 +125,6 @@ async def send_tensordict(
         consolidated (bool): if ``True``, consolidates into a single buffer
             before sending. Defaults to ``True``.
     """
-
     if consolidated:
         if td.is_consolidated():
             td_c = td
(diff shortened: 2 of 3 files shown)


vmoens:
@tensordictbot rebase

vmoens:
@tensordictbot lint

PR #1701, opened 2026-05-14
[BugFix] Attempting to make tensordict device transfer marking thread safe.

### LINKED ISSUE #1700: [BUG] Tensordict is not Threadsafe for device transfer in all cases
## Describe the bug

Sorry I'm not a tensordict expert.  I'm gonna try to summarize what I think is happening and why, and I've got a PR I'm preparing that I believe fixes this issue if you think it's appropriate.

When creating a tensordict in a multi-threaded python environment, and especially when the tensordicts are somewhat large, the `_RecordDeviceTransfer` class contains a `mark` step.  The mark operation is not thread safe: it's a module-global singleton so it can be accessed by every thread, and concurrent threads doing independent work (and hopefully using independent streams for device transfer) can have collisions with each other.  This actually doesn't need cuda to reproduce: You can just create tensordicts with `device="cpu"` and not specify "non_blocking" I believe and it will call mark.

## To Reproduce

Steps to reproduce the behavior.

It's a race condition, so you can only see it in select conditions, but it isn't too hard to reproduce.  The error looks like this:

```
  File "/lustre/fs1/portfolios/coreai/projects/coreai_modulus_cae/users/coreya/workdir/repro_tensordict_thread_safety.py", line 76, in _worker
    TensorDict(source, device="cpu")
  File "/lustre/fsw/portfolios/coreai/users/coreya/venvs/geotransolver25.06/lib/python3.12/site-packages/tensordict/_td.py", line 294, in __init__
    _device_recorder.mark()
  File "/lustre/fsw/portfolios/coreai/users/coreya/venvs/geotransolver25.06/lib/python3.12/site-packages/tensordict/base.py", line 257, in mark
    raise RuntimeError("Can only mark one TensorDict at a time.")
RuntimeError: Can only mark one TensorDict at a time.
```

A full reproducer is here:

```python

#!/usr/bin/env python3
"""Standalone reproducer for tensordict's non-thread-safe ``_device_recorder``.

Background
----------
``tensordict.base._device_recorder`` is a *process-global* singleton.
``TensorDict.__init__`` calls ``_device_recorder.mark()`` whenever it is
constructed with a non-CUDA ``device=`` argument and no explicit
``non_blocking=``.  ``mark()`` is guarded by a plain boolean and raises
``RuntimeError("Can only mark one TensorDict at a time.")`` if it sees
``self.marked == True``.

Because the recorder is module-global (not ``threading.local``), two
threads constructing CPU-device TensorDicts at the same time race on the
guard and the loser crashes.

This script demonstrates the race deterministically by spawning N
threads that each loop ``TensorDict({...}, device="cpu")`` behind a
barrier so they fire near-simultaneously.

Usage
-----
    python repro_tensordict_thread_safety.py
    python repro_tensordict_thread_safety.py --threads 16 --iters 5000

Exit code is 0 if the race fires (bug reproduced), 1 if no error was
seen.  If you don't see the race, widen the window with larger
``--threads`` / ``--iters`` / ``--keys`` / ``--size``.

Dependencies: ``torch`` and ``tensordict`` only.
"""

from __future__ import annotations

import argparse
import sys
import threading
import traceback
from typing import Optional


def _print_env() -> None:
    import torch  # noqa: PLC0415

    import tensordict  # noqa: PLC0415

    print(f"python       = {sys.version.split()[0]}")
    print(f"torch        = {torch.__version__}")
    print(f"tensordict   = {tensordict.__version__}")


def _build_source(n_keys: int, size: int):
    """Build a dict big enough that TensorDict.__init__ holds the mark
    long enough for another thread to race on it."""
    import torch  # noqa: PLC0415

    return {f"k{i}": torch.zeros(size, dtype=torch.float32) for i in range(n_keys)}


def _worker(
    barrier: threading.Barrier,
    iters: int,
    source: dict,
    errors: list[BaseException],
    stop: threading.Event,
) -> None:
    from tensordict import TensorDict  # noqa: PLC0415

    barrier.wait()
    for _ in range(iters):
        if stop.is_set():
            return
        try:
            # device="cpu" + no non_blocking is the path that triggers
            # _device_recorder.mark() in tensordict/_td.py.
            TensorDict(source, device="cpu")
        except BaseException as e:  # noqa: BLE001 - we want to record any failure
            errors.append(e)
            stop.set()
            return


def reproduce(
    threads: int,
    iters: int,
    n_keys: int,
    size: int,
) -> Optional[BaseException]:
    """Run *threads* worker threads, each constructing *iters* TensorDicts.

    Returns the first exception observed by any worker, or ``None`` if
    all workers completed without raising.
    """
    source = _build_source(n_keys=n_keys, size=size)

    barrier = threading.Barrier(threads)
    errors: list[BaseException] = []
    stop = threading.Event()

    workers = [
        threading.Thread(
            target=_worker,
            args=(barrier, iters, source, errors, stop),
            name=f"repro-{i}",
        )
        for i in range(threads)
    ]
    for t in workers:
        t.start()
    for t in workers:
        t.join()

    return errors[0] if errors else None


def main() -> int:
    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
    parser.add_argument(
        "--threads",
        type=int,
        default=8,
        help="Number of worker threads (default: 8).",
    )
    parser.add_argument(
        "--iters",
        type=int,
        default=2000,
        help="TensorDict constructions per worker (default: 2000).",
    )
    parser.add_argument(
        "--keys",
        type=int,
        default=64,
        help="Tensors per TensorDict (default: 64). Larger widens the race window.",
    )
    parser.add_argument(
        "--size",
        type=int,
        default=64,
        help="Elements per tensor (default: 64). Larger widens the race window.",
    )
    args = parser.parse_args()

    _print_env()
    print(f"threads      = {args.threads}")
    print(f"iters/thread = {args.iters}")
    print(f"keys / td    = {args.keys}")
    print(f"size / tensor= {args.size}")
    print("---")

    err = reproduce(args.threads, args.iters, args.keys, args.size)

    if err is None:
        print("No race observed.")
        print(
            "Try larger --threads / --iters / --keys / --size to widen the window.",
            file=sys.stderr,
        )
        return 1

    print(f"Race reproduced: {type(err).__name__}: {err}")
    print("---")
    traceback.print_exception(type(err), err, err.__traceback__)
    return 0


if __name__ == "__main__":
    raise SystemExit(main())

```

DISCLAIMER:
I am a real person (I Promise!) but I used Claude to help condense the error and behavior into a standalone reproducer.  I ran it myself several times to validate.

## Expected behavior

Ideal behavior: tensordict should be threadsafe.  Marking multiple tensordicts concurrently should be safe, provided no more than one is marked _per thread_.  I believe that is the natural extension of the design, but again it's not my repository :).

## Screenshots

N/A

## System info

`tensordict` installed via uv

```python
```

## Reason and Possible fixes

I have a fix ready and will open a PR to propose the fix; I obviously don't know the full scope of the repo, sorry.  I will not be offended if you don't merge my PR, I just consider it a suggestion of "this might be a viable path to address, would you consider it?"

## Checklist

- [X] I have checked that there is no similar issue in the repo (**required**)
- [X] I have read the [documentation](https://github.com/pytorch/tensordict/tree/main/docs/) (**required**)
- [X] I have provided a minimal working example to reproduce the bug (**required**)

coreyjadams:
## Description

This PR would close #1700 , ideally.  But, I am not the maintainer.  I do promise, in this day of AI bots all over github, that I'm at least a real person ...

## Motivation and Context

The thread-safety issue I raised in #1700 is actually a challenging performance issue in our datapipes in NVIDIA-physicsnemo.  Using a hotfix (not this branch, actually, I did it way messier to verify and monkey-patched tensordict ... no judgement please 😅 )I was able to increase our data loading concurrency dramatically from 1 -> 8x or more.  Since that's our main bottleneck, it immediately gives a corresponding speedup (up to network bandwidth limits).  So, the motivation here is really "I want threads for IO performance, please".

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

I don't think this needs updates to the docs ... not sure how much you want to advertise thread safety especially if the whole product isn't deliberately thread-safe?  Maybe it already is, actually - please, I'm speaking mostly from ignorance!

I also have not yet looked at updating tests, since I do not know how much python threading tests you want to include?

At the end of the day, if you prefer to ignore this PR and go another route, I will not be insulted!  I am proposing this as one option that might work to fix #1700, but I leave it up to you to fix or not?


vmoens (approved):
Thanks for the fix! LGTM
would approve even if it were a bot! 

I added a test inspired by your MRE

Can you sign the CLA? I can't merge unless that flag is green

Will do a minor with it quickly, 0.13 also coming soon

coreyjadams:
Awesome!  Yes to the CLA - I have to go through the proper channels for NVIDIA <---> Meta CLA, which may take a business day or two.  I'll update here when I'm added, I've kicked off the process on the NVIDIA side.  We've got a standing CLA I think I just need to have myself added through the right people.

coreyjadams:
I _think_ the CLA has gone through on the NVIDIA side.  Whenever you'd like, would you try again? Thanks!

vmoens:
image
It doesn't really appear so right now

vmoens:
Superseeded by #1711 
thanks @coreyjadams!

PR #1704, opened 2026-05-16
[BugFix] Fix dynamic-shape export (#1003): handle scalar SymInt in _parse_batch_size and store batch_dims in pytree context

### LINKED ISSUE #1003: [BUG] `TensorDict` with dynamic, input-dependent  `batch_size` is not `torch.export.export`able
## To Reproduce

```python
import torch
import tensordict

class Test(torch.nn.Module):
    def forward(self, x: torch.Tensor, y: torch.Tensor):
        return tensordict.TensorDict({
            'x': x,
            'y': y,
            },
            batch_size=x.shape[0] # comment this line out and it works, but batch_size = [] and not x.shape[0]
         )
    
test = Test()
result = torch.export.export(
    test, 
    args=(torch.zeros(2,100), torch.zeros(2,100)),
    strict=False,
    dynamic_shapes={
        'x': {0: torch.export.Dim('batch'), 1: torch.export.Dim('time')},
        'y': {0: torch.export.Dim('batch'), 1: torch.export.Dim('time')}
        }
    )
print(result.module()(torch.zeros(5,100), torch.zeros(5,100)))
```

```bash
(myenv) egaznep@...@volta: $ python scripts/mwe.py                                                                                                                                  
  File "./myenv/lib/python3.10/site-packages/tensordict/_td.py", line 2082, in _parse_batch_size                                                                                               
    return torch.Size(batch_size)                                                                                                                                                                                               
TypeError: 'SymInt' object is not iterable

During handling of the above exception, another exception occurred:

Traceback (without unrelated parts):
  File "./scripts/mwe.py", line 6, in forward
    return tensordict.TensorDict({
  File "./myenv/lib/python3.10/site-packages/tensordict/_td.py", line 285, in __init__
    self._batch_size = self._parse_batch_size(source, batch_size)
  File "./myenv/lib/python3.10/site-packages/tensordict/_td.py", line 2090, in _parse_batch_size
    raise ValueError(ERR)
ValueError: batch size was not specified when creating the TensorDict instance and it could not be retrieved from source.
```

## Expected behavior

I would expect the module to be successfully converted into an `ExportedProgram`.

## System info

Describe the characteristic of your environment:
 * Describe how the library was installed (pip, source, ...)
 * Python version
 * Versions of any other relevant libraries

```
2024.09.19 2.0.2 3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:45:18) [GCC 12.3.0] linux 2.6.0.dev20240919
```

## Reason and Possible fixes

If you know or suspect the reason for this bug, paste the code lines and suggest modifications.

## Checklist

- [ ] I have checked that there is no similar issue in the repo (**required**)
- [ ] I have read the [documentation](https://github.com/pytorch/tensordict/tree/main/docs/) (**required**)
- [ ] I have provided a minimal working example to reproduce the bug (**required**)

felixmaximilian:
Fixes #1003.

## Problem

`torch.export.export()` with `dynamic_shapes` failed for any module that constructs a `TensorDict` with an input-dependent `batch_size`. Two independent bugs caused this:

**Bug 1 — `_parse_batch_size` rejects scalar `SymInt`** (`_td.py`)

When `batch_size=x.shape[0]` is passed during compilation, `batch_size` is a `torch.SymInt`. The `is_compiling()` path checks `isinstance(batch_size, Number)`, but `torch.SymInt` does not subclass `numbers.Number`, so the value falls through to `raise ValueError()`:

```
ValueError: batch size was not specified when creating the TensorDict instance
```

**Bug 2 — `_tensordict_flatten` stores `batch_size` (may contain `SymInt`) in the pytree context** (`_pytree.py`)

When the exported module returns a `TensorDict`, `torch.export` serializes the output pytree spec. The context stored `batch_size: torch.Size` which may contain `SymInt` values. `torch.export` calls `as_python_constant()` on every context item — `SymInt` is not a Python constant, raising:

```
AsPythonConstantNotImplementedError: SymNodeVariable() is not a constant
```

## Fix

**`tensordict/_td.py`**: extend the `Number` check in `_parse_batch_size` to also accept `torch.SymInt`:

```python
elif isinstance(batch_size, (Number, torch.SymInt)):
    return torch.Size([batch_size])
```

**`tensordict/_pytree.py`**: store `batch_dims: int` (the count, always a Python constant) in the pytree context instead of `batch_size`. In `_tensordict_unflatten`, reconstruct `batch_size` from the leading dimensions of the actual tensor shapes. The existing jacrev/jacfwd basis-vector mismatch logic is preserved via a backward-compatible `"batch_size"` legacy path.

## Tests

Two new tests added to `TestExport` in `test/test_compile.py`:

- `test_export_dynamic_batch_size_scalar`: covers `batch_size=x.shape[0]` (scalar `SymInt`, `strict=False`) — the original issue repro
- `test_export_dynamic_batch_size_multi_dim`: covers `batch_size=[b, t]` (multi-dim `SymInt`, `strict=True`) — the pytree serialization fix

Both tests fail on unpatched code and pass after the fix. The full `TestExport` suite (5 tests) passes in a clean venv.

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/pytorch/tensordict/blob/main/CONTRIBUTING.md)
- [x] Tests added for both bug fixes
- [x] Black + flake8 clean
- [x] Links to issue #1003

### DIFF
--- a/tensordict/_pytree.py
+++ b/tensordict/_pytree.py
@@ -93,7 +93,11 @@ def _tensordict_flatten(d: TensorDict) -> Tuple[List[Any], Context]:
         values = []
     return values, {
         "keys": keys,
-        "batch_size": d.batch_size,
+        # Store batch_dims (int) rather than batch_size (torch.Size that may
+        # contain SymInts during torch.export with dynamic shapes). The unflatten
+        # reconstructs the actual batch_size from tensor shapes at that point.
+        # See: https://github.com/pytorch/tensordict/issues/1003
+        "batch_dims": len(d.batch_size),
         "names": d.names if d._has_names() else None,
         "device": d.device,
         "constructor": _constructor(type(d)),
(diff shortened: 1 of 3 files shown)


felixmaximilian:
Hi, the CLA is signed. Could a maintainer please approve the CI run? Happy to address any review feedback.

vmoens (approved):
LGTM thanks!

PR #1709, opened 2026-05-20
[BugFix] Run __post_init__ and apply field defaults under torch.compile (#1708, #1710)

### LINKED ISSUE #1708: [BUG] `@tensorclass` silently skips `__post_init__` under `torch.compile`, producing wrong output
## Summary

Since v0.12.0, `__post_init__` is not called when a `@tensorclass` is constructed inside a `torch.compile`-traced function. There is no error: the tensorclass is returned with fields exactly as they were passed to the constructor, and any normalization, derivation, or validation the user wrote in `__post_init__` is silently dropped. Compiled output diverges from eager output.

## Minimal reproducer

```python
import torch
from tensordict import tensorclass


@tensorclass
class Foo:
    x: torch.Tensor

    def __post_init__(self):
        self.x = self.x * 2


print("Eager   :", Foo(x=torch.ones(3)).x)                              # tensor([2., 2., 2.])
print("Compiled:", torch.compile(lambda x: Foo(x=x).x)(torch.ones(3)))  # tensor([1., 1., 1.])
```

Output:

```
Eager   : tensor([2., 2., 2.])
Compiled: tensor([1., 1., 1.])
```

## Root cause

PR [#1552](https://github.com/pytorch/tensordict/pull/1552) replaced the unconditional `__init__(self, **kwargs)` call in `_init_wrapper` with a fast-path bypass branch. The `__post_init__` invocation at the end of that branch is guarded by `not is_compiling()`:

```python
# tensordict/tensorclass.py @ v0.12.3, line 1412
if not is_compiling() and hasattr(type(self), "__post_init__"):
    self.__post_init__()
```

Under `torch.compile`, the guard short-circuits and `__post_init__` is never called. The symmetric skip exists on the `_from_tensordict` path at [line 1664](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/tensorclass.py#L1664).

Pre-#1552 (v0.11.0 and earlier), the wrapper unconditionally called the user `__init__(self, **kwargs)` ([v0.11.0 line 1295](https://github.com/pytorch/tensordict/blob/v0.11.0/tensordict/tensorclass.py#L1295)), which in turn called `__post_init__` as part of dataclass semantics. The skip is purely a side effect of bypassing the user `__init__`.

## Version timeline

| tensordict | Compiled output of repro | Notes |
|------------|--------------------------|-------|
| 0.10.0 | `tensor([2., 2., 2.])` (correct) | unconditional `__init__(self, **kwargs)` |
| 0.11.0 | `tensor([2., 2., 2.])` (correct) | same |
| 0.12.0 | `tensor([1., 1., 1.])` (**wrong**) | PR [#1552](https://github.com/pytorch/tensordict/pull/1552) |
| 0.12.1, 0.12.2, 0.12.3 | wrong | code path unchanged |
| `main` @ HEAD | wrong | same guard at [tensorclass.py:1418](https://github.com/pytorch/tensordict/blob/main/tensordict/tensorclass.py) |

## Expected behavior

`__post_init__` should be called under `torch.compile`, matching eager semantics. The dataclass contract is that `__post_init__` always runs as part of construction. Users cannot detect or guard against it being skipped at the call site, because the construction itself doesn't raise — the tensorclass just has the wrong state.

The performance motivation of PR #1552 (avoiding dataclass `__init__` / `__setattr__` dispatch) is unaffected by calling `__post_init__`: it runs *after* the fast direct-dict assignment and adds at most one Python function call per construction.

## Suggested fix

Drop the `not is_compiling()` half of both guards:

```diff
- if not is_compiling() and hasattr(type(self), "__post_init__"):
+ if hasattr(type(self), "__post_init__"):
      self.__post_init__()
```

at [`tensorclass.py:1412`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/tensorclass.py#L1412) (bypass branch in `_init_wrapper`) and the symmetric guard at [`tensorclass.py:1664`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/tensorclass.py#L1664) (`_from_tensordict`).

If `__post_init__` was intentionally guarded because Dynamo has trouble tracing it, the alternative is to take the `_has_custom_setattr=True` slow path (call user `__init__` -> calls `__post_init__`) whenever `is_compiling()` is true. The slow path was the only path pre-#1552 and worked correctly under compile.

## System info

```
Python      : 3.12.12
torch       : 2.11.0+cu130
tensordict  : 0.12.3
OS          : Linux x86_64
```

## Checklist

- [x] I have checked that there is no similar issue in the repo (searched `post_init`, `__post_init__`, `tensor_only compile`; the only `__post_init__` issue is [#172](https://github.com/pytorch/tensordict/issues/172), the original feature request from 2023)
- [x] I have read the [documentation](https://github.com/pytorch/tensordict/tree/main/docs/)
- [x] I have provided a minimal working example to reproduce the bug

### LINKED ISSUE #1710: [BUG] `@tensorclass` field defaults are not applied under `torch.compile`, raising `KeyError`
## Summary

Since v0.12.0, when a `@tensorclass` is constructed inside a `torch.compile`-traced function and the caller omits a field that has a default value, the field is left genuinely absent from the tensorclass's underlying `_tensordict`. The next access raises `KeyError`. Eager construction applies the default correctly.

This is somewhat related to [#1708](https://github.com/pytorch/tensordict/issues/1708), though in practice they're actually distinct: both bugs originate from the bypass branch introduced by PR [#1552](https://github.com/pytorch/tensordict/pull/1552), but they fire independently (each reproducer here uses no `__post_init__`; the #1708 reproducer uses no field defaults) and target different guards in the same wrapper. Fixing one will not necessarily automatically fix the other, depending on how the fix is done.

## Minimal reproducer

```python
import torch
from tensordict import tensorclass


@tensorclass
class Foo:
    cache: torch.Tensor = torch.zeros(3)


print("Eager   :", Foo().cache)                              # tensor([0., 0., 0.])
print("Compiled:", torch.compile(lambda: Foo().cache)())     # KeyError: 'key "cache" not found...'
```

Output:

```
Eager   : tensor([0., 0., 0.])
Compiled: KeyError: 'key "cache" not found in TensorDict with keys []'
```

Passing `cache=...` explicitly works in both modes; only the *defaulting* path is broken under compile.

## Root cause

PR [#1552](https://github.com/pytorch/tensordict/pull/1552) replaced the unconditional `__init__(self, **kwargs)` call in `_init_wrapper` with a fast-path bypass branch that writes only the kwargs the caller passed into `_tensordict` / `_non_tensordict`. The default-handling loops that fill in missing kwargs from the dataclass field defaults are gated by `not is_compiling()`:

```python
# tensordict/tensorclass.py @ v0.12.3

# Skips dataclass default / default_factory under compile
if not is_compiling():                                       # line 1329
    for key, field in type(self).__dataclass_fields__.items():
        ...
        if default not in (None, dataclasses.MISSING):
            kwargs.setdefault(key, default)

# Skips filling missing keys with None under compile
if not is_compiling():                                       # line 1390
    for key in self.__expected_keys__:
        kwargs.setdefault(key, None)

# Both branches below iterate kwargs.items() only - they write nothing for
# keys the caller didn't pass, so omitted-but-defaulted fields end up missing.
if tensor_only:
    for key, value in kwargs.items():
        if value is None:
            _non_td[key] = None
        else:
            _td_dict[key] = _validate(value, ...)
else:
    _set = type(self).set
    for key, value in kwargs.items():
        _set(self, key, value)
```

Under `torch.compile`, both `setdefault` loops are skipped. Fields the caller omitted are written to neither `_tensordict` nor `_non_tensordict`. Subsequent access goes through [`_getattr`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/tensorclass.py#L1867) → [`_get_str`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/_td.py#L2825) → [`_default_get`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/base.py#L7977), which raises `KeyError`.

Pre-#1552 (v0.11.0 and earlier), the wrapper unconditionally called the dataclass-generated `__init__(self, **kwargs)`, which applied defaults via the dataclass machinery.

## Version timeline

| tensordict | Compiled output of repro | Notes |
|------------|--------------------------|-------|
| 0.10.0 | `tensor([0., 0., 0.])` (correct) | unconditional `__init__(self, **kwargs)` |
| 0.11.0 | `tensor([0., 0., 0.])` (correct) | same |
| 0.12.0 | `KeyError: 'cache' ...` | PR [#1552](https://github.com/pytorch/tensordict/pull/1552) |
| 0.12.1, 0.12.2, 0.12.3 | `KeyError: 'cache' ...` | code path unchanged |
| `main` @ HEAD | `KeyError: 'cache' ...` | same guards in [tensorclass.py](https://github.com/pytorch/tensordict/blob/main/tensordict/tensorclass.py) |

## Expected behavior

`@tensorclass` construction under `torch.compile` should apply field defaults the same way eager does. A tensorclass with a defaulted field should be constructible with that argument omitted, in both eager and compiled contexts.

## Suggested fix

Drop the `not is_compiling()` half of the guards in `_init_wrapper`:

```diff
- if not is_compiling():
-     for key, field in type(self).__dataclass_fields__.items():
-         ...
-         if default not in (None, dataclasses.MISSING):
-             kwargs.setdefault(key, default)
+ for key, field in type(self).__dataclass_fields__.items():
+     ...
+     if default not in (None, dataclasses.MISSING):
+         kwargs.setdefault(key, default)
  ...
- if not is_compiling():
-     for key in self.__expected_keys__:
-         kwargs.setdefault(key, None)
+ for key in self.__expected_keys__:
+     kwargs.setdefault(key, None)
```

at [`tensorclass.py:1329`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/tensorclass.py#L1329) and [`tensorclass.py:1390`](https://github.com/pytorch/tensordict/blob/v0.12.3/tensordict/tensorclass.py#L1390).

If `kwargs.setdefault` loops have known Dynamo-tracing issues, the alternative is to take the `_has_custom_setattr=True` slow path (call the dataclass-generated `__init__(self, **kwargs)`) whenever `is_compiling()` is true. The slow path was the only path pre-#1552 and worked correctly under compile.

## System info

```
Python      : 3.12.12
torch       : 2.11.0+cu130
tensordict  : 0.12.3
OS          : Linux x86_64
```

## Checklist

- [x] I have checked that there is no similar issue in the repo
- [x] I have read the [documentation](https://github.com/pytorch/tensordict/tree/main/docs/)
- [x] I have provided a minimal working example to reproduce the bug

vmoens:
## Summary

Fixes #1708 and #1710. Both bugs originate from PR #1552 (the `__init__` fast-path bypass) — distinct guards in the same wrapper, gated by `not is_compiling()`, that silently skip behavior the dataclass `__init__` would otherwise have provided.

### #1708 — `__post_init__` silently skipped under compile

```python
@tensorclass
class Foo:
    x: torch.Tensor
    def __post_init__(self):
        self.x = self.x * 2

Foo(x=torch.ones(3)).x                              # tensor([2., 2., 2.])  (correct)
torch.compile(lambda x: Foo(x=x).x)(torch.ones(3))  # tensor([1., 1., 1.])  (wrong, pre-fix)
```

### #1710 — field defaults silently skipped under compile

```python
@tensorclass
class Foo:
    cache: torch.Tensor = torch.zeros(3)

Foo().cache                              # tensor([0., 0., 0.])  (correct)
torch.compile(lambda: Foo().cache)()     # KeyError: 'key "cache" not found...'  (pre-fix)
```

## Root cause

PR #1552 replaced the unconditional dataclass `__init__(self, **kwargs)` with a fast-path bypass that writes only the kwargs the caller passed. Four pieces of dataclass-construction behavior were re-added to the fast path but guarded by `not is_compiling()`:

1. `__post_init__` invocation in the bypass branch of `_init_wrapper`
2. `__post_init__` invocation in `_from_tensordict`
3. Dataclass field default / `default_factory` fill (`kwargs.setdefault(key, default)`)
4. None-fill for all expected keys

Under `torch.compile`, all four were short-circuited. (1) and (2) silently produced wrong output; (3) and (4) caused omitted-but-defaulted fields to be missing from `_tensordict`, raising `KeyError` on access.

## Fix

Drop the `not is_compiling()` half of all four guards. Also remove a dead `else: pass` branch (with stale `# TODO: Decide what to do here`) that the second guard removal made redundant. The perf motivation of #1552 is preserved — these loops only matter at construction time and only do work proportional to the number of fields.

## Test plan

Two new regression test classes in `test/test_compile.py`:

- **`TestTCPostInitCompile`** (#1708):
  - `__post_init__` runs on direct construction under compile
  - `__post_init__` runs via `_from_tensordict` under compile
- **`TestTCDefaultsCompile`** (#1710):
  - Concrete default applied under compile
  - `dataclasses.field(default_factory=...)` applied under compile
  - Omitted field with `= None` default doesn't raise under compile

Results:

- [x] 5 new regression tests pass
- [x] Full `test/test_tensorclass.py` — 144 passed, 1 skipped
- [x] Tensorclass-related `test/test_compile.py` — 68 passed, 9 skipped (CUDA), 2 xfailed (pre-existing)

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1349,21 +1349,22 @@ def _init_wrapper(
                     "dynamo doesn't support arguments when building a tensorclass, pass the keyword explicitly."
                 )
 
+        # Use `is`/isinstance instead of `in (..., dataclasses.MISSING)`:
+        # under torch.compile, Dynamo can't proxy `_MISSING_TYPE` for `==`
+        # comparisons against a tensor default value.
+        _missing_type = getattr(
+            dataclasses, "_MISSING_TYPE", type(dataclasses.MISSING)
+        )
         for key, field in type(self).__dataclass_fields__.items():
             # Only process fields that are in __expected_keys__ (excludes ClassVar fields)
             if key in self.__expected_keys__:
-                if field.default_factory not in (
-                    dataclasses.MISSING,
-                ) and not isinstance(
-                    field.default_factory,
-                    getattr(
-                        dataclasses, "_MISSING_TYPE", type(dataclasses.MISSING)
-                    ),
+                if field.default_factory is not dataclasses.MISSING and not isinstance(
+                    field.default_factory, _missing_type
                 ):
                     default = field.default_factory()
                 else:
                     default = field.default
-                if default not in (None, dataclasses.MISSING):
+                if default is not None and not isinstance(default, _missing_type):
                     kwargs.setdefault(key, default)
 
         missing_params = [p for p in required_params if p not in kwargs]
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1352,9 +1352,7 @@ def _init_wrapper(
         # Use `is`/isinstance instead of `in (..., dataclasses.MISSING)`:
         # under torch.compile, Dynamo can't proxy `_MISSING_TYPE` for `==`
         # comparisons against a tensor default value.
-        _missing_type = getattr(
-            dataclasses, "_MISSING_TYPE", type(dataclasses.MISSING)
-        )
+        _missing_type = getattr(dataclasses, "_MISSING_TYPE", type(dataclasses.MISSING))
         for key, field in type(self).__dataclass_fields__.items():
             # Only process fields that are in __expected_keys__ (excludes ClassVar fields)
             if key in self.__expected_keys__:
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1432,7 +1432,7 @@ def _init_wrapper(
                 _set = type(self).set
                 for key, value in kwargs.items():
                     _set(self, key, value)
-            if not is_compiling() and hasattr(type(self), "__post_init__"):
+            if hasattr(type(self), "__post_init__"):
                 self.__post_init__()
         if lock:
             self._tensordict.lock_()
@@ -1688,7 +1688,7 @@ def _from_tensordict(
         object.__setattr__(tc, "_non_tensordict", non_tensordict)
     # since we aren't calling the dataclass __init__, we need to manually
     # check whether a __post_init__ method has been defined and invoke it
-    if not is_compiling() and hasattr(cls, "__post_init__"):
+    if hasattr(cls, "__post_init__"):
         tc.__post_init__()
     return tc
 
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1349,26 +1349,22 @@ def _init_wrapper(
                     "dynamo doesn't support arguments when building a tensorclass, pass the keyword explicitly."
                 )
 
-        if not is_compiling():
-            for key, field in type(self).__dataclass_fields__.items():
-                # Only process fields that are in __expected_keys__ (excludes ClassVar fields)
-                if key in self.__expected_keys__:
-                    if field.default_factory not in (
-                        dataclasses.MISSING,
-                    ) and not isinstance(
-                        field.default_factory,
-                        getattr(
-                            dataclasses, "_MISSING_TYPE", type(dataclasses.MISSING)
-                        ),
-                    ):
-                        default = field.default_factory()
-                    else:
-                        default = field.default
-                    if default not in (None, dataclasses.MISSING):
-                        kwargs.setdefault(key, default)
-        else:
-            # TODO: Decide what to do here
-            pass
+        for key, field in type(self).__dataclass_fields__.items():
+            # Only process fields that are in __expected_keys__ (excludes ClassVar fields)
+            if key in self.__expected_keys__:
+                if field.default_factory not in (
+                    dataclasses.MISSING,
+                ) and not isinstance(
+                    field.default_factory,
+                    getattr(
+                        dataclasses, "_MISSING_TYPE", type(dataclasses.MISSING)
+                    ),
+                ):
+                    default = field.default_factory()
+                else:
+                    default = field.default
+                if default not in (None, dataclasses.MISSING):
+                    kwargs.setdefault(key, default)
 
         missing_params = [p for p in required_params if p not in kwargs]
         if missing_params:
@@ -1410,9 +1406,8 @@ def _init_wrapper(
             # Fields with None defaults are intentionally skipped by the
             # default-handling above, but the dataclass __init__ would
             # still assign them. Ensure every expected key is present.
-            if not is_compiling():
-                for key in self.__expected_keys__:
-                    kwargs.setdefault(key, None)
+            for key in self.__expected_keys__:
+                kwargs.setdefault(key, None)
             if tensor_only:
                 # Fast path: validate and assign directly to the
                 # underlying dict, skipping the set → _set_str →
--- a/test/test_compile.py
+++ b/test/test_compile.py
@@ -71,6 +71,14 @@ pytestmark = pytest.mark.skipif(
 )
 
 
+@pytest.fixture(autouse=True)
+def _reset_dynamo_code_caches():
+    # Tests in this file all exercise torch.compile; a fresh Dynamo code cache
+    # between tests keeps recompile/guard-count assertions reliable and prevents
+    # one test's frame from leaking into the next.
+    torch._dynamo.reset_code_caches()
+
+
 def test_vmap_compile():
     # Since we monkey patch vmap we need to make sure compile is happy with it
     def func(x, y):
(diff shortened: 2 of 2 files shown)


vmoens:
@peterdsharpe
i'm willing to make a minor with these fixes but I'm flying today, so that will depend on wifi quality

peterdsharpe:
Thank you Vincent! We're about to cut a release for NVIDIA PhysicsNeMo and are currently planning to pin to `tensordict<0.12` to sidestep these; once the TensorDict minor release goes out we can switch it to `tensordict>=0.12.4`.

PR #1724, opened 2026-06-12
[Feature] Add tensordict.llm with History moved from torchrl

vmoens:
## Description

Moves `History`, `ContentBase`, `add_chat_template` and the base chat templates from torchrl (`torchrl/data/llm/history.py` at `13c7eca6`) into a new `tensordict.llm` module, making tensordict the canonical home.

`History` is already tensordict-native (a `TensorClass` whose only torchrl-internal dependencies were the logger and the `default_spec` helper). Moving it down the stack lets consumers that only need conversation parsing/formatting depend on released tensordict wheels instead of torchrl. A follow-up torchrl PR will re-export `torchrl.data.llm.History` from here so existing import paths keep working.

## Changes relative to the torchrl source

- `torchrl._utils.logger` replaced with `tensordict.utils.logger`.
- `History.default_spec` still returns a torchrl `Composite` spec; the torchrl import is gated behind `importlib.util.find_spec("torchrl")` with an informative `ImportError` when torchrl is not installed.
- Local `json`/`re`/`lazy_stack` imports hoisted to module top.
- Docstrings re-pointed at `tensordict.llm` import paths; torchrl-specific integration prose condensed into a note.
- `tensordict.llm` resolves lazily from the top-level package (PEP 562 `__getattr__`) so `import tensordict` stays light.

## Tests

`test/test_llm.py` ports `TestHistory` from torchrl `test/llm/test_data.py`, with `skipif` guards on transformers/torchrl/pillow so the suite degrades gracefully in minimal environments (torchrl runs these in a dedicated LLM CI where those deps are guaranteed). The torchrl-modules-dependent test (`_extract_responses_from_full_histories`) stays in torchrl.

Local run: 59 passed, 2 skipped (torchrl, pillow not installed), 11 xfailed (same as torchrl).

## Docs

New reference page `docs/source/reference/llm.rst` wired into the API reference toctree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tensordict/llm/history.py
+++ b/tensordict/llm/history.py
@@ -1,0 +1,1465 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from __future__ import annotations
+
+import dataclasses
+import importlib.util
+import json
+
+import re
+from typing import Literal, TYPE_CHECKING
+
+import torch
+
+from tensordict import (
+    lazy_stack,
+    LazyStackedTensorDict,
+    list_to_stack,
+    TensorClass,
+    TensorDict,
+)
+from tensordict.utils import _maybe_correct_neg_dim, logger as tensordict_logger
+
+if TYPE_CHECKING:
+    import transformers
+
+_has_torchrl = importlib.util.find_spec("torchrl") is not None
+
+
+# Global storage for custom templates and their metadata
+_CHAT_TEMPLATES = {
+    "chatml_format": """{%- for message in messages %}
+{%- if message['role'] == 'assistant' -%}
+{{ '<|im_start|>' + message['role'] + '\n' }}{% generation %}{{ message['content'] }}{% endgeneration %}{{ '<|im_end|>\n' }}
+{%- else -%}
+{{ '<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>\n' }}
+{%- endif %}
+{%- endfor %}
+{%- if add_generation_prompt -%}
+{{ '<|im_start|>assistant\n' }}
+{%- endif %}
+""",
+    "qwen": """
+{%- if tools %}
+    {{- '<|im_start|>system\\n' }}
+    {%- if messages[0]['role'] == 'system' %}
+        {{- messages[0]['content'] }}
+    {%- else %}
+        {{- 'You are a helpful assistant.' }}
+    {%- endif %}
+    {{- "\\n\\n# Tools\\n\\nYou may call one or more functions to assist with the user query.\\n\\nYou are provided with function signatures within <tools></tools> XML tags:\\n<tools>" }}
+    {%- for tool in tools %}
+        {{- "\\n" }}
+        {{- tool | tojson }}
+    {%- endfor %}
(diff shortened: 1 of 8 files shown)


test/test_llm.py
```diff
@@ -0,0 +1,1012 @@
...
+from tensordict import lazy_stack, set_list_to_stack
+from tensordict.llm import add_chat_template, ContentBase, History
+from tensordict.llm.history import (
+    _CHAT_TEMPLATES,
+    _CUSTOM_INVERSE_PARSERS,
+    _CUSTOM_MODEL_FAMILY_KEYWORDS,
+)
+from tensordict.utils import logger as tensordict_logger
+
+_has_transformers = importlib.util.find_spec("transformers") is not None
```
vmoens, line 26: untested in the CI

test/test_llm.py
```diff
@@ -0,0 +1,1012 @@
...
+_has_torchrl = importlib.util.find_spec("torchrl") is not None
```
vmoens: this is kind of circular - we should focus on tests that don't require torchrl

test/test_llm.py
```diff
@@ -0,0 +1,1012 @@
...
+        recovered = history._inv_chatml(data_str)
+        assert recovered.role == history.role, (recovered.role, history.role)
+        assert recovered.content == history.content
+        data_token = history.apply_chat_template(
+            tokenizer=tokenizer, tokenize=True, add_generation_prompt=False
+        )
+        recovered = history._inv_chatml(tokenizer.batch_decode(data_token)[0])
+
+    @pytest.mark.skipif(
+        not _has_torchrl, reason="History.default_spec requires torchrl"
```
vmoens: not great

tensordict/llm/history.py
```diff
@@ -0,0 +1,1465 @@
...
+                return self
+        if history.device != self.device:
+            if self.device is None:
+                history = history.copy().clear_device_()
+            else:
+                history = history.to(self.device)
+        return torch.stack(list(self.unbind(dim)) + list(history.unbind(dim)), dim=dim)
+
+    @classmethod
+    def default_spec(cls, shape=(-1,)):
```
vmoens: that is poorly designed. 
Either torchrl makes a free function, or we need another way of dealing with this

test/test_llm.py
vmoens: Fixed in 31819eed5 — transformers and pillow are now installed in the CI test env (environment.yml, plus a best-effort install in the 3.14t pip branch of setup_env.sh), so the chat-template tests actually run in CI.

_Response generated by Claude_

test/test_llm.py
vmoens: Fixed in 31819eed5 — the torchrl gate is gone; the test file no longer references torchrl at all.

_Response generated by Claude_

test/test_llm.py
vmoens: Fixed in 31819eed5 — test removed here. The spec test lives in torchrl's test_data.py, extended in pytorch/rl#3862 to also cover the new free function.

_Response generated by Claude_

tensordict/llm/history.py
vmoens: Fixed in 31819eed5 — default_spec is removed from tensordict entirely; the module now has zero torchrl references. The spec machinery moved to torchrl's re-export shim (pytorch/rl#3862): a history_default_spec() free function is the canonical API (internal call sites in mlgym and policies.common migrated to it), and the shim also attaches it as History.default_spec so the established classmethod API (mirroring ChatHistory/Text/Tokens) keeps working for existing callers without deprecation churn.

_Response generated by Claude_

PR #1725, opened 2026-06-14
[BugFix] Preserve NonTensor entries in pad; fix NonTensorStack boolean-mask indexing

vmoens:
## Summary

`tensordict.pad` silently destroyed NonTensor entries: padding a TensorDict containing a `NonTensorStack` returned the entry as a stack of **empty** TensorDicts for *all* positions, valid ones included. Repro:

```python
import torch
from tensordict import NonTensorStack, TensorDict, pad
td = TensorDict({"x": torch.arange(3).unsqueeze(-1), "instr": NonTensorStack("a", "b", "c")}, batch_size=[3])
p = pad(td, [0, 2])
p["instr"]  # before: stack of empty TensorDicts — 'a','b','c' gone;  after: ['a','b','c',None,None]
p["x"]      # correctly zero-padded
```

Root cause: `NonTensorStack` is a `LazyStackedTensorDict` subclass, so it fell into the generic lazy-stack branch and had its leaves filled with the numeric `value`.

## What changed

**`pad` (functional.py)** now routes non-tensor entries to a dedicated path. Valid positions keep their original values; each pad slot holds a `NonTensorData` with `data=None` (a non-tensor value can't be filled with a numeric constant). Covered cases:

- top-level `NonTensorData` / `NonTensorStack` inputs and nested entries
- multi-dimensional padding and negative (cropping) pads
- broadcast `NonTensorData` (single value over a batch dim)
- unbatched (`batch_size=[]`) non-tensor metadata is carried through untouched
- `inplace=True`: `NonTensorStack` identity is preserved; the lazy-stack stack-dim grow path fills new slots with `None` for non-tensor leaves
- `safe=True` pre-flight validates non-tensor pad sizes too

**Boolean-mask indexing (`_lazy.py`)** — two related sharp edges surfaced by the same downstream code:

- Masking dim 0 of a 2-D `NonTensorStack` raised `IndexError` because a scalar bool inside a tuple index was treated like a scalar integer. Scalar bools now follow numpy new-axis semantics (`True` -> size-1 dim, `False` -> size-0 dim) in both `__getitem__` and `__setitem__`.
- Indices selecting nothing (empty slice along the stack dim, all-False mask) raised `"items cannot be empty"` or produced a wrong batch size (`[0, 0, ...]`); they now return a correctly-shaped zero-length lazy stack that still interops with `torch.cat`.

**`LinkedList` (utils.py)** printed doubly-wrapped (`LinkedList(LinkedList([...]))`) because `__str__` called `super().__str__()`, which dispatched back to `__repr__`. Dropped the redundant `__str__`.

## Tests

Added to `test/test_tensordict.py` (extends existing pad / lazy-stack / non-tensor coverage):

- `test_pad_nontensor`, `_2dim`, `_broadcast_data`, `_inplace`, `_top_level`, `_safe`, `_lazy_stack`
- `test_lazy_mask_nested_stack`, `test_lazy_scalar_bool_index`, `test_lazy_empty_selection_batch_size`
- `test_nontensorstack_boolean_mask`, `_2d`, `_scalar_bool_index`, `test_linked_list_str`

Full `test/test_tensordict.py` (7716 passed), pad/index/mask subsets, `test_compile.py` subset, and `test_tensorclass.py` non-tensor/stack subset all pass.

## Downstream

torchrl's collector `trajs_per_batch` path (`torchrl/collectors/utils.py::_traj_emit`) carries a workaround that pads non-tensor keys separately; once this lands that workaround can be removed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tensordict/_lazy.py
+++ b/tensordict/_lazy.py
@@ -89,7 +89,6 @@ from tensordict.utils import (
 from torch import Tensor
 from torch.nn.utils.rnn import pad_sequence
 
-
 try:
     from functorch import dim as ftdim
 
@@ -853,10 +852,25 @@ class LazyStackedTensorDict(TensorDictBase):
             cursor_incr = 1
             # if idx is None:
             #     idx = True
+            if (
+                isinstance(idx, torch.Tensor)
+                and idx.ndim == 0
+                and idx.dtype == torch.bool
+            ):
+                # A scalar boolean behaves like a new axis (numpy semantics),
+                # not like a scalar integer index: don't let it fall through
+                # to the _is_number paths below.
+                idx = bool(idx)
             if idx is None or idx is True:
                 out.append(None)
                 num_none += cursor <= self.stack_dim
                 continue
+            if idx is False:
+                raise NotImplementedError(
+                    "Indexing a LazyStackedTensorDict with a scalar False mask "
+                    "inside a tuple index is not supported. Use a length-1 "
+                    "boolean mask on an unsqueezed tensordict instead."
+                )
             if cursor == self.stack_dim:
                 # we need to check which tds need to be indexed
                 if isinstance(idx, ftdim.Dim):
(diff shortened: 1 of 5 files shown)


vmoens:
Added a second commit `[CI] Fix broken logging call in test_distributed exposed by pytest 9.1`.

The CPU/macOS jobs were failing on `TestDTensor::test_dtensor` (and the FSDP tests) with `TypeError: not all arguments converted during string formatting`. Root cause is unrelated to the pad change:

- `test/test_distributed.py` had `tdlogger.info("start method already set to", mp.get_start_method())` — a logging call whose message lacks a `%s` placeholder, so `msg % args` raises.
- This was latent for a long time (the logging framework swallowed the formatting error) until **pytest 9.1.0** (landed between Jun 12 and Jun 15) changed its log-capture handler to let the exception propagate, turning it into a hard test failure. Confirmed: passing runs used pytest 9.0.3, failing runs use 9.1.0.

The fix adds the missing `%s` placeholder. The remaining red jobs are the GPU / `Unit-tests (RL)` jobs, which are pre-existing failures on `main` (flaky distributed/RL-GPU), not introduced here.

PR #1734, opened 2026-07-08
[BugFix] Skip missing TensorDictParams submodule when loading state_dict with strict=False

### LINKED ISSUE #1472: [BUG] `load_state_dict` fails if checkpoint lacks entries for TensorDictParams
The context is rather hard to explain, but I have already identified the problem.

Culprit: 
https://github.com/pytorch/tensordict/blob/f7b5e059899f7bfb18aeb6b0f29ce68489121265/tensordict/nn/params.py#L1244-L1247

If `prefix` is not present in `data`, L1246 would set `data = None` and the subsequent line fails. I guess the following codeblock would solve it?

```python
        if prefix:
            data = data.get(prefix)
        if data is not None:
            self.data.load_state_dict(data)

```

javierdejesusda:
## Description

Loading a `state_dict` that has no entries under a `TensorDictParams` submodule
prefix crashes. In `_load_from_state_dict`, `data = data.get(prefix)` returns
`None` when the checkpoint contains nothing for that prefix, and the code then
calls `self.data.load_state_dict(None)` unconditionally, which raises. This is
the normal path for a `strict=False` load where a submodule is simply absent
from the checkpoint.

This PR guards the load so a missing submodule under `strict=False` becomes a
no-op, matching the `nn.Module` contract:

```python
if data is not None:
    self.data.load_state_dict(data)
```

## Motivation and Context

close #1472

A module holding `TensorDictParams` cannot be partially restored from a
checkpoint that omits those params, even with `strict=False`. `nn.Module`
treats missing keys as a no-op in that case, so `TensorDictParams` should too
instead of raising.

## Types of changes

## Checklist

### DIFF
--- a/tensordict/nn/params.py
+++ b/tensordict/nn/params.py
@@ -1257,9 +1257,20 @@ class TensorDictParams(TensorDictBase, nn.Module):  # type: ignore[override,misc
             },
             [],
         ).unflatten_keys(".")
-        prefix = tuple(key for key in prefix.split(".") if key)
-        if prefix:
-            data = data.get(prefix)
+        nested_prefix = tuple(key for key in prefix.split(".") if key)
+        if nested_prefix:
+            data = data.get(nested_prefix)
+        if data is None:
+            return nn.Module._load_from_state_dict(
+                self,
+                state_dict,
+                prefix,
+                local_metadata,
+                strict,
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks!

PR #1753, opened 2026-07-21
[BugFix] Convert Mapping values for TensorDict fields in tensor-only tensorclasses

### LINKED ISSUE #1752: [BUG] tensor_only tensorclasses wrap non-dict Mapping values assigned to TensorDict fields
## Describe the bug

For a `tensor_only=True` tensorclass field annotated as `TensorDict`, a plain `dict` value is converted to a `TensorDict`, but other `collections.abc.Mapping` implementations are not. For example, `UserDict` is wrapped as `NonTensorData` instead.

As a result, contained mapping keys are unavailable through the expected TensorDict interface. This also affects downstream mapping-compatible containers such as PyVista's `DataSetAttributes`.

## To Reproduce

```python
from collections import UserDict

import torch

from tensordict import TensorDict, tensorclass


@tensorclass(tensor_only=True)
class State:
    data: TensorDict


state = State(data=UserDict({"position": torch.ones(())}))

print(type(state.data))
print(state)
```

Current output:

```text
<class 'tensordict.tensorclass.NonTensorData'>
State(
    data=NonTensorData(data={'position': tensor(1.)}, batch_size=torch.Size([]), device=None),
    batch_size=torch.Size([]),
    device=None,
    is_shared=False)
```

A plain `dict` containing the same data is converted to a `TensorDict`.

## Expected behavior

When the destination field has a TensorDict-like annotation and the supplied value is a `Mapping`, the value should be constructed as a `TensorDict`, including nested mappings.

The behavior should be annotation-aware: arbitrary mappings assigned to fields annotated as `Any` or explicitly intended as non-tensor data should retain their current behavior.

Existing `TensorDict` values should also remain unchanged rather than being reconstructed, so identity, lock state, and other metadata are preserved.

## System info

TensorDict was installed from PyPI.

```text
tensordict 0.12.4
numpy 2.4.6
python 3.13.8
platform linux
torch 2.12.0+cu130
```

## Additional context

The issue is particularly visible in `tensor_only=True` tensorclasses, whose optimized initialization path passes values directly through tensor validation. Downstream code currently has to materialize mapping implementations as plain dictionaries before constructing the tensorclass.

The same behavior should be consistent across construction, attribute assignment, `.set()`, and `from_dataclass`.

## Reason and Possible fixes

The TensorDict conversion path recognizes plain dictionaries, but does not apply equivalent handling to general `Mapping` implementations.

A narrowly scoped fix could cache fields whose resolved annotations are TensorDict-like and normalize `Mapping` values only for those fields before tensor validation. This should include generic and optional/union TensorDict annotations, preserve existing TensorDict instances, and retain the `tensor_only=True` fast path for classes without relevant fields.

## Checklist

peterdsharpe:
## Description

This PR converts arbitrary `collections.abc.Mapping` values assigned to
TensorDict-annotated fields on `tensor_only=True` tensorclasses.

The annotation-aware conversion applies consistently during construction,
attribute assignment, `.set()`, and `from_dataclass`. It also:

- recursively normalizes nested mappings;
- supports generic and optional/union TensorDict annotations;
- preserves existing TensorDict instances, including identity and lock state;
- leaves mappings annotated as `Any` or `NonTensorData` unchanged;
- preserves the direct constructor fast path when no relevant fields exist.

## Motivation and Context

Closes #1752.

Non-`dict` Mapping implementations were previously wrapped as `NonTensorData`
even when the destination field was annotated as `TensorDict`. The fix is
scoped using resolved field annotations so that it does not globally reinterpret
mappings intended to remain non-tensor data.

## Test plan

Regression tests cover:

- `dict` and `UserDict` inputs;
- nested mappings;
- construction, attribute assignment, and `.set()`;
- generic and optional TensorDict annotations;
- class and instance forms of `from_dataclass`;
- preservation of existing and nested locked TensorDict instances;
- `Any` and `NonTensorData` mapping fields remaining non-tensor.

Local validation:

- `pytest test/tensorclass/test_tensorclass.py -k "TestTensorOnly"`:
  16 passed
- `pytest test/tensorclass/test_tensorclass.py -k "not multiprocessing"`:
  160 passed, 1 skipped, 1 deselected
- `pytest test/tensorclass/test_compatibility.py test/tensorclass/test_typedtensordict.py`:
  174 passed, 14 skipped
- Ruff and `git diff --check` pass

## Types of changes

## Checklist

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -19,11 +19,13 @@ import pickle
 import shutil
 import sys
 import warnings
+from collections.abc import Mapping
 from copy import copy, deepcopy
 from dataclasses import dataclass
 from pathlib import Path
 from textwrap import indent
 from typing import (
+    AbstractSet,
     Any,
     Callable,
     get_args,
@@ -772,7 +774,7 @@ def from_dataclass(
         else:
             cls = obj
         clz = _tensorclass(cls, frozen=frozen, shadow=shadow, tensor_only=tensor_only)
-        clz._type_hints = get_type_hints(obj)
+        _set_tensorclass_type_hints(clz, get_type_hints(obj))
         clz._autocast = autocast
         clz._nocast = nocast
         clz._shadow = shadow
@@ -794,6 +796,13 @@ def from_dataclass(
             shadow=shadow,
             tensor_only=tensor_only,
         )
+        try:
+            type_hints = get_type_hints(type(obj))
+        except (NameError, TypeError):
+            # Preserve the existing fallback for unresolved annotations.
+            pass
+        else:
+            _set_tensorclass_type_hints(clz, type_hints)
         clz._autocast = autocast
         clz._nocast = nocast
         clz._shadow = shadow
@@ -1440,10 +1449,15 @@ def _init_wrapper(
                 _td_dict = _td._tensordict
                 _non_td = self._non_tensordict
                 _validate = _td._validate_value
+                _tensordict_fields = type(self)._tensordict_fields
                 for key, value in kwargs.items():
                     if value is None:
                         _non_td[key] = None
                     else:
+                        if _tensordict_fields:
+                            value, _ = _convert_mapping_for_field(
+                                key, value, _tensordict_fields
+                            )
                         _td_dict[key] = _validate(
                             value, check_shape=True, non_blocking=False
                         )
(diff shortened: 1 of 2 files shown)


vmoens (approved):
Made a small edit in [849271b](https://github.com/pytorch/tensordict/pull/1753/commits/849271ba583deaaa74f3c08e77716855ffec5cbe), otherwise lgtm!

PR #1754, opened 2026-07-21
[Feature] Add zarr storage backend (to_zarr/from_zarr)

vmoens:
## Description

Adds zarr (`zarr>=3.0`) as a storage backend for `PersistentTensorDict`, mirroring the h5 integration:

- `td.to_zarr(path_or_store)` / `TensorDict.from_zarr(...)` / `tensordict.from_zarr(...)` / `from_schema(storage="zarr")`, all backed by `PersistentTensorDict(backend="zarr")`. `memmap()` is untouched: its zero-copy mmap contract does not apply to zarr's chunked model.
- Stores: local directory (`LocalStore`), `ZipStore`, or any `zarr.abc.store.Store` (e.g. fsspec/S3).
- Non-tensor entries are stored as 1D `uint8` JSON/pickle payloads marked with a `__tensordict_non_tensor__` array attribute; strings round-trip as `str` (h5 returns `bytes`).
- `batch_size` and dim names are persisted in a `__tensordict__` group attribute and restored by `from_zarr`; stores written by other tools (xarray, raw zarr) fall back to automatic batch-size inference.
- Default layout is one uncompressed chunk per leaf (checkpoint-friendly, honest apples-to-apples with the other formats); `chunks=`/`compressors=` pass through to `zarr.Group.create_array`, with `chunks` interpreted against the leading dims of each leaf so a single spec works across heterogeneous ranks.
- Out-of-core indexed reads/writes (`get_at`/`set_at_`) use a tiered strategy: basic indexing, `oindex` for dim-0 boolean masks, and a materializing fallback with a performance warning for index types zarr cannot serve out-of-core.

## Commit breakdown

1. `[Refactor]` extracts a `_PersistentBackend` adapter from `PersistentTensorDict` — all h5py-specific I/O (including both boolean-mask workarounds) moves into `_H5Backend`, behavior-preserving, so backends are selected via a registry.
2. `[Feature]` adds `_ZarrBackend`, the public API, exports/stubs/tensorclass/`TensorDictParams` wiring, the `zarr` extra, CI install (tolerant: zarr 3 needs Python >= 3.11), and `test/memmap/test_zarr.py` (64 tests). Also fixes `PersistentTensorDict.memmap()` and `clone(recurse=True)` silently dropping non-tensor entries (affected h5 too).
3. `[Benchmark]` adds zarr to `benchmarks/scripts/serialization_formats_bench.py` (save/open/read/copy) and a CI pytest-benchmark pair in `memmap_benchmarks_test.py`.
4. `[Doc]` adds the storage.rst backend section, saving.rst format-comparison notes, td.rst autosummary entries, and a new `zarr_storage.py` sphinx tutorial; the serialization-speed docs figure now includes zarr.
5. `[Performance]` cuts read-path metadata overhead: `to_zarr` writes consolidated metadata (kept fresh across structural mutations, skipped on append-only stores), read-mode opens use it, and batch-size validation runs on store metadata instead of re-fetching every array. Deep-tree reads went from 9.3s to 30ms, many-small reads from 4.6s to 1.05s on the benchmark payloads.

## Benchmarks (laptop SSD, warm cache, 1 GB payloads)

| layout | op | memmap dir | consolidated | .tdz | torch.save | zarr |
|---|---|---|---|---|---|---|
| flat, 8 leaves | save | 128ms | 113ms | 106ms | 605ms | 266ms |
| flat, 8 leaves | read | 23ms | 20ms | 31ms | 32ms | 84ms |
| 2000 small leaves | save | 245ms | 149ms | 503ms | 674ms | 4.2s |
| 2000 small leaves | open (lazy) | 101ms | 5.6ms | 25ms | 52ms | 41ms |
| 2000 small leaves | read | 164ms | 65ms | 95ms | 114ms | 1.15s |

zarr opens lazily and competitively but does not mmap and pays one store round-trip per array, so bulk reads and many-tiny-leaf saves favor the native formats; its value is chunking, compression, object-store backends and xarray/dask interop. This trade-off is stated in the docs.

## Tests

- `pytest test/memmap/` (627 incl. 64 new zarr tests) and `pytest test/tensordict/test_generic.py` green; PTD-related selections of `test_methods.py`/`test_nontensor.py`/`test_mp.py`/`test_compatibility.py` green.
- Skip-gated on zarr availability; `zarr_storage.py` tutorial executes end to end.

Generated with [Claude Code](https://claude.com/claude-code)


docs/source/storage.rst
```diff
@@ -192,6 +201,65 @@ Keyword arguments forwarded by ``from_schema``:
...
+implementation, and chunked arrays support parallel reads and writes to
+separate chunks.  Requires ``zarr>=3.0`` (``pip install "tensordict[zarr]"``).
+
+.. code-block:: python
+
+   >>> import torch, tempfile
+   >>> from tensordict import TensorDict
+
+   >>> d = tempfile.mkdtemp()
+   >>> td = TensorDict.from_schema(
```
vmoens, line 221: @peterdsharpe i really like this!

PR #1756, opened 2026-07-29
[CI] Add Linux aarch64 nightly wheels

peterdsharpe:
## Description

Build and publish `tensordict-nightly` wheels for Linux aarch64 using the existing PyTorch binary build matrix.

## Motivation and Context

Closes #1755.

- [x] I have raised an issue to propose this change ([required](https://github.com/pytorch/tensordict/issues) for new features and bug fixes)

## Types of changes

- [x] New feature (non-breaking change which adds core functionality)

## Checklist

- [x] I have read the [CONTRIBUTION](https://github.com/pytorch/tensordict/blob/main/CONTRIBUTING.md) guide (**required**)
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.

PR #1757, opened 2026-08-04
[CI] Fix Linux aarch64 nightly wheel builds

peterdsharpe:
Today's nightly build for the new Linux aarch64 target (added by my previous PR #1756) failed because of a pretty silly shell command-scoping bug: only the initial `sed` command ran through test-infra's Conda wrapper. Everything after the first `&&`, including `pip wheel`, ran outside the prepared build environment.

Log:
https://github.com/pytorch/tensordict/actions/runs/30919780779

This fixes that by moving the complete nightly wheel build into a small script, so the whole sequence runs inside the Conda environment and `pyproject.toml` is restored even if the build fails. It also builds with `--no-deps`, since test-infra has already installed nightly PyTorch, and installs the split `libpython-static` package that CMake needs for Linux Python 3.15/3.15t builds.

I reproduced the full Linux aarch64 matrix natively (Python 3.10–3.15, including 3.14t and 3.15t). All eight wheels built successfully, passed test-infra's manylinux repair step, and passed `twine check`.

### DIFF
--- a/.github/scripts/build-aarch64-nightly-wheel.sh
+++ b/.github/scripts/build-aarch64-nightly-wheel.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -euo pipefail
+
+sed -i.bak 's/name = "tensordict"/name = "tensordict-nightly"/' pyproject.toml
+trap 'mv pyproject.toml.bak pyproject.toml' EXIT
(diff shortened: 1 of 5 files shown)


vmoens (approved):
LGTM

Issue #1755, opened 2026-07-29
[Feature Request] Linux arm64 wheel builds for `tensordict-nightly`


peterdsharpe:
## Motivation

I noticed that [mainline TensorDict releases (e.g., 0.13.0) on PyPI (`tensordict`) have wheels for](https://pypi.org/project/tensordict/0.13.0/#files):

- Windows x86-64
- macOS arm64
- Linux x86-64
- Linux arm64

However, [`tensordict-nightly` releases only have](https://pypi.org/project/tensordict-nightly/2026.7.29/#files):

- Windows x86-64
- macOS universal2 (arm64, x86-64)
- Linux x86-64

Notably, Linux arm64 is missing.

We'd like to use TensorDict in Linux `aarch64` environments (e.g., B200 supercompute systems with ARM-based NVIDIA Grace CPUs), particularly with latest features from nightly.

## Solution

Top priority:

- Include Linux aarch64 wheels in nightly releases

Lower priorities, but perhaps worth knocking out as part of the same fix:

- Include wheels for free-threaded Python in nightly releases
- Switch mainline macOS releases from arm64 to universal2 (covering arm64 and x86-64)

## Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

PR #1760, opened 2026-08-05
[CI] Fix Linux aarch64 nightly wheel uploads

peterdsharpe:
## Description

The aarch64 nightly saga continues (#1755, #1756, #1757).

In [last night's run](https://github.com/pytorch/tensordict/actions/runs/31015768639), aarch64 builds worked just fine on Python 3.10-3.14t. 

However, uploads failed because they tried to upload to S3+PyPI, rather than just PyPI. The S3 half fails because the PyTorch org's OIDC trust policy doesn't include `pytorch/tensordict` (unlike vision/audio/ao etc.). Because these S3 and PyPI steps were bundled, the PyPI step never ran.

This fixes that by borrowing the upload code from the other `tensordict-nightly` platforms.

Side notes:

- Python 3.15/3.15t builds fail for an unrelated reason (pytorch/test-infra#8357 made 3.15 smoke tests resolve all runtime deps from download.pytorch.org, where pyvers isn't mirrored, I'll chase that separately). Not aarch64 specific, but aarch64 is the only platform that attempts a 3.15/3.15t build (other platforms stop at 3.14), so it's surfacing here. Not immediately critical to fix as 3.15 isn't released yet, but something to fix later.
- The upload runs with `!cancelled()` specifically because 3.15/3.15t is a WIP - this lets successful builds (3.10-3.14t) still publish.
- No `continue-on-error` on the new upload job: if uploads break again I'd rather the run go red than silently green.

Fingers crossed...

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/build-wheels-aarch64-linux.yml
+++ b/.github/workflows/build-wheels-aarch64-linux.yml
@@ -24,9 +24,6 @@ on:
         required: false
         type: boolean
         default: false
-    secrets:
-      PYPI_TOKEN:
-        required: false
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


vmoens (approved):
Thanks
We'll dig into 3.15 later. 3.14 compatibility isn't a fully told story either so we'll need to think this through

PR #1763, opened 2026-08-10
[Deprecation] Make memmap pickle loading explicit

peterdsharpe:
## Summary

- expose an explicit pickle policy across the memmap load and refresh APIs
- preserve compatibility in 0.13 with a warning ahead of the 0.14 default change
- preserve trusted save round-trips and remove obsolete pickle sidecars when metadata no longer needs them

## User impact

Callers can reject pickle while loading less-trusted artifacts, while workflows that deliberately store arbitrary Python metadata can opt in explicitly. Existing callers receive a transition warning only when a pickle sidecar is encountered.

## Root cause

Non-tensor values deliberately fall back to pickle when JSON cannot represent them, but the corresponding load path did not expose a caller-selectable policy. Reusing a save directory could also retain an obsolete sidecar after the value became JSON-compatible.

## Validation

- 341 relevant tests passed, with 3 expected skips
- 33 focused policy, execution-boundary, overwrite, and slow multiprocessing tests passed against the isolated commit
- ufmt, Black, flake8, pydocstyle, RST title, and diff checks passed


vmoens (approved):
LGTM thanks!

PR #1771, opened 2026-08-25
[Feature] add key-level skip_existing

### LINKED ISSUE #352: [Feature Request] key-level granularity in `skip_existing`
Currently `skip_existing` operates on all keys without any granularity.

This is a problem in RL when in a loss module for example you may want to skip existing "values" but you definitely never want to skip existing "memory" in a memory based model (RNN). AKA if you use skip_existing on memory keys you will never update your memory.

This is needed to support rnns in torch rl (issue https://github.com/pytorch/rl/pull/1060)

We need a solution to make `skip_existing` more granular.

This is really simple and consists in feeding to the `set_skip_existing` funtion the keys we actually want to skip.
```
with set_skip_existing(["value", "value_target]):
     loss(td) # Will use existing values but not existing hidden memory
```
by default, if no keys are passed, the beahvior remains the same as the current `set_skip_existing=True`

gtnv:
### TLDR

`set_skip_existing(["value"])` used list truthiness, so any non-empty key list behaved like `True` and could skip unrelated recurrent-state updates.

This normalizes the selected keys and skips a module only when all of its outputs are selected and already present. Cached outputs stay untouched, while modules with unselected outputs still run.

Closes #352

<details>
<summary>Testing</summary>

### Reproduction

```python
import torch
from tensordict.nn import TensorDictModule, TensorDictSequential, set_skip_existing

model = TensorDictSequential(
    TensorDictModule(
        lambda value: value + 1,
        in_keys=["input"],
        out_keys=["value"],
    ),
    TensorDictModule(
        lambda memory: memory + 1,
        in_keys=["memory"],
        out_keys=[("next", "memory")],
    ),
)
data = TensorDict(
    {
        "input": torch.tensor(0),
        "value": torch.tensor(10),
        "memory": torch.tensor(0),
        "next": {"memory": torch.tensor(10)},
    },
    [],
)

with set_skip_existing(["value"]):
    model(data)

assert data["value"].item() == 10
assert data["next", "memory"].item() == 1
```

Before this change, the final assertion fails because ("next", "memory") remains 10.

```
python -m pytest test/nn/test_nn.py::TestSkipExisting -q
python -m pytest test/nn -q
pre-commit run --files tensordict/nn/utils.py test/nn/test_nn.py
```

Results:
- 8 passed in the focused suite.
- 611 passed, 1 skipped in test/nn; ONNX was unavailable.
- All configured formatting and lint hooks pass.

</details>

### DIFF
--- a/tensordict/nn/utils.py
+++ b/tensordict/nn/utils.py
@@ -12,7 +12,8 @@ from enum import Enum
 from typing import Any, Callable
 
 import torch
-from tensordict.utils import _ContextManager, strtobool
+from tensordict._nestedkey import NestedKey
+from tensordict.utils import _ContextManager, strtobool, unravel_key_list
 from torch import nn
 
 from torch.utils._contextlib import _DecoratorContextManager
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks!

PR #1772, opened 2026-08-25
[BugFix] keep ragged lazy stacks during batched indexing

### LINKED ISSUE #1296: [BUG] DataLoader with LazyStackedTensorDict of different sizes
## Describe the bug

Not sure if I am missing something but I am trying to load a batch of `lazy_stacked` `TensorDicts` that have variable size tensors.
The `DataLoader `throws an error however that it cannot stack the tensors.

## To Reproduce
```python
import tensordict
import torch

tensors = [{"x": torch.rand((i,))} for i in range(10)]
tensordicts_stacked = tensordict.lazy_stack(
    [tensordict.TensorDict.from_dict(x) for x in tensors]
)

dl = DataLoader(tensordicts_stacked, batch_size=4, collate_fn=lambda x: x)
next(iter(dl))
```
## Expected behavior

Not sure if this is supposed to work but I would have expected it to return a batch of `LazyStackedTensorDict`s.

## Reason and Possible fixes

I think it's because `__getitems__` actually points to `<bound method TensorDictBase.__getitem__ of LazyStackedTensorDict`.
If I add `__getitems__ = __getitem__` after the `__getitem__` in `LazyStackedTensorDict` (the same as in the Base) it returns a `LazyStackedTensorDict` without an error.

## Checklist

gtnv:
### TLDR

LazyStackedTensorDict inherited batched indexing from the base class, which tried to stack ragged values. Routing it through its own indexing keeps the result lazy.

Closes #1296.

<details>
<summary>Reproduction</summary>

```python
import torch

from tensordict import LazyStackedTensorDict, TensorDict, lazy_stack

tensordict = lazy_stack([TensorDict(x=torch.zeros(i)) for i in range(5)])
indices = [3, 0, 4]
batch = tensordict.__getitems__(indices)

assert isinstance(batch, LazyStackedTensorDict)
assert [len(td["x"]) for td in batch] == indices
```

On `main`, batched indexing raises while stacking the unequal tensors. With this change, the ragged selection stays lazy.

</details>

<details>
<summary>Tests</summary>

```bash
python -m pytest -q \
  test/tensordict/test_lazy.py::TestLazyStackedTensorDict::test_batched_indexing_ragged_lazy_stack

python -m pytest -q test/tensordict/test_lazy.py
python -m pytest -q test/tensordict
```

Results: 1 passed; 667 passed, 3 skipped; 7,944 passed, 899 skipped.

</details>

### DIFF
--- a/tensordict/_lazy.py
+++ b/tensordict/_lazy.py
@@ -2696,6 +2696,8 @@ class LazyStackedTensorDict(TensorDictBase):
                     result.lock_()
                 return result
 
+    __getitems__ = __getitem__
+
     def __eq__(self, other):
         return self._dispatch_comparison(other, "__eq__", "__eq__", default=False)
 
--- a/test/tensordict/test_lazy.py
+++ b/test/tensordict/test_lazy.py
@@ -275,6 +275,13 @@ class TestLazyStackedTensorDict:
         test_td = LazyStackedTensorDict(*batches)
         assert_allclose_td(td, test_td)
 
+    def test_batched_indexing_ragged_lazy_stack(self):
+        tensordict = lazy_stack([TensorDict(x=torch.zeros(i)) for i in range(5)])
+        indices = [3, 0, 4]
+        batch = tensordict.__getitems__(indices)
+        assert isinstance(batch, LazyStackedTensorDict)
+        assert [len(td["x"]) for td in batch] == indices
+
     def test_all_keys(self):
         td = TensorDict({"a": torch.zeros(1)}, [])
         td2 = TensorDict({"a": torch.zeros(2)}, [])


vmoens (approved):
Please remove `DataLoader` references from the entire PR artifact. 

It appears in the title, body/reproduction, branch name, commit subject, and [[new test](https://github.com/pytorch/tensordict/blob/6cc46a5d8b77852fcb816cffdda3c30997ecdded/test/tensordict/test_lazy.py#L278-L288)](https://github.com/pytorch/tensordict/blob/6cc46a5d8b77852fcb816cffdda3c30997ecdded/test/tensordict/test_lazy.py#L278-L288). Frame this as batched indexing instead, and test `__getitems__` directly. Suggested title: `[BugFix] Preserve ragged lazy stacks during batched indexing`.

The implementation itself is correct so I am approving it but there should not be cross-reference between rl and td in PR/issues unless it provides some context and helps discovery.

gtnv:
cant change branch name bc of [this](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch) (gh docs). everything else will be fixed. sorry about that.

PR #1775, opened 2026-08-27
[BugFix] Restore unpickling of frozen tensorclasses

### LINKED ISSUE #1774: [BUG] frozen tensorclass cannot be unpickled (FrozenInstanceError in __setstate__)
## Describe the bug

A `frozen=True` tensorclass can be pickled but not unpickled. `_setstate` restores the instance by assigning to `self._tensordict` / `self._non_tensordict`, and the dataclass frozen guard rejects that, so `pickle.loads` raises `FrozenInstanceError`. `copy.copy` fails the same way, since it also goes through `__setstate__`. (`copy.deepcopy` is fine — it is routed through `TensorDictBase.__deepcopy__` and never reaches `_setstate`.)

In practice this surfaces as a `DataLoader` with `num_workers > 0` dying in the pin-memory thread having produced no batch, because the cross-process hand-off is the first thing that unpickles the batch.

## To Reproduce

```python
import copy
import pickle

import torch
from tensordict import TensorClass


class Frozen(TensorClass, frozen=True):
    x: torch.Tensor


f = Frozen(x=torch.zeros(3), batch_size=[3])
payload = pickle.dumps(f)  # fine
pickle.loads(payload)      # FrozenInstanceError
copy.copy(f)               # FrozenInstanceError
```

```bash
  File "repro.py", line 15, in <module>
    pickle.loads(payload)
  File ".../tensordict/tensorclass.py", line 1966, in _setstate
    self._tensordict = state.get("tensordict")
    ^^^^^^^^^^^^^^^^
  File "<string>", line 4, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field '_tensordict'
```

## Expected behavior

A frozen tensorclass round-trips through `pickle` and `copy.copy`, like a non-frozen one does.

## System info

Installed with `pip`. Also reproduced against `main` built from source (`ebcae8c65c1f896f0b1f4cfd656a39120cfcaba8`), where the two lines are unchanged.

```python
# 0.13.0 2.5.0 3.12.12 linux 2.11.0+cu128
```

## Reason and Possible fixes

[`tensordict/tensorclass.py#L1958-L1965`](https://github.com/pytorch/tensordict/blob/ebcae8c65c1f896f0b1f4cfd656a39120cfcaba8/tensordict/tensorclass.py#L1958-L1965):

```python
def _setstate(self, state: dict[str, Any]) -> None:
    self._tensordict = state.get("tensordict")
    self._non_tensordict = state.get("non_tensordict")
```

`_setstate` is installed unconditionally on every tensorclass in `_tensorclass`, frozen ones included, but a frozen class has the dataclass guard as its `__setattr__`.

Going through `object.__setattr__` fixes both cases and is a no-op for non-frozen classes:

```python
object.__setattr__(self, "_tensordict", state.get("tensordict"))
object.__setattr__(self, "_non_tensordict", state.get("non_tensordict"))
```

This is the same escape hatch the frozen `__init__` already uses (see the comment above `_has_custom_setattr`).

Happy to send a PR with this plus regression tests.

## Checklist

simonschlaepfer:
## Description

`_setstate` now restores `_tensordict` / `_non_tensordict` through `object.__setattr__` instead of plain assignment, so a `frozen=True` tensorclass can be unpickled.

## Motivation and Context

Closes #1774.

`_setstate` is installed on every tensorclass in `_tensorclass`, frozen ones included, but it assigns to `self._tensordict`, which the dataclass frozen guard rejects. A frozen tensorclass therefore pickles fine and raises `FrozenInstanceError` on load; `copy.copy` fails identically, since it goes through the same `__setstate__`. (`copy.deepcopy` is unaffected — it is routed through `TensorDictBase.__deepcopy__`.)

The practical symptom is a `DataLoader` with `num_workers > 0` dying in the pin-memory thread having produced no batch.

`object.__setattr__` is the same escape hatch the frozen `__init__` already uses, and is a no-op for non-frozen classes.

## Types of changes

## Test plan

Added `TestTensorClass::test_pickle_frozen` and `TestTensorClass::test_copy_frozen`, alongside the existing `test_pickle`, plus a module-level `MyDataFrozen` for them to pickle (a locally defined class cannot).

Both fail on `main` with `FrozenInstanceError` and pass with the fix. Local validation, built from source:

- `pytest test/tensorclass/` — 320 passed, 16 skipped
- `pytest test/tensordict -k "pickle or copy or serial or setstate"` — 120 passed, 2 skipped

## Checklist

### DIFF
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1963,11 +1963,14 @@ def _setstate(self, state: dict[str, Any]) -> None:  # noqa: D417
     Args:
         state (dict): State parameter to set the object
     """
-    # after dataclass(), frozen=True installs a guard __setattr__ that would reject a
-    # plain assignment here. object.__setattr__ is the same bypass the frozen __init__
-    # uses (see _has_custom_setattr in _tensorclass).
-    object.__setattr__(self, "_tensordict", state.get("tensordict"))
-    object.__setattr__(self, "_non_tensordict", state.get("non_tensordict"))
+    if type(self)._frozen:
+        # Bypass the dataclass guard while preserving custom __setattr__ behavior
+        # for non-frozen tensorclasses.
+        object.__setattr__(self, "_tensordict", state.get("tensordict"))
+        object.__setattr__(self, "_non_tensordict", state.get("non_tensordict"))
+    else:
+        self._tensordict = state.get("tensordict")
+        self._non_tensordict = state.get("non_tensordict")
 
 
 def _getattr_tensor_only(self, item: str, **kwargs) -> Any:
--- a/tensordict/tensorclass.py
+++ b/tensordict/tensorclass.py
@@ -1963,8 +1963,11 @@ def _setstate(self, state: dict[str, Any]) -> None:  # noqa: D417
     Args:
         state (dict): State parameter to set the object
     """
-    self._tensordict = state.get("tensordict")
-    self._non_tensordict = state.get("non_tensordict")
+    # after dataclass(), frozen=True installs a guard __setattr__ that would reject a
+    # plain assignment here. object.__setattr__ is the same bypass the frozen __init__
+    # uses (see _has_custom_setattr in _tensorclass).
+    object.__setattr__(self, "_tensordict", state.get("tensordict"))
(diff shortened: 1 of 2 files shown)


vmoens (approved):
LGTM thanks

PR #1805, opened 2026-09-21
[BugFix] Release CudaGraphModule without cyclic GC

### LINKED ISSUE #1804: [BUG] CudaGraphModule retains captured graphs in a reference cycle
### System information

- TensorDict `main`: `7256634b4c19b36518d8c9fdde8724e1471c09ea`
- PyTorch: `2.15.0.dev20260917+cu130` and `2.15.0.dev20260920+cu130`
- Python: 3.12 in downstream CUDA CI, independently reproduced on Python 3.13 for the reference cycle

### Problem

`CudaGraphModule` stores `_call_func` on itself, while the nested `_call` function closes over the same `CudaGraphModule` instance. Dropping the wrapper therefore does not release it by reference counting. Its captured `torch.cuda.CUDAGraph` remains alive until cyclic GC runs.

If GC collects an older wrapper while another CUDA stream is capturing a graph, destruction calls `CUDAGraph.reset()`. CUDA forbids graph destruction during an unrelated capture and invalidates the active capture.

This currently breaks the same TorchRL GPU test on `main` and unrelated pull requests:

- https://github.com/pytorch/rl/actions/runs/35493280682/job/106031760651
- https://github.com/pytorch/rl/actions/runs/35512678445/job/106098349464

The test passed with PyTorch nightly `20260914` and fails with nightly `20260917` while using the same TensorDict commit, which exposed the pre-existing lifetime issue.

### Minimal reproduction of the cycle

```python
import gc
import weakref
import warnings

from tensordict.nn import CudaGraphModule

warnings.simplefilter("ignore")
wrapper = CudaGraphModule(lambda x: x)
ref = weakref.ref(wrapper)
del wrapper

assert ref() is None  # currently fails
assert gc.collect() > 0
assert ref() is None
```

The same cycle exists in both the TensorDict-module and ordinary-callable dispatch paths.

### Expected behavior

Dropping the last external reference to `CudaGraphModule` should release the wrapper and its CUDA graph immediately. The stored callable should not keep its owner alive through a closure cycle.

### Proposed fix

Use a weak reference from both nested call paths back to the wrapper, and add CPU-safe regression coverage that verifies both wrapper variants are released without an explicit `gc.collect()`.

AI assistance disclosure: AI-assisted development tools were used during investigation and issue drafting. The reproduction was run against the stated code path.

aswanth-07:
Fixes #1804.

## Summary

- Replace the strong closure reference from both `CudaGraphModule` call paths with a weak proxy.
- Let the last external reference release the wrapper and its captured CUDA graph immediately instead of waiting for cyclic GC.
- Cover the TensorDict-module and ordinary-callable paths with a CPU-safe lifetime regression while preserving the process-wide PyTree registry.

## Context

The stored `_call_func` closed over its owning `CudaGraphModule`, creating a reference cycle. If Python collected an older wrapper while another CUDA stream was capturing, destruction of the older `CUDAGraph` invalidated the active capture. This surfaced in the TorchRL GPU suite under recent PyTorch nightlies.

## Tests

- Patched-module lifetime and PyTree-registry restoration check for both call paths: 2 passed
- `py -3.13 -m py_compile tensordict/nn/cudagraphs.py test/compile/test_compile.py`
- `py -3.10 -m black --check test/compile/test_compile.py`
- `py -3.13 -m ruff check --ignore E731 tensordict/nn/cudagraphs.py test/compile/test_compile.py`
- `git diff --check`

The CUDA capture suite was not run locally. The regression itself is CPU-safe, and upstream CI will exercise the CUDA path.

## AI assistance disclosure

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.


vmoens (approved):

aswanth-07:
Updated in `a6016e0`. The lifetime regression now scopes TensorDict's PyTree exclusion and restores the registry after each case. This addresses the 31 export and Jacobian failures across the Linux and macOS jobs, plus the additional GPU device-map failure. Every affected test passed on the workflow's immediate rerun.

The focused two-case lifetime and registry-restoration check passes, along with Black 24.4.2, Ruff, Python compilation, and `git diff --check`. Fresh workflows are now awaiting maintainer approval.

